PR #1574, opened 2014-07-09
Trusty support with 1.9.1 ruby

nlake44:


obino:
we have a new pull request coming for this.

PR #1666, opened 2015-01-07
Cassandra repair

menivaitsi:
A script that performs repair and/or cleanup of the database nodes, one-by-one.
Must be called on demand, or be put in as cron job on the head node.


AppDB/cassandra_utils.sh
```diff
@@ -0,0 +1,156 @@
...
+# Author: graziano
+
+# Cassandra nodetool.
+CMD="/root/appscale/AppDB/cassandra/cassandra/bin/nodetool"
+
+# The keyspace to work on.
+KEYSPACE="Keyspace1"
+
+set -e
+
```
obino: There should be a check like
[ -x "$CMD" ] || { echo "Cannot find nodetool command!"; exit 1 }
to ensure CMD exists.

AppDB/cassandra_utils.sh
```diff
@@ -0,0 +1,156 @@
...
+        return 1
+}
+
+# Check if we are on the head node.
+if ! am_i_login_node ; then
+        echo "This script should run on the head node."
+        exit 1
+fi
+
+# Check for other running instances of this script.
```
obino: We should work with PID or lockfile instead of grepping and psing blindly. One option could be to save our PID and then remove it. If the PIDfile is there, we assume the process is running, and after too much time we can check if the PID is running.

The other option would be to use something lockfile to create a file and use it as a lock. Create file, make sure we obtain the lock on it, and then delete at the end of repair. If I cannot get the lock within few retries, quit.

AppDB/cassandra_utils.sh
```diff
@@ -0,0 +1,156 @@
...
+
+                # Let's query the Cassandra pool (via the master node) and capture
+                # the following:
+                #  $1 - the status of the node
+                #  $2 - the IP address of the node.
+        done < <(ssh $MASTER $CMD status|grep ^U|awk '{print $1, $2}')
+
+        # Perform operation.
+        for x in ${IPS_TO_USE} ; do
+		echo
```
obino: different indendation?

AppDB/cassandra_utils.sh
menivaitsi: I like the second option. Thanks

AppDB/cassandra_utils.sh
menivaitsi: Will do, thanks.

PR #1667, opened 2015-01-08
close file descriptors

nlake44:


obino:
Not sure I read it correctly, but should we have try() pieces also in the close () functions? And should we have a self.handle.close in a different try catch then self.handle.stop? ie a failure of one may impact the runninng of the latter?

nlake44:
I think its fine being in the same group. If there is a error on doing a stop, then it'll retry doing the stop/close.

PR #1668, opened 2015-01-09
Fixed bug in reestablish_connection.

fastgeert:


obino:
Thanks @fastgeert! We are going to close this pull, since we incorporated your changes in https://github.com/AppScale/appscale/pull/1671.

PR #1670, opened 2015-01-16
Backup script

menivaitsi:
Ready to merge.


AppDB/backup_data.py
```diff
@@ -0,0 +1,367 @@
...
+"""
+import argparse
+import cPickle
+import logging
+import os
+import random
+import re
+import shutil
+import sys
+import threading
```
nlake44: Instead of threading you may want to use the multiprocessing module which will use all the cores on the system.

AppDB/backup_data.py
```diff
@@ -0,0 +1,367 @@
...
+        break
+      else:
+        time.sleep(self.DB_ERROR_PERIOD)
+
+    return success
+
+  def run_backup(self):
+    """ Runs the backup process. Loops on the entire dataset and dumps it into
+    a file.
+    """
```
nlake44: Forgot the Returns section here. If it only returns True then remove the last return in this function.

AppDB/zkappscale/zktransaction.py
```diff
@@ -1340,18 +1345,20 @@ def get_datastore_groomer_lock(self):
...
 
-  def release_datastore_groomer_lock(self):
-    """ Releases the datastore groomer lock. 
+  def release_lock_with_path(self, path):
+    """ Releases lock based on path.
    
+    Args:
+      path: A str, the lock path.
     Returns:
       True on success, False on system failures.
```
nlake44: This function says it returns True/False but in the function there are None returns.

nlake44:
Make sure to run the TCK against this since core datastore was modified/refactored.

AppDB/zkappscale/zktransaction.py
menivaitsi: Actually the return statements are there.

PR #1691, opened 2015-03-04
Use git tags

obino:


bootstrap.sh
```diff
@@ -104,6 +115,15 @@ apt-get install -y git
 if [ ! -d appscale ]; then
         git clone ${APPSCALE_REPO} --branch ${APPSCALE_BRANCH}
         git clone ${APPSCALE_TOOLS_REPO} --branch ${APPSCALE_TOOLS_BRANCH}
+
+        # Use tags if we specified it.
+        if [ -n "$GIT_TAG" ]; then
+                if [ "$GIT_TAG" = "last" ]; then
+                        GIT_TAG="$(cd appscale; git tag|tail -n 1)"
```
menivaitsi: This tail gives the testing-candidate tag.
Is that the intended behavior?

obino: It doesn't if it is a fresh clone. I did remove the tags,since we don't use them, but you still have them locally.

menivaitsi:
:+1:

PR #1693, opened 2015-03-17
Use app-specific crontabs instead of root crontab

cdonati:
This fixes the issue where updating an app's cron.yaml file would not remove the obsolete cron job.

### DIFF
--- a/AppController/lib/cron_helper.rb
+++ b/AppController/lib/cron_helper.rb
@@ -103,10 +103,12 @@ CRON
   end
 
 
-  # Erases all cron jobs on this machine.
-  def self.clear_crontab()
-    Djinn.log_run("crontab -r")
-    self.add_line_to_crontab(NO_EMAIL_CRON)
+  # Erases all cron jobs for all applications.
+  #
+  # Args:
+  #   app: A String that names the appid of this application.
+  def self.clear_app_crontabs
+    Djinn.log_run('rm /etc/cron.d/appscale-*')
   end
 
 
--- a/AppController/lib/cron_helper.rb
+++ b/AppController/lib/cron_helper.rb
@@ -110,6 +110,15 @@ CRON
   end
 
 
+  # Erases all cron jobs for application.
+  #
+  # Args:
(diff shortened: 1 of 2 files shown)


AppController/lib/cron_helper.rb
```diff
@@ -89,20 +93,31 @@ def self.update_cron(ip, port, lang, app)
...
   end
 
 
-  # Erases all cron jobs on this machine.
-  def self.clear_crontab()
-    Djinn.log_run("crontab -r")
-    self.add_line_to_crontab(NO_EMAIL_CRON)
+  # Erases all cron jobs for all applications.
+  #
+  # Args:
```
menivaitsi: Remove this arg section.

AppController/lib/cron_helper.rb
```diff
@@ -63,15 +64,18 @@ def self.update_cron(ip, port, lang, app)
...
       }
+      write_app_crontab(app_crontab, app)
+
     elsif lang == "java"
       cron_file = "/var/apps/#{app}/app/war/WEB-INF/cron.xml"
       return unless File.exists?(cron_file)
       cron_xml = Document.new(File.new(cron_file)).root
       return if cron_xml.nil?
+
+      app_crontab = NO_EMAIL_CRON + "\n"
```
menivaitsi: Could this variable be set before the if..else?

AppController/lib/cron_helper.rb
```diff
@@ -89,20 +93,31 @@ def self.update_cron(ip, port, lang, app)
           Cron Schedule: #{line}
 CRON
           Djinn.log_debug(cron_info)
-          Djinn.log_info("Adding cron line: [#{line}]")
-          add_line_to_crontab(line) if !is_line_in_crontab?(line)
+          app_crontab << line + "\n"
         }
       }
+      write_app_crontab(app_crontab, app)
```
menivaitsi: Also if it's the same deal, call this function once after the if..else.

menivaitsi:
:+1:

PR #1702, opened 2015-03-25
fix for ZK checking blacklisted transactions at scale

nlake44:


AppDB/zkappscale/zktransaction.py
```diff
@@ -951,22 +951,11 @@ def is_blacklisted(self, app_id, txid):
...
-
-    try:
-      blacklist = self.run_with_retry(self.handle.get_children, blacklist_root)
-      return str(txid) in blacklist
-    except kazoo.exceptions.NoNodeError:  # there is no blacklist
-      return False
+      blacklist_root = self.get_blacklist_root_path(app_id)
+      blacklist_txn = PATH_SEPARATOR.join([blacklist_root, 
+        str(txid)]) 
+      return self.run_with_retry(self.handle.exists, blacklist_txn)
```
obino: Not sure if it matters, but it seems that the logic changed: before the code was checking to see if the blacklist_root existed and if not created it, now it check for the full transaction ID path, and presumably will return an exception or error if the root does not exist. Are we guaranteed the root exists?

PR #1705, opened 2015-04-02
Validate crontab lines before inserting them.

cdonati:
This uses the system's crontab to validate each line. Since the only way to do so (that I know of) is to try to actually insert the line, the last step is to restore the previously existing crontab.

### DIFF
--- a/AppController/lib/cron_helper.rb
+++ b/AppController/lib/cron_helper.rb
@@ -129,17 +129,30 @@ CRON
   private
 
 
-  # Adds a single line to this user's crontab. It is assumed that the line is
-  # correctly formatted in cron format.
+  # Checks if a crontab line is valid.
   #
   # Args:
-  #   line: A String that contains the line to add to root's crontab.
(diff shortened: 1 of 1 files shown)


AppController/lib/cron_helper.rb
```diff
@@ -129,17 +129,30 @@ def self.clear_app_crontab(app)
...
-    `crontab -l >> crontab.tmp`
-    `echo "#{line}" >> crontab.tmp`
-    `crontab crontab.tmp`
+  #   line: A String that contains a crontab line.
+  # Returns:
+  #   A boolean that expresses the validity of the line.
+  def self.valid_crontab_line(line)
+    crontab_exists = system('crontab -l')
+    if crontab_exists
+      `crontab -l > crontab.backup`
```
obino: Can this be run in parallel? If so, we may need to make the name unique (ie contab.$timestamp) ...

menivaitsi:
:+1:

PR #1707, opened 2015-04-08
Service

obino:
Remove progenitor, and make sure we start the script only if AppScale is configured.


AppController/scripts/appcontroller
```diff
@@ -16,10 +16,16 @@ DAEMON=$DIR/djinnServer.rb
 DAEMON_NAME=appcontroller
```
cdonati: Does this need to be changed to appscale-controller as well?

nlake44:
This looks good. Has it run through OCD?

Issue #1716, opened 2015-04-12
Error in package finding in golang


onlyangel:
I get:
`2015/04/12 22:14:35 Can't find package "github.com/go-sql-driver/mysql" in $GOPATH: cannot find package "github.com/go-sql-driver/mysql" in any of:
	/root/appscale/AppServer/gopath/src/pkg/github.com/go-sql-driver/mysql (from $GOROOT)
	/root/appscale/AppServer/goroot/src/github.com/go-sql-driver/mysql (from $GOPATH)
2015/04/12 22:14:36 go-app-builder: build timing: 7Ã—6g (237.372602ms total), 6Ã—gopack (28.930674ms total), 0Ã—6l (0 total)
2015/04/12 22:14:36 go-app-builder: failed running 6g: exit status 1
`

The compiler is searching in 
`/root/appscale/AppServer/gopath/src/pkg/{packageurl}` when it suposes to be `/root/appscale/AppServer/gopath/src/{packageurl}`
and in
`/root/appscale/AppServer/goroot/src/{packageurl}`when it suposes to be `/root/appscale/AppServer/goroot/src/pkg/{packageurl}`

As seen in the filesystem.

onlyangel:
I think I can create a patch for this. Let me clone the repo.

obino:
@onlyangel thanks!

PR #1726, opened 2015-04-24
Fixing enviroment paths GOPATH and GOROOT

onlyangel:
I was working in my appengine app when a client ask me to make it work in situ, I try appscale with go and I find a problem with the imports and file the issue 1716 (https://github.com/AppScale/appscale/issues/1716) 

With a few days of experimentation I find this little error maybe in a copy paste of the path's common error and really hard to find. 

Now it works as a charm with my apps. But no with the app suggested in the contribute page ( github.com/mjibson/goread ) because it uses a library from golang ( transform ) where I have finded an other error. That I am working on right now. 

The problem was the confusion of the assignation of two environment variables. Really a two lines switch.

### DIFF
--- a/AppManager/app_manager_server.py
+++ b/AppManager/app_manager_server.py
@@ -119,8 +119,8 @@ def start_app(config):
   start_cmd = ""
   stop_cmd = ""
   env_vars = config['env_vars']
-  env_vars['GOPATH'] = '/root/appscale/AppServer/goroot/'
-  env_vars['GOROOT'] = '/root/appscale/AppServer/gopath/'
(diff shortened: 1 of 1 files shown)


nlake44:
:+1:  Good work @onlyangel

onlyangel:
Thanks @nlake44  =D. There is other issue that I'm not sure to be able to help, just let me document it better. Any way @obino already tell me that you have identified it and are workign arround it.

nlake44:
What's the other issue?

onlyangel:
It's a problem running appscale over docker.

PR #1730, opened 2015-05-01
Add AppController functions to check and set deployment ID

cdonati:


AppController/djinn.rb
```diff
@@ -1164,6 +1164,28 @@ def set_property(property_name, property_value, secret)
...
+
+  # Sets deployment ID in ZooKeeper.
+  # Args:
+  #   id: A string that contains the deployment ID.
+  def set_deployment_id(secret, id)
+    if !valid_secret?(secret)
+      return BAD_SECRET_MSG
+    end
+
+    ZKInterface.set('/deployment_id', id, false)
```
menivaitsi: What happens if the deployment ID already exists?

menivaitsi: Please add 'return' here. Increases readability.

PR #1736, opened 2015-05-27
Backup service

menivaitsi:
DO NOT MERGE.


AppController/lib/backup_restore_service.rb
```diff
@@ -0,0 +1,31 @@
...
+# Starts and stops the backup and recovery service.
+module BackupRecoveryService
+
+  BR_PORT = '8423'
+
+  # Starts the BR Service on this machine. We don't want to monitor
+  # it ourselves, so just tell monit to start it and watch it.
+  def self.start()
+    start_cmd = "/usr/bin/python /root/appscale/AppDB/backup/backup_recovery_service.py"
+    # The stop command doesn't work and relies on terminate.rb.
```
cdonati: I don't understand this comment. Is it just because the the start method requires a stop_cmd argument?

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+  """ Backup Cassandra snapshot data directories/files. 
+  
+  Args:
+    storage: A str, the storage that is used for storing the backup.
+    path: A str, the full backup filename path to use for cloud backup.
+  Returns:
+    The path to the backup file on success, None otherwise.
+  """
+  logging.info("Starting new db backup.")
+  clear_old_snapshots()
```
cdonati: Not sure if this makes sense to do, but would we want to keep track of which snapshots we create so that we don't delete all of them?

AppController/lib/backup_restore_service.rb
menivaitsi: I'll remove that comment. Thanks for pointing it out.

AppDB/backup/cassandra_backup.py
menivaitsi: So old snapshots are not useful at this point.
We clean up, we create a new snapshot which we then tar up and we use that for recovery.

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+  except CalledProcessError as error:
+    logging.error("Error while creating dir '{0}'. Error: {1}".
+      format(BACKUP_DIR_LOCATION, str(error)))
+    return None
+
+  backup_file_location = BACKUP_FILE_LOCATION
+
+  # Delete previous backup.
+  try:
+    call(["rm", "-f", backup_file_location])
```
cdonati: Since the file is deleted at the end of the process, is this needed? It seems like it could hide issues if something isn't working.

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+
+  # Create backups dir if not there.
+  try:
+    call(["mkdir", "-p", BACKUP_DIR_LOCATION])
+  except CalledProcessError as error:
+    logging.error("Error while creating dir '{0}'. Error: {1}".
+      format(BACKUP_DIR_LOCATION, str(error)))
+    return None
+
+  backup_file_location = BACKUP_FILE_LOCATION
```
cdonati: Would we want to add a timestamp to the file name for debugging purposes if needed?

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+  return True
+
+def remove_local_backup_file(local_file=BACKUP_FILE_LOCATION):
+  """ Removes the local backup file.
+
+  Args:
+    local_file: A str, the path to the backup file to delete.
+  """
+  try:
+    call(['rm', '-rf', local_file])
```
cdonati: Does this need '-r'?

AppDB/backup/cassandra_backup.py
menivaitsi: It's deleted only if the storage backend is remote (like GCS).
For local backups, we don't delete it.

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+def restore_data(storage, path=''):
+  """ Restores the Cassandra snapshot. 
+
+  Args:
+    storage: A str, one of the StorageTypes class members.
+    path: A str, the name of the backup file to restore from.
+  """
+  if storage == StorageTypes.GCS:
+    # Download backup file and store locally with a fixed name.
+    if not gcs_helper.download_from_bucket(path, BACKUP_FILE_LOCATION):
```
cdonati: It seems like we'd want to use a different location to avoid confusion if the system ever ends up in a broken state.

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,308 @@
...
+    remove_local_backup_file()
+
+  logging.info("Done with restore.")
+  return True
+
+if "__main__" == __name__:
+  logging.getLogger().setLevel(logging.DEBUG)
+
+  backup_data(storage='', path='')
+  restore_data(storage='', path='')
```
cdonati: What is this doing? If the file is run directly, it performs a backs up and then immediately restores?

AppDB/backup/cassandra_backup.py
menivaitsi: Do you mean keeping multiple backups?
If not, then the metadata for that file are stored both in the local fs and on GCS, so that's not needed.

AppDB/backup/cassandra_backup.py
menivaitsi: It's an example of how you can run backup/restore by hand.

AppDB/backup/cassandra_backup.py
menivaitsi: No, I'll remove it.

AppDB/backup/gcs_helper.py
```diff
@@ -0,0 +1,195 @@
...
+import os
+import re
+import requests
+import subprocess
+import urllib
+
+from backup_recovery_constants import HTTP_OK
+
+# The upload request timeout in seconds (12 hours).
+REQUEST_TIMEOUT = 12*60*60
```
cdonati: Will this eventually be based on the size of the data?

menivaitsi: There are a few options here. We can discuss it later on.

AppDB/backup/cassandra_backup.py
cdonati: What if users are creating their own snapshots for other purposes?

AppDB/backup/cassandra_backup.py
cdonati: I might be confused, but does this mean that the local backup they did previously gets destroyed?

AppDB/backup/cassandra_backup.py
cdonati: Yeah, for multiple backups. For example, if one backup fails and we want to go back and see the state of it after another backup starts.

AppDB/backup/cassandra_backup.py
menivaitsi: You only need the most recent snapshot for backup purposes I believe.

AppDB/backup/cassandra_backup.py
menivaitsi: Correct. We could rename it and keep it so long as there's space.

AppDB/backup/cassandra_backup.py
menivaitsi: Ok, we could rename it instead of deleting it when a new backup starts. Depends on if there's available space or not.

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,352 @@
...
+from backup_recovery_constants import BACKUP_ROLLBACK_SUFFIX
+from backup_recovery_constants import CASSANDRA_DATA_SUBDIRS
+from backup_recovery_constants import PADDING_PERCENTAGE
+from backup_recovery_constants import StorageTypes
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "../"))
+import dbconstants
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "../cassandra/"))
+import start_cassandra
```
cdonati: Would it be cleaner to do something like `from cassandra import start_cassandra`, etc. to reduce the number of times the sys path is modified?

AppDB/backup/cassandra_backup.py
```diff
@@ -0,0 +1,352 @@
...
+def backup_data(storage, path=''):
+  """ Backup Cassandra snapshot data directories/files.
+
+  Args:
+    storage: A str, the storage that is used for storing the backup.
+    path: A str, the full backup filename path to use for cloud backup.
+  Returns:
+    The path to the backup file on success, None otherwise.
+  """
+  if storage not in (StorageTypes()).get_storage_types():
```
cdonati: Are the parens around StorageTypes() needed?

PR #1749, opened 2015-06-19
Upgrade

obino:
This patch is to allow bootstrap to use tag when ugrading. This is needed to ensure that by default the upgrade is toward a released version not the HEAD of master.


menivaitsi:
:+1:

obino:
https://ocd.appscale.com:8080/job/Daily%20Build/214/ All clean from OCD.

PR #1750, opened 2015-06-22
Zookeeper backups

menivaitsi:
Ready for merge.

### DIFF
--- a/AppDB/backup/backup_recovery_helper.py
+++ b/AppDB/backup/backup_recovery_helper.py
@@ -2,6 +2,38 @@
 
 import logging
 import os
+import statvfs
+import sys
+import tarfile
+from os.path import getsize
+
+import backup_recovery_constants
+import backup_exceptions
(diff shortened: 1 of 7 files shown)


AppDB/backup/backup_recovery_helper.py
```diff
@@ -13,6 +45,81 @@ def does_file_exist(path):
...
+def get_snapshot_paths(service):
+  """ Returns a list of file names holding critical data for the given service.
+
+  Args:
+    service: A str, the service for which we're getting the data files.
+    Currently there is support for Cassandra and Zookeeper.
+  Returns:
+    A list of full paths.
+  """
+  if service == 'cassandra':
```
cdonati: This probably doesn't need to be an if/else. Just return if !=. "look_for" doesn't need to be defined conditionally.

menivaitsi: Leftover.. Thanks for pointing out.

PR #1758, opened 2015-07-09
Skip unneeded rollback instead of throwing exception

cdonati:


obino:
https://ocd.appscale.com:8080/job/Daily%20Build/237/ OCD run.

obino:
good job.

PR #1760, opened 2015-07-13
Fetch more references

cdonati:
This addresses the issue of insufficient valid entities being fetched from the datastore.

### DIFF
--- a/AppDB/datastore_server.py
+++ b/AppDB/datastore_server.py
@@ -527,7 +527,10 @@ class DatastoreDistributed():
                                       column_names=dbconstants.COMPOSITE_SCHEMA)
  
   def delete_index_entries(self, entities):
-    """ Deletes the entities in the DB.
+    """ Deletes the index entries for the given entities.
+
+    This only deletes the indexes in the ascending and descending property
+    tables.
 
     Args:
        entities: A list of entities for which their 
--- a/AppDB/datastore_server.py
+++ b/AppDB/datastore_server.py
@@ -1860,27 +1860,64 @@ class DatastoreDistributed():
         entities.append(result[key][dbconstants.APP_ENTITY_SCHEMA[0]])
     return entities 
 
-  def __fetch_entities(self, refs, app_id):
-    """ Given the results from a table scan, get the references.
-    
-    Args: 
-      refs: key/value pairs where the values contain a reference to 
-            the entitiy table.
-      app_id: A string, the application identifier.
+  def __extract_rowkeys_from_refs(self, refs):
+    """ Extract the rowkeys to fetch from a list of references
+
+    Args:
+      refs: key/value pairs where the values contain a reference to the
+            entitiy table.
     Returns:
-      Entities retrieved from entity table.
+      A list of rowkeys.
     """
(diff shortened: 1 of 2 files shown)


AppDB/datastore_server.py
```diff
@@ -2294,19 +2294,57 @@ def __kind_query(self, query, filter_info, order_info):
...
+        current_limit,
+        offset=0,
+        start_inclusive=start_inclusive,
+        end_inclusive=end_inclusive
+      )
+
+      new_entities = self.__fetch_entities(references, clean_app_id(query.app()))
+      entities.extend(new_entities)
+
+      # If we have enough valid entities to satisfy the query, we can stop
```
obino: I think the comment reads better with something like: If we satisfied the original query, we are done.

AppDB/datastore_server.py
```diff
@@ -2294,19 +2294,57 @@ def __kind_query(self, query, filter_info, order_info):
...
+      # fetching more references.
+      if len(entities) >= limit:
+        return entities[:limit]
+
+      # If we received fewer references than we asked for, we do not need to ask
```
obino: I think the rationale here is that if we have fewer entities than references, we had some invalid entities, so we need to fetch more entities. Or something like that?

AppDB/datastore_server.py
```diff
@@ -2294,19 +2294,57 @@ def __kind_query(self, query, filter_info, order_info):
...
+        end_inclusive=end_inclusive
+      )
+
+      new_entities = self.__fetch_entities(references, clean_app_id(query.app()))
+      entities.extend(new_entities)
+
+      # If we have enough valid entities to satisfy the query, we can stop
+      # fetching more references.
+      if len(entities) >= limit:
+        return entities[:limit]
```
obino: This should return something like  entities = [self.default_namespace()] + entities as the other exit point?

AppDB/datastore_server.py
```diff
@@ -1862,28 +1877,65 @@ def __fetch_entities_from_row_list(self, rowkeys, app_id):
...
 
-  def __fetch_entities(self, refs, app_id):
-    """ Given the results from a table scan, get the references.
-    
-    Args: 
-      refs: key/value pairs where the values contain a reference to 
-            the entitiy table.
-      app_id: A string, the application identifier.
+  def __extract_rowkeys_from_refs(self, refs):
+    """ Extract the rowkeys to fetch from a list of references
```
menivaitsi: Add '.' at the end.

AppDB/datastore_server.py
```diff
@@ -3177,71 +3324,70 @@ def composite_v2(self, query, filter_info):
 
     return self.__fetch_entities(index_result, clean_app_id(query.app()))
 
-  def __extract_entities_from_indexes(self, index_result, direction):
-    """ Takes the index values and creates partial entities out of them.
-   
+  def __extract_value_from_index(self, index_entry, direction):
+    """ Takes the index values and returns the value of the property.
```
menivaitsi: Should "index values" be "index value"?

PR #1764, opened 2015-07-22
Don't crash the AC when a cron line is unparseable

cdonati:
The function should return a list.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/284/

PR #1767, opened 2015-07-24
Fixing bug in blobstore uploads via HTTPS.

menivaitsi:
Ready for merge.

https://ocd.appscale.com:8080/job/Daily%20Build/305/

### DIFF
--- a/AppServer/google/appengine/api/blobstore/blobstore_stub.py
+++ b/AppServer/google/appengine/api/blobstore/blobstore_stub.py
@@ -289,9 +289,9 @@ class BlobstoreServiceStub(apiproxy_stub.APIProxyStub):
                                   max_bytes_total,
                                   bucket_name)
 
-    _, host, _, _, _, _ = urlparse.urlparse(
-      self.request_data.get_request_url(request_id))
-    protocol = self.request_data.get_scheme(request_id)
+    scheme = self.request_data.get_scheme(request_id)
+    protocol, host, _, _, _, _ = urlparse.urlparse(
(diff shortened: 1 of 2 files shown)


AppServer/google/appengine/api/blobstore/blobstore_stub.py
```diff
@@ -289,8 +289,9 @@ def _Dynamic_CreateUploadURL(self, request, response, request_id):
                                   max_bytes_total,
                                   bucket_name)
 
-    protocol, host, _, _, _, _ = urlparse.urlparse(
-        self.request_data.get_request_url(request_id))
+    _, host, _, _, _, _ = urlparse.urlparse(
+      self.request_data.get_request_url(request_id))
```
cdonati: Can the call to get_scheme go inside of the get_request_url method? That way, anything else that calls get_request_url would also be fixed.

AppServer/google/appengine/tools/devappserver2/wsgi_request_info.py
```diff
@@ -159,3 +159,16 @@ def get_instance(self, request_id):
...
+
+    Args:
+      request_id: The string id of the request making the API call.
+
+    Returns:
+      One of 'http', 'https'.
+    """
+    with self._lock:
+      environ = self._request_wsgi_environ[request_id]
+      return environ['HTTP_X_FORWARDED_PROTO']
```
cdonati: I would suggest setting it to http (or https) and overriding it if the header indicates otherwise. If for some reason a request comes in without that header, we still want to return one of the two values.

PR #1770, opened 2015-07-28
Handle db values that contain the key delimiter

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/323/

PR #1771, opened 2015-07-29
Filter certain entities with repeated properties

cdonati:
When a query filter has multiple equality filters on a repeated
property, remove entities that do not satisfy both filters.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/335/

PR #1773, opened 2015-07-30
Make the writing to log synchronous.

obino:
This prevent log files to be clipped in case of errors.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/336/

PR #1776, opened 2015-08-04
Fix boolean parsing

obino:
Second part: also integers needs to be translated into String, otherwise we may see SOAP errors. Also make sure that @options[] is not used outside djinn, but we use explicit parameters instead.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/360/

PR #1777, opened 2015-08-04
Fetch more entities if needed during a composite query

cdonati:
This fix is needed due to the additional filtering logic we use for multiple equality filters.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/363/

PR #1778, opened 2015-08-05
Use Ruby's URI library to sanitize URL

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/367/

PR #1780, opened 2015-08-07
Added group to FastStart AppScalefile.

sjuncal11:


scripts/fast-start.sh
```diff
@@ -137,6 +137,7 @@ if [ ! -e AppScaleFile ]; then
         echo "admin_user : $ADMIN_EMAIL" >> AppScalefile
         echo "admin_pass : $ADMIN_PASSWD" >> AppScalefile
     fi
+    echo "group : faststart-${PROVIDER}" >> AppScalefile
```
obino: This will make highly likely that one will be in the same group (just use faststart twice with the same provider). Have you tested if there is any sideeffect to try it twice? Or to try while another faststart instance is already running? Good work! :+1:

menivaitsi:
:+1:

sjuncal11:
There are a number of AppScale deployments with the same "group" parameter in the logs currently, so I don't think that would be an issue. I tested it by a. running the script multiple times on the same machine and b. by having multiple instances running with the same group defined (faststart-AWS). No unusual behavior in either scenario.

PR #1782, opened 2015-08-08
When validating an index, check all prop values

cdonati:
This fix can be optimized later to return faster when the property does not contain repeated values.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/376/

cdonati:
Demo completed successfully.

PR #1788, opened 2015-08-20
Changed the tag logic for bootstrap.

obino:
First of all, one cannot specify --branch and --tag at the same time in
the bootstrap script. We also added a 'dev' tag that allows to pull from
the HEAD.

For upgrade, we don't want to change branch: that will need to
be done by hand before calling the script. The idea is to disable
--branch when upgrading, then if the current branch is master, the upgrade
will be the specified tag. If the current branch is not master, we simply
do a pull.


bootstrap.sh
```diff
@@ -98,6 +98,26 @@ while [ $# -gt 0 ]; do
...
+if [ -z "${GIT_TAG}" ]; then
+        GIT_TAG="last"
+else
+        # We don't use Tag and Branch at the same time.
+        if [ "${APPSCALE_BRANCH}" != "master" ]; then
+                echo "--branch cannot be specified with --tag"
+                exit 1
+        fi
+        if [ "${APPSCALE_TOOLS_BRANCH}" != "master" ]; then
+                echo "--repo-branch cannot be specified with --tag"
```
cdonati: Should this be "tools-branch"?

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/436/

bootstrap.sh
```diff
@@ -116,12 +136,12 @@ if [ ! -d appscale ]; then
         # We split the commands, to ensure it fails if branch doesn't
         # exists (Precise git will not fail otherwise).
         git clone ${APPSCALE_REPO} appscale
-        (cd appscale; git checkout ${APPSCALE_BRANCH})
+        (cd appscale; git checkout ${APPSCALE_BRANCH}; git fetch --all)
```
cdonati: We probably just want to do `get fetch origin` here. `--all` is for all remotes.

bootstrap.sh
```diff
@@ -133,12 +153,49 @@ fi
...
         APPSCALE_MINOR="$(sed -n 's/.*\([0-9]\)\+\.\([0-9]\)\+\.[0-9]/\2/gp' appscale/VERSION)"
         if [ -z "$APPSCALE_MAJOR" -o -z "$APPSCALE_MINOR" ]; then
                 echo "Cannot determine version of AppScale!"
                 exit 1
         fi
+
+        # This is an upgrade, so let's make sure we use a tag that has
+        # been passed, or the last one available. Let's fetch all the
+        # available tags first.
+        (cd appscale; git fetch --all)
```
cdonati: Same for `git fetch origin`

bootstrap.sh
```diff
@@ -180,16 +237,15 @@ if [ -d appscale/.appscale/certs ]; then
...
-        (cd appscale-tools; git fetch --all)
-        if [ -z "$GIT_TAG" -o "$GIT_TAG" = "last" ]; then
-                GIT_TAG="$(cd appscale; git tag|tail -n 1)"
+
+        # Let's upgrade the repository: if GIT_TAG is empty we are on HEAD.
+        if [ -n "${GIT_TAG}" ]; then
+                (cd appscale; git checkout "$GIT_TAG")
+                (cd appscale-tools; git checkout "$GIT_TAG")
+        else
+                (cd appscale; git pull)
```
cdonati: I think we can just do `git checkout "${CURRENT_BRANCH}` here, right? We don't need to re-check origin at this point.

bootstrap.sh
obino: I think it can be removed alltogether since we are after a clone.

bootstrap.sh
obino: We decided to leave the pull since CURRENT_BRANCH is assigned above.

PR #1796, opened 2015-08-31
Mail

obino:


scripts/sendmail
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
nlake44: Hooray. More bash from the bash master.

obino: I can do also posix shell :)

PR #1806, opened 2015-09-09
Don't raise exception if prop_name is not found

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/497/

PR #1815, opened 2015-09-27
Recover composite indices

cdonati:
Since reverse_lex is applied to values after the delimiter is replaced, it's possible for the delimiter to get stored as part of the property value.

### DIFF
--- a/AppDB/datastore_server.py
+++ b/AppDB/datastore_server.py
@@ -49,6 +49,8 @@ from google.appengine.ext import db
 from google.appengine.ext.db.metadata import Namespace
 from google.appengine.ext.remote_api import remote_api_pb
 
+from google.net.proto.ProtocolBuffer import ProtocolBufferDecodeError
+
 from M2Crypto import SSL
 
 # Buffer type used for key storage in the datastore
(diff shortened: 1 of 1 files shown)


AppDB/datastore_server.py
```diff
@@ -3653,10 +3655,30 @@ def __extract_entities_from_composite_indexes(self, query, index_result):
       if definition.ancestor() == 1:
         ancestor = tokens.pop(0)[:-1]
       distinct_str = ""
+      value_index = 0
       for def_prop in definition.property_list():
-        value = tokens.pop(0) 
+        # If the value contained the separator, try to recover the value.
+        if len(tokens[:-1]) != len(definition.property_list()):
```
obino: Question: could that ever be that definition.propety.list() is bigger than the possible number of values in tokens (ie wrong query or something like that)?

obino:
:+1:

AppDB/datastore_server.py
cdonati: I don't think it's possible if things are working as intended. The AppServer stops the query if there's not an index defined that matches the query. It shouldn't use a composite index if there are more values in the index than properties in the query.

I changed it to '>', though, since I think it conveys the issue more clearly.

PR #1828, opened 2015-10-14
Copy the key file so the tools can use it

cdonati:
If appscale is started with a cluster configuration with tools that are outside of the deployment, the tools will not know where to find the key file.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/596/

AppController/djinn.rb
```diff
@@ -1139,6 +1139,14 @@ def upload_app(archived_file, file_suffix, email, secret)
         File.rename(archived_file_old, archived_file)
       end
 
+      # Make sure the tools have access to the keyfile.
+      # TODO: Don't use the tools to deploy the application.
+      appscale_key_file = "/etc/appscale/keys/cloud1/#{keyname}.key"
```
obino: keyname is defined below: the assignment should be moved before using it.

PR #1829, opened 2015-10-18
Better checks and logs in Hermes.

menivaitsi:


AppDB/backup/backup_recovery_helper.py
```diff
@@ -371,7 +371,7 @@ def deploy_apps(app_paths):
     appscale_info.get_secret())
 
   # Wait for Cassandra to come up after a restore.
-  time.sleep(5)
+  time.sleep(15)
```
obino: This should have something similar to what we do during boot: check when the port cassandra listen to is open. Not quite sure in python, but open the TCP socket should be enough to ensure the daemon is listening.

AppDB/backup/gcs_helper.py
```diff
@@ -112,7 +112,7 @@ def download_from_bucket(full_object_name, local_path):
     return False
 
   # Compare to GCS file size.
-  if not content['size'] < available_space:
+  if int(content['size']) >= int(available_space):
     logging.error('Not enough space on the VM to download a backup.')
```
obino: The message shouldn' have 'VM' in it.

obino: That is: remove 'on the VM'.

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/624/

PR #1841, opened 2015-11-07
Only decode results when necessary

cdonati:
This should substantially speed up queries that return large entities or a large amount of entities.


menivaitsi:
:+1:

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/677/

PR #1843, opened 2015-11-12
The app_manager do not use the list of DBs.

obino:
Thus this patch removes it from the paramenters it accepts. This will
remove the need for get_nearest_db_ip in the djinn.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/706

PR #1848, opened 2015-11-21
Handle order operation

cdonati:


obino:
:+1: nicely done :) Would it be worth it to add some comment in the function to explain this addition and why was needed?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/775/

PR #1860, opened 2015-12-04
Poll for new tasks

cdonati:


Hermes/hermes.py
```diff
@@ -98,6 +102,9 @@ def main():
   logging.info("Hermes is up and listening on port: {0}.".
     format(options.port))
 
+  tornado.ioloop.PeriodicCallback(poll, hermes_constants.POLLING_INTERVAL).\
```
menivaitsi: Add a comment above this to explain the loop.

Hermes/hermes.py
```diff
@@ -10,6 +10,7 @@
 import tornado.web
 import urllib
 
+import tornado.ioloop
 from tornado.ioloop import IOLoop
```
menivaitsi: If you import the whole module, you don't need this.
Just remove it and change: IOLoop.instance().start() accordingly.

PR #1864, opened 2015-12-05
Cleanup AppScaleTaskQueueClient Get AppId Code

caseyoneill:
Attempting to fix issue where the task queue would throw a null pointer exception while adding a job to the queue. The attempted fix now accesses the appId through the proper private method instead of directly reading the private variable. This ensures that appId is lazily loaded or always loaded is null.

### DIFF
--- a/AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
+++ b/AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
@@ -62,13 +62,12 @@ public class AppScaleTaskQueueClient
         connManager.setMaxForRoute(new HttpRoute(localhost), MAX_CONNECTIONS_PER_ROUTE_LOCALHOST);
         
         client = new DefaultHttpClient(connManager);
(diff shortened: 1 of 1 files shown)


AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
```diff
@@ -135,8 +134,11 @@ private String getNginxPort()
 
     private String getAppId()
     {
-        String appId = System.getProperty("APPLICATION_ID");
-        return appId;
+    	if(this.appId == null)
+    	{
+    		this.appId = System.getProperty("APPLICATION_ID");
```
menivaitsi: Reading twice in a row has no value IMHO.
I think it's better to print a warning instead in this loop.

menivaitsi: Never mind. I can't read, it seems.

menivaitsi:
OCD build: https://ocd.appscale.com:8080/job/Daily%20Build/831/

PR #1869, opened 2015-12-09
Limit the numer of zookeeper logs to keep.

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/847/

PR #1872, opened 2015-12-11
Use combined app log if unable to parse cron xml

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/854/

PR #1874, opened 2015-12-11
Encapsulate url

cdonati:
This is to prevent cron from evaluating the characters in the url.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/858/

PR #1880, opened 2015-12-16
Use curl throughout.

obino:
We had both curl and wget in our scripts. This patch uniform our usage to
curl which seems the one mostly used in the google SDK.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/876/

PR #1883, opened 2015-12-17
Get available disk space in bytes

cdonati:
Previously, we were comparing the size of the object in bytes to free space in kb.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/881/

PR #1884, opened 2015-12-17
Don't run a 'nodetool refresh' right after cassandra restore

cdonati:
These commands were getting stuck when restoring a multinode deployment.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/883/

PR #1886, opened 2015-12-18
Put deployment ID node inside appscale path

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/886/

PR #1889, opened 2015-12-22
Ignore not empty nodes while restoring Zookeeper data.

menivaitsi:


menivaitsi:
OCD: https://ocd.appscale.com:8080/job/Daily%20Build/904/

PR #1892, opened 2015-12-23
Make sure lsof catches only servers.

obino:
Before we were catching anything related to the port, also outgoing
connections (ie a curl to a web site for example). We need to only check
the services ie the processes that are listening to the port.


obino:
OCD run: https://ocd.appscale.com:8080/job/Daily%20Build/914/

menivaitsi:
:+1:

PR #1894, opened 2015-12-28
Added a hourly status print in the controller logs.

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/913

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/915/.

AppController/djinn.rb
```diff
@@ -1778,6 +1786,18 @@ def job_start(secret)
           scale_appservers_across_nodes()
         end
       }
+
+      # We want to print some statistic in the logs every hour or so, to
```
menivaitsi: This comment is too verbose. Please change it to something like:

"Print stats in the log recurrently; works as a heartbeat mechanism."

AppController/djinn.rb
```diff
@@ -1778,6 +1786,18 @@ def job_start(secret)
...
+      # ensure that the thread is working correctly.
+      if last_print < (Time.now.to_i - 60 * PRINT_STATS_MINUTES)
```
menivaitsi: Please make 60 a constant.

menivaitsi:
:+1:

obino:
OCD https://ocd.appscale.com:8080/job/Daily%20Build/917/

AppController/djinn.rb
obino: This constant would be to identify the number of seconds per minute, and will only be used here (there is another 60 in the code, but the context is slightly different). Given that the unit is fixed (60) it seems very verbose and redundant to me to create a constant for it , since everyone knows that unit. Thoughts?

PR #1899, opened 2016-01-06
Change pid file location for trusty

cdonati:
The os check was needed due to this issue: https://github.com/miketheman/nginx/issues/248

### DIFF
--- a/AppController/lib/nginx.rb
+++ b/AppController/lib/nginx.rb
@@ -769,22 +769,12 @@ CONFIG
 
   # Set up the folder structure and creates the configuration files necessary for nginx
   def self.initialize_config
-    nginx_pid_file = '/var/run/nginx.pid'
-    begin
-      os_version = `lsb_release -c`
-      if os_version.split.last == 'trusty'
-        nginx_pid_file = '/run/nginx.pid'
-      end
-    rescue Errno::ENOENT => exception
-      Djinn.log_debug(exception)
-    end
-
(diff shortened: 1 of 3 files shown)


AppController/lib/nginx.rb
```diff
@@ -769,12 +769,22 @@ def self.create_app_config(my_public_ip, my_private_ip, proxy_port,
 
   # Set up the folder structure and creates the configuration files necessary for nginx
   def self.initialize_config
+    nginx_pid_file = '/var/run/nginx.pid'
+    begin
+      os_version = `lsb_release -c`
+      if os_version.split.last == 'trusty'
+        nginx_pid_file = '/run/nginx.pid'
+      end
```
obino: Given that all modern distro have /run, and /var/run is a symlink to it, I would propose either to try /run also on precise, or to reverse the logic, and put the pid in /var/run only if it is precise, so we will be able to remove it as soon as we move away from precise.

obino:
:+1: good find on the nginx issue!

cdonati:
Precise: https://ocd.appscale.com:8080/job/Daily%20Build/936/
Trusty: https://ocd.appscale.com:8080/job/Daily%20Build/937/

cdonati:
Precise: https://ocd.appscale.com:8080/job/Daily%20Build/940/
Trusty: https://ocd.appscale.com:8080/job/Daily%20Build/941/

PR #1901, opened 2016-01-08
Have Monit properly check (via the port) that the appserver is running.

menivaitsi:
OCD: https://ocd.appscale.com:8080/job/Daily%20Build/942/


lib/monit_app_configuration.py
```diff
@@ -28,6 +28,7 @@ def create_config_file(watch, start_cmd, stop_cmd, ports, env_vars={},
     max_memory: An int that names the maximum amount of memory that this process
       is allowed to use (in megabytes) before monit should restart it.
     syslog_server: The IP of the remote syslog server to use.
+    host: The IP of the server where this file will reside.
```
obino: I think the comment is not clear. What about something like: The private IP of the host to check the port. Used to reliably detect a running dev_appserver.

PR #1907, opened 2016-01-15
Match the right pgid when restarting the datastore

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/956/

PR #1916, opened 2016-01-28
Avoid extra messages in the logs.

obino:
Added silencing options to curl to avoid extra messages in syslog.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/985/

PR #1919, opened 2016-01-31
Use monit to stop all running services.

obino:
All AppScale services are monitored by monit, so instead of using kill to
stop them, we can use monit. This allow us to not kill generic processes
(like python) and to use our script to stop services (which means we stop
also child processes).


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1000/

obino:
OCD  precise run: https://ocd.appscale.com:8080/job/Daily%20Build/1020/

PR #1923, opened 2016-02-03
Make sure an application is restarted correctly upon appscale up

obino:
This patch make sure the application is enabled when restarting it, or adding appserver.


menivaitsi:
:+1: 

@cdonati you can merge this.

PR #1924, opened 2016-02-03
Save application ports to persistent storage.

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1066/

AppController/djinn.rb
```diff
@@ -4628,27 +4630,41 @@ def setup_appengine_application(app, state)
...
       end
-      Djinn.log_info("Waiting for app data to have instance info for app named #{app}: #{app_data}")
+      Djinn.log_info("Waiting for app data to have instance info for app named #{app}")
       Kernel.sleep(SMALL_WAIT)
     }
 
     my_public = my_node.public_ip
     my_private = my_node.private_ip
 
+    # Let's create an entry for the application if we don't already have # it.
```
cdonati: There's an extra `#` from a previous line break.

AppController/djinn.rb
```diff
@@ -4695,9 +4713,12 @@ def setup_appengine_application(app, state)
 
       if @app_info_map[app]['nginx'].nil?
         @app_info_map[app]['nginx'] = nginx_app_port
+      end
+      if @app_info_map[app]['nginx_https'].nil?
+        @app_info_map[app]['nginx_https'] = nginx_app_https
```
cdonati: Do you think we should test what happens if this port isn't free? For example, a user might quickly deploy an app before AppScale can restart the old apps on an up.

obino: I think we are covered by find_lowest_free_port and because this function is under lock. Do you think I miss a use case? btw I changed find free_port to check also partially allocated ports, just in case.

cdonati: It looks like the port found by `find_lowest_free_port` is ignored when there's a value in the `@app_info_map`. Can't the port be taken before the lock is established?

obino: ah, understood. To be the best of knowledge that can happen only from restoring an old state (look for the variable @restored in start_appengine()), hence there shouldn't no be collision.

obino:
ocd run https://ocd.appscale.com:8080/job/Daily%20Build/1087/

obino:
This PS needs to be merged with https://github.com/AppScale/appscale-tools/pull/479

PR #1926, opened 2016-02-04
Fix logic so that 'stop' actually works

solresol:
Previous version of this file would check to see if the service was running, and if it was running, refuse to stop it.

### DIFF
--- a/AppController/scripts/appcontroller
+++ b/AppController/scripts/appcontroller
@@ -44,14 +44,18 @@ do_stop()
 
 case "$1" in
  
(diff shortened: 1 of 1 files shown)


menivaitsi:
Good catch! We'll run it through testing.

Trusty build: https://ocd.appscale.com:8080/job/Daily%20Build/1033/

obino:
Tested on a 4 nodes: the stop is still a no-op, since it's not hooked to the AC internals, but the script does its job stopping the AC once.

PR #1931, opened 2016-02-06
Remove monit config files when stopping services

cdonati:


AppController/lib/monit_interface.rb
```diff
@@ -133,17 +118,13 @@ def self.write_monit_config(watch, start_cmd, stop_cmd, port,
...
-
-  def self.remove(watch, remote_ip=nil, remote_key=nil)
-    self.execute_remote_command("#{MONIT} stop -g #{watch}", remote_ip, remote_key)
-    self.execute_remote_command("#{MONIT} unmonitor -g #{watch}", remote_ip, remote_key)
-  end
-
-  def self.shutdown(remote_ip=nil, remote_key=nil)
-    self.execute_remote_command("#{MONIT} quit", remote_ip, remote_key)
+  def self.stop(watch)
+    Dir.glob("/etc/monit/conf.d/appscale-#{watch}-*.cfg") { |config_file|
```
menivaitsi: Is there a constant for the config directory that you can use? If not, please define a new one.

PR #1938, opened 2016-02-11
Allow ssh-copy-ip to work at boot time.

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1112/

PR #1947, opened 2016-02-17
Don't start the ZK GC thread from the groomer

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1126/

PR #1953, opened 2016-02-24
Alternative install for zookeeper removed

lmandres:
The alternative install for ruby-zk no longer seems to be needed.  Running:

```
gem install zookeeper
```

from command line works on Raspberry Pi 2/Zero


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1150/
Thanks, @lmandres

PR #1956, opened 2016-02-25
start cron in docker for faststart.sh

selfcommit:
Cron must be started manually inside of Docker.


cdonati:
Thanks, @selfcommit

PR #1963, opened 2016-03-01
Remove unused function

cdonati:
The child function it calls also no longer exists.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1169/

PR #1979, opened 2016-03-12
Renaming leftover non-existent AC method.

menivaitsi:
This branch fixes a crash while trying to call get_monitors. 
This method has a different name in the current codebase.


menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/1218/

PR #1983, opened 2016-03-17
Don't use killall to stop the controller.

obino:
killall -r will not work if the controller is not started by
start-stop-daemon.


obino:
ocd run https://ocd.appscale.com:8080/job/Daily%20Build/1225/

PR #1985, opened 2016-03-18
Update Go to version 1.4.3

airamrguez:
This pull request updates Go to version 1.4.3 compatible with Appengine SDK 1.8.0.
Despite the fact that the official release of GAE 1.8.0 was tied to Go 1.1.2 I think all Appscalers-Gophers would be grateful to see Go 1.4.3 merged into master. Moreover the current release of Go is 1.6 which seems so far away from the current supported version by Appscale.

### DIFF
--- a/AppDB/datastore_server.py
+++ b/AppDB/datastore_server.py
@@ -835,7 +835,7 @@ class DatastoreDistributed():
     value(s): The string representation of mulitiple properties.
     entity_key: The entity key (full path) used as a means of having a unique
       identifier. This prevents two entities with the same values from
-      colliding. 
+      colliding.
 
     Args:
       index: A datstore_pb.CompositeIndex.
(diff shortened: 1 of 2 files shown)


menivaitsi:
@airamrguez thank you for submitting a Pull Request.
Thanks to contributors like yourself, AppScale gets better and better!

We'll do some testing and reviewing before we merge. :)
CI Trusty build: https://ocd.appscale.com:8080/job/Daily%20Build/1228/

cdonati:
@airamrguez: Would you be willing to remove the binaries from this pull request? I know the go binaries were already present in the repo, but I think it would be better if they were created during the build process.

Also, I think it would be best to create a new branch without these particular commits so that the extra ~35MB doesn't get added to the size of the repo.

I haven't tried building this, so let me know if there would be any problems with adding the process to `debian/appscale_install_functions.sh`.

cdonati:
If the build process takes an unreasonably long time, another option would be to download the go binaries from a trusted source during the AppScale build.

PR #1993, opened 2016-03-24
Go 1.4.3

airamrguez:
@cdonati I've removed the binaries from the previous pull request and added a function to `debian/appscale_install_functions.sh` to build Go from source. I've also modified `debian/appscale_install.sh`. I don't think it takes to long to build Go from source.

I'm submitting this changes into a new pull request because I think is cleaner than the previous one. Sorry for the inconvenience.


cdonati:
Thanks, @airamrguez. I sent you a pull request on your branch to fix some of the issues I was seeing during my attempt to build on Trusty. Let me know if any of my changes don't make sense.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1264/

airamrguez:
By the way, what do you think about adding `export GOPATH=/root/appscale/AppServer/gopath` in the `~/.bashrc` file? It's common to add external Go libraries on your projects and to do that you have to  manually set up `GOPATH` in order to execute `/root/appscale/AppServer/goroot/bin/go get ...`. And also, what do you think about adding `/root/appscale/AppServer/goroot/bin` into the `PATH` variable so that you can simply do `go get ...`?

cdonati:
@airamrguez Would that just be for if you were developing the app from inside the deployment? I'm hesitant to change `PATH` and `GOPATH` for the root user since Go might already be separately installed on the system.

As for the external Go libraries, I'm not sure how that works. Right now, if a user wanted to deploy an app that requires external libraries, do they need to log in to the deployment and manually `go get` those packages in order for their app to work?

PR #2007, opened 2016-04-15
Allow does_user_exist to return exception text

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1329/

PR #2010, opened 2016-04-19
Only include app owner if defined

cdonati:
Also, rename `get_app_data` and change description.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1344/

PR #2013, opened 2016-04-20
Use consistently ' for @options['

obino:
The mixed usage of @options[' and @options[" makes it difficult to search
for options.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1352/

PR #2016, opened 2016-04-22
Restart DjinnServer if it raises an exception

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1378/

PR #2025, opened 2016-05-01
Don't show error when some file don't exists.

obino:
Current master has these errors in the logs early on in the start process:

D, [2016-05-01T04:00:20.083682 #2696] DEBUG -- : Running bash /root/appscale/firewall.conf
cat: /etc/appscale/masters: No such file or directory
cat: /etc/appscale/slaves: No such file or directory

this patch cleans them up.


obino:
ocd run https://ocd.appscale.com:8080/job/Daily%20Build/1395/

PR #2030, opened 2016-05-04
Install Cassandra to /opt/cassandra

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1417/

AppController/lib/cassandra_helper.rb
```diff
@@ -17,14 +17,18 @@
 
 
 # A String that indicates where we install Cassandra on this machine.
-CASSANDRA_DIR = "#{APPSCALE_HOME}/AppDB/cassandra"
+CASSANDRA_DIR = "/opt/cassandra"
```
menivaitsi: Is this dir persisted in cloud mode?
I think it should be something like /opt/appscale/cassandra-env to make sure we store in external volumes.

cdonati: Do we need to persist the Cassandra binaries? I don't know why it should be different than the other system packages.

menivaitsi: @cdonati ok, that makes sense. Thanks!

PR #2039, opened 2016-05-15
Validate ssl cert with pyOpenSSL as a fallback

cdonati:
Versions of Python before 2.7.9 don't have native client support for SNI, so this is a quick fix that opens an additional connection with pyOpenSSL to get the right certificate.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1464/

menivaitsi:
👍

obino:
Is this a problem also with the SDK and GAE?

cdonati:
It's just a problem with the SDK. I think they want to support Python 2.6, and I they probably want to avoid including a large external dependency like pyOpenSSL, so it will probably be an issue there for awhile.

PR #2041, opened 2016-05-24
Datastore upgrade script and creating monit config file for required services.

tmarballi:


AppDB/zkappscale/zktransaction.py
```diff
@@ -118,7 +118,7 @@ class ZKTransaction:
   """
 
   # The number of times we should retry ZooKeeper operations, by default.
-  DEFAULT_NUM_RETRIES = 0
+  DEFAULT_NUM_RETRIES = 5
```
cdonati: What was the reason for this change?

lib/monit_interface.py
```diff
@@ -127,3 +134,46 @@ def restart(watch):
 
   logging.info("Restarting watch {0}".format(watch))
   return run_with_retry([MONIT, 'restart', '-g', watch])
+
+def create_monit_config_file(watch, start_cmd, stop_cmd, ports, match_cmd,
```
cdonati: Is there a reason you didn't use `create_config_file` in `lib/monit_app_configuration.py`?

lib/templates/monit_template_for_upgrade.conf
```diff
@@ -0,0 +1,6 @@
+check process {0}-{3} matching "{4}"
```
cdonati: I think named parameters would be clearer here.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+PID_FILE = "/var/appscale/appscale-cassandra.pid"
+
+# The default port to connect to ZooKeeper.
+ZK_PORT = ":2181"
+
+# The location on the local filesystem where we should store ZooKeeper data.
+ZK_DATA_LOCATION = "/opt/appscale/zookeeper"
+
+# The maximum number of entities to be returned from the query.
+_MAX_ENTITIES = 1000000
```
cdonati: I think you'll want to fetch 100 entities at a time. Cassandra may not be able to return this many entity rows before the query times out.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+
+def ensure_app_is_not_running():
+    """ Ensures AppScale is not running as this is an offline script. """
```
cdonati: Though I personally prefer four-space indentation, you'll probably want to use two spaces to match the rest of the codebase.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    print ("Ensure AppScale is not currently running...")
+    appscale_running = subprocess.call(['service', CONTROLLER_SERVICE, 'status']) == 0
+    if appscale_running:
+        print ("AppScale is running, please shut it down and try again.")
+        exit(1)
+
+def start_cassandra():
+    """ Creates a monit configuration file and prompts Monit to start Cassandra. """
+    print ("Starting Cassandra...")
```
cdonati: Since the `logging` module is more configurable, we typically use it instead of print statements.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    if not monit_interface.start(watch):
+        print ("Unable to start Cassandra.")
+        exit(1)
+    else:
+        print ("Successfully started Cassandra.")
+
+def start_zookeeper():
+    """ Creates a monit configuration file and prompts Monit to start ZooKeeper. """
+    print ("Starting ZooKeeper...")
+    os.system("rm -rfv /var/lib/zookeeper")
```
cdonati: It's best to avoid making system calls when necessary. In this case you can use `os.remove`. If you must make system calls, the `subprocess` module is recommended.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+            if not os.system("/usr/sbin/service " + "zookeeper" + " init"):
+                print ("Failed to start zookeeper!")
+                exit(1)
+    os.system("ln -sfv /etc/zookeeper/conf/myid " + ZK_DATA_LOCATION + "/myid")
+    watch = "zookeeper"
+    start_cmd = "/usr/sbin/service " + "zookeeper" + " start"
+    stop_cmd = "/usr/sbin/service " + "zookeeper" + " stop"
+    match_cmd = "org.apache.zookeeper.server.quorum.QuorumPeerMain"
+    monit_interface.create_monit_config_file(watch, start_cmd, stop_cmd,
+        ports=[2181], match_cmd=match_cmd)
```
cdonati: Use a constant for the port to reduce repetition.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    if not monit_interface.start(watch):
+        print ("Unable to start ZooKeeper")
+        exit(1)
```
cdonati: While they are pretty similar, `sys.exit` is recommended for use in programs while `exit` is mainly meant as a convenience function for the interactive shell.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+def validate_and_update_entities(datastore, ds_distributed):
+    """ Returns the SOAP server accessor to deal with application and users.
+    Args:
+        datastore: A reference to the batch datastore interface.
+    Returns:
+        A soap server accessor.
+    """
+    # Fetch all entities from the APP_ENTITY_TABLE.
+    try:
+        current_entities = datastore.range_query(APP_ENTITY_TABLE, APP_ENTITY_SCHEMA, "", "", _MAX_ENTITIES)
```
cdonati: You'll want to do a loop here because you don't want to try fetching all of the entity rows at the same time. At the end of each loop, modify the `start_row` with the last key you fetched, and make sure you use `start_inclusive=True`.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    except Exception as ex:
+        print ("Exception on range query: {}".format(str(ex)))
+        exit(1)
+
+    # Remove the tombstoned entities from the list.
+    tombstoneless_entities = ds_distributed.remove_tombstoned_entities(current_entities)
+    app_ids = get_app_ids_from_entities(tombstoneless_entities)
+    entities_to_validate  = tombstoneless_entities
```
cdonati: Is this extra assignment needed?

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # Validate the entities grouped by app id.
+    for app_id in app_ids:
```
cdonati: The entities are not grouped by app ID here since you are always passing the same `entities_to_validate` var. To make this more efficient, I would change `validated_list_result` to extract the app ID. You don't want to process each entity more than once, and you definitely don't want to try looking up a transaction for a given app with an entity from another app.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+        validated_entities_for_app = ds_distributed.validated_list_result(app_id, entities_to_validate)
+        entities_to_validate = validated_entities_for_app
```
cdonati: Since you'll be processing entities in batches, you'll probably want to call an update function after validating each batch (instead of passing all the entities back to the parent function).

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+
+            if is_entity_same(entity,entity_key, valid_entity) and is_txn_id_same(entity, entity_key, valid_entity):
+                continue
+
+            if not entity_key in row_keys_to_update:
+                row_keys_to_update.append(entity_key)
+                cell_value = {}
+                cell_value[APP_ENTITY_SCHEMA[0]] = valid_entity[entity_key][APP_ENTITY_SCHEMA[0]]
+                cell_value[APP_ENTITY_SCHEMA[1]] = valid_entity[entity_key][APP_ENTITY_SCHEMA[1]]
+                cell_values_to_update[entity_key] = cell_value
```
cdonati: You can use a dictionary literal notation here:
```
cell_values_to_update[entity_key] = {
  key1: value,
  key2: value
}
```

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    return valid_row_keys
+
+def get_zk_locations_string(zk_location_ips):
+    """ Generates a ZooKeeper IP locations string.
+    Args:
+        zk_location_ips: A list of ZooKeeper IP locations.
+    Returns:
+        A string of ZooKeeper IP locations
+    """
+    return (ZK_PORT + ",").join(zk_location_ips) + ZK_PORT
```
cdonati: You'll probably want to use an integer for ZK_PORT so that you can format it with a colon when necessary and use it in cases when it doesn't need a colon.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,311 @@
...
+    ds_distributed = get_datastore_distributed(datastore, zk_location_ips)
+    current_entities, tombstoneless_entities , validated_entities = \
+        validate_and_update_entities(datastore, ds_distributed)
+
+    # Delete the tombstoned entities from the APP_ENTITY_TABLE.
+    delete_tombstoned_entities_from_table(datastore, current_entities, validated_entities)
+
+    # Update and replace each non valid entity with the right value from the journal.
+    update_entities_table(datastore, tombstoneless_entities, validated_entities)
+
```
cdonati: Remove extra empty lines.

lib/monit_interface.py
tmarballi: Yes, the way the config file created for apps is somehow different from the ones created for cassandra and zookeeper. In Djinn it does it differently for them, so I did the same as that! Thats why I added a new template too!

cdonati: Okay, I would recommend trying to adjust the behavior of the function with parameters if possible. If not, I would rename this function so that it's not used as a general way to create a monit config file.

AppDB/zkappscale/zktransaction.py
tmarballi: You were right, I think it was just waiting on the ZooKeeper to start. Maybe it was too quick before and I am now sleeping a bit to make sure ZK and Cassandra are up before making the call to ZKTransaction. Will remove this!

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+import appscale_datastore_batch
+import datastore_server
+from dbconstants import *
+from zkappscale import zktransaction as zk
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "../AppServer"))
+from google.appengine.datastore import entity_pb
+
+# The location on the local filesystem where Cassandra is installed.
+CASSANDRA_DIR = "/opt/cassandra"
```
menivaitsi: Is this defined somewhere in AppDB?

Please re-use constants for DB and ZK if they already exist.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+# The location of the Cassandra binary on the local filesystem.
+CASSANDRA_EXECUTABLE = CASSANDRA_DIR + "/cassandra/bin/cassandra"
+
+PID_FILE = "/var/appscale/appscale-cassandra.pid"
+
+# The default port to connect to Cassandra.
+CASSANDRA_PORT = 9999
+
+# The default port to connect to ZooKeeper.
+ZOOKEEPER_PORT = 2181
```
menivaitsi: You can use the constant defined in AppDB/zkappscale/zktranscation.py instead of creating a new one.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  if not monit_interface.start(watch):
+    logging.info("Unable to start Cassandra.")
+    sys.exit(1)
+  else:
+    logging.info("Successfully started Cassandra.")
+
+def start_zookeeper():
+  """ Creates a monit configuration file and prompts Monit to start ZooKeeper. """
+  logging.info("Starting ZooKeeper...")
+  os.system("rm -rfv /var/lib/zookeeper")
```
menivaitsi: Look [here](https://docs.python.org/2/library/os.html) for removing files, creating dirs, etc.

As Chris mentioned, if a function is not available for what you want to do it's better to use subprocess.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+    logging.info("Unable to start ZooKeeper")
+    sys.exit(1)
+  else:
+    logging.info("Successfully started ZooKeeper")
+
+def validate_and_update_entities(datastore, ds_distributed):
+  """ Returns the SOAP server accessor to deal with application and users.
+  Args:
+    datastore: A reference to the batch datastore interface.
+
```
menivaitsi: Remove empty line.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+    current_entities: A list of entities currently in the APP_ENTITY_TABLE.
+    validated_entities: A list of validated entities to compare from.
+  """
+  valid_entity_keys = get_valid_entity_keys(validated_entities)
+  invalid_entity_keys = compare_and_get_invalid_keys(current_entities, valid_entity_keys)
+  try:
+    datastore.batch_delete(APP_ENTITY_TABLE, invalid_entity_keys)
+    logging.info("Batch delete was successful")
+  except AppScaleDBConnectionError as error:
+    logging.info("Exception on batch delete: {}".format(str(error)))
```
menivaitsi: Make this a logging.exception().

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  Returns:
+    A list of current entities in APP_ENTITY_TABLE, entities without tombstoned
+    values and validated entities.
+  """
+  # Fetch all entities from the APP_ENTITY_TABLE.
+  try:
+    current_entities = datastore.range_query(APP_ENTITY_TABLE, APP_ENTITY_SCHEMA, "", "", _MAX_ENTITIES)
+
+  except Exception as ex:
+    logging.info("Exception on range query: {}".format(str(ex)))
```
menivaitsi: Use logging.exception()

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  os.system("ln -sfv /etc/zookeeper/conf/myid " + ZK_DATA_LOCATION + "/myid")
+  watch = "zookeeper"
+  start_cmd = "/usr/sbin/service " + "zookeeper" + " start"
+  stop_cmd = "/usr/sbin/service " + "zookeeper" + " stop"
+  match_cmd = "org.apache.zookeeper.server.quorum.QuorumPeerMain"
+  monit_app_configuration.create_config_file(watch, start_cmd, stop_cmd,
+    ports=[ZOOKEEPER_PORT], upgrade_flag=True, match_cmd=match_cmd)
+
+  if not monit_interface.start(watch):
+    logging.info("Unable to start ZooKeeper")
```
menivaitsi: This should be logging.error()

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+
+  if not os.path.isdir(ZK_DATA_LOCATION):
+    logging.info("Initializing ZooKeeper")
+    os.system("/usr/sbin/service " + "zookeeper" + " stop")
+    os.system("mkdir -pv " + ZK_DATA_LOCATION)
+    os.system("chown -Rv zookeeper:zookeeper " + ZK_DATA_LOCATION)
+
+  if zk_server == "zookeeper-server":
+    if not os.system("/usr/sbin/service " + "zookeeper" + " init"):
+      logging.info("Failed to start zookeeper!")
```
menivaitsi: Use logging.error() here.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+        cell_value = {}
+        cell_value[APP_ENTITY_SCHEMA[0]] = valid_entity[entity_key][APP_ENTITY_SCHEMA[0]]
+        cell_value[APP_ENTITY_SCHEMA[1]] = valid_entity[entity_key][APP_ENTITY_SCHEMA[1]]
+        cell_values_to_update[entity_key] = cell_value
+
+  try:
+    datastore.batch_put_entity(APP_ENTITY_TABLE, row_keys_to_update, APP_ENTITY_SCHEMA, cell_values_to_update)
+    logging.info("Batch put entities was successful")
+  except  AppScaleDBConnectionError as error:
+    logging.info("Exception on batch put entity: {}".format(str(error)))
```
menivaitsi: Make a pass on these and change the log level (info, debug, warning, error, exception) when appropriate.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+
+def is_txn_id_same(entity, key, valid_entity):
+  """ Compares if the transaction id is the same for the current and
+  validated entity.
+  Args:
+    entity: Current tombstonedless entity.
+    key: A key to look for in the validated entity.
+    valid_entity: A validated entity to compare with.
+  Returns:
+    True if transaction id is the same.
```
menivaitsi: Add a "False otherwise" at the end of this sentence.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+
+def is_entity_same(entity, key, valid_entity):
+  """ Compares if the transaction id is the same for the current and
+  validated entity.
+  Args:
+    entity: Current tombstonedless entity.
+    key: A key to look for in the validated entity.
+    valid_entity: A validated entity to compare with.
+  Returns:
+    True is the encoded entity is the same.
```
menivaitsi: Same here.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+
+def compare_and_get_invalid_keys(current_entities, valid_entity_keys):
+  """ Looks for the current entities keys in a list of valid entity keys
+  and returns the invalid keys.
+  Args:
+    current_entities: A list of entities currently in APP_ENTITY_TABLE.
+    key: A key to look for in the validated entity.
+    valid_entity_keys: A list valid entity keys to compare from.
+  Returns:
+    A list of invalid entity keys
```
menivaitsi: Add missing period.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  """
+  invalid_entity_keys = []
+  for entity in current_entities:
+    key = entity.keys()[0]
+    if key not in valid_entity_keys:
```
menivaitsi: Can these two ifs be combined? Something like:
> if key not in valid_entity_keys and key not in invalid_entity_keys:

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+def get_valid_entity_keys(validated_entities):
+  """ Loops through the validated entities and creates a list of valid
+  entity keys.
+  Args:
+    validated_entities: A list of validated entities.
+  Returns:
+    A list of valid entity keys.
+  """
+  valid_row_keys = []
+  # Get all the keys from the validated entities
```
menivaitsi: Add missing period.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+    if valid_key not in valid_row_keys:
+      valid_row_keys.append(valid_key)
+  return valid_row_keys
+
+def get_zk_locations_string(zk_location_ips):
+  """ Generates a ZooKeeper IP locations string.
+  Args:
+    zk_location_ips: A list of ZooKeeper IP locations.
+  Returns:
+    A string of ZooKeeper IP locations
```
menivaitsi: Same here.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  """
+  return (ZK_PORT_WITH_COLON + ",").join(zk_location_ips) + ZK_PORT_WITH_COLON
+
+def get_app_ids_from_entities(entities_without_tombstoned_values):
+  """ Returns a list of app_ids stripped of the entity keys.
+  Args:
+    entities_without_tombstoned_values: A list of entities which do not have
+    any tombstoned values.
+  Returns:
+    A list of app_ids
```
menivaitsi: Same here.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  datastore_distributed = datastore_server.DatastoreDistributed(datastore, zookeeper=zookeeper)
+  return datastore_distributed
+
+if __name__ == "__main__":
+  total = len(sys.argv)
+  if not total > 1:
+    sys.exit(1)
+
+  zk_location_ips = []
+  for i in range(total):
```
menivaitsi: Avoid single letter variables. Use something like index.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+    zk_location_ips: A list of ZooKeeper location ips.
+  """
+  zookeeper_locations = get_zk_locations_string(zk_location_ips)
+  zookeeper = zk.ZKTransaction(host=zookeeper_locations)
+  datastore_distributed = datastore_server.DatastoreDistributed(datastore, zookeeper=zookeeper)
+  return datastore_distributed
+
+if __name__ == "__main__":
+  total = len(sys.argv)
+  if not total > 1:
```
menivaitsi: Change this to:
> if total < 1:

Using the positive condition is easier to read.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+    datastore: A reference to the batch datastore interface.
+    zk_location_ips: A list of ZooKeeper location ips.
+  """
+  zookeeper_locations = get_zk_locations_string(zk_location_ips)
+  zookeeper = zk.ZKTransaction(host=zookeeper_locations)
+  datastore_distributed = datastore_server.DatastoreDistributed(datastore, zookeeper=zookeeper)
+  return datastore_distributed
+
+if __name__ == "__main__":
+  total = len(sys.argv)
```
menivaitsi: This var name is not very intuitive. Change it to something like _args_?

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+  # This datastore upgrade script is to be run offline, so make sure
+  # appscale is not up while running this script.
+  ensure_app_is_not_running()
+
+  # Start Cassandra and ZooKeeper.
+  start_cassandra()
+  start_zookeeper()
+
+  # Sleep time for Cassandra and ZooKeeper to be started.
+  time.sleep(20)
```
menivaitsi: Define 20 as a constant up at the top, please.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,330 @@
...
+sys.path.append(os.path.join(os.path.dirname(__file__), "../AppServer"))
+from google.appengine.datastore import entity_pb
+
+# The location on the local filesystem where Cassandra is installed.
+CASSANDRA_DIR = "/opt/cassandra"
+
+# The location of the Cassandra binary on the local filesystem.
+CASSANDRA_EXECUTABLE = CASSANDRA_DIR + "/cassandra/bin/cassandra"
+
+PID_FILE = "/var/appscale/appscale-cassandra.pid"
```
menivaitsi: Add comment for this constant.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1505/

lib/monit_interface.py
```diff
@@ -17,6 +17,13 @@
 
 NUM_RETRIES = 10
 
+# Template used for creating monit coniguration files.
+TEMPLATE_LOCATION = os.path.join(os.path.dirname(__file__)) +\
+                    "/templates/monit_template_for_upgrade.conf"
+
+# The directory used when storing a service's config file.
+MONIT_CONFIG_DIR = '/etc/monit/conf.d'
```
cdonati: Are these changes still needed?

tmarballi: No they are not! I will clean them up!

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,324 @@
...
+    datastore: A reference to the batch datastore interface.
+  Returns:
+    True: if deleting the entity was successful.
+    False: if an exception was encountered while deleting the entity.
+  """
+  try:
+    datastore.batch_delete(APP_ENTITY_TABLE, [key])
+  except AppScaleDBConnectionError as connection_error:
+    logger.error("Error deleting key {0}: {1}".format(key, connection_error))
+    return False
```
cdonati: I think it would be better to let the exception propagate here rather than return false. As it is now, the return value is ignored in the validate function, so the deletion will silently fail. It would be better to know there is a failure.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,324 @@
...
+  version = entity[key][APP_ENTITY_SCHEMA[1]]
+  if one_entity == datastore_server.TOMBSTONE:
+    return delete_entity_from_table(key, datastore)
+
+  entity_proto = entity_pb.EntityProto(one_entity)
+  app_id = entity_proto.key().app()
+  validated_entity = ds_distributed.validated_result(app_id, entity)
+
+  if not validated_entity:
+    return True
```
cdonati: If `validated_entity` is an empty dict, that means that there's no valid transaction for the entity. It should be removed from the entities table.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,324 @@
...
+    datastore: A reference to the batch datastore interface.
+  Returns:
+    True: if updating entity was successful.
+    False: if an exception was encountered while updating the entity.
+  """
+  cell_values_to_update = {}
+  cell_values_to_update[key] = {
+    APP_ENTITY_SCHEMA[0]: validated_entity[key][APP_ENTITY_SCHEMA[0]],
+    APP_ENTITY_SCHEMA[1]: validated_entity[key][APP_ENTITY_SCHEMA[1]]
+  }
```
cdonati: You don't need to create a new dict here. You can just pass `validated_entity[key]` to `batch_put_entity` instead of `cell_values_to_update`.

AppDB/dbconstants.py
```diff
@@ -33,6 +33,7 @@
 APP_ENTITY_TABLE = "ENTITIES__"
 APP_KIND_TABLE = "KINDS__"
 METADATA_TABLE = "METADATA__"
+DATASTORE_VERSION_TABLE = "DATASTORE_VERSION__"
```
menivaitsi: Can we make this DATASTORE_METADATA_TABLE? It seems more generic and doesn't conflict with anything that's there or that we have in store for the future.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+import logging
+import os
+import subprocess
+import sys
+import time
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../lib'))
+import appscale_info
+from constants import CONTROLLER_SERVICE
+from constants import APPSCALE_HOME
```
menivaitsi: Move this import 1 line up to maintain alphabetical order.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+from zkappscale import zktransaction as zk
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "../AppServer"))
+from google.appengine.api import datastore_errors
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "../InfrastructureManager"))
+from utils import utils
+
+# Default ZooKeeper port formatted with a colon.
+ZK_PORT_WITH_COLON = ":" + str(zk.DEFAULT_PORT)
```
menivaitsi: Please keep the port separate from ":" and format accordingly below. It's cleaner imho.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+  return datastore_distributed, zookeeper
+
+def get_zk_locations_string(zk_location_ips):
+  """ Generates a ZooKeeper IP locations string.
+  Args:
+    zk_location_ips: A list of ZooKeeper IP locations.
+  Returns:
+    A string of ZooKeeper IP locations.
+  """
+  return (ZK_PORT_WITH_COLON + ",").join(zk_location_ips) + ZK_PORT_WITH_COLON
```
menivaitsi: You can change this into something like: 
`(':' + ZK_PORT + ",").join(zk_location_ips) + ZK_PORT`

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+# Monit watch name for Cassandra.
+CASSANDRA_WATCH_NAME = "cassandra"
+
+# Monit watch name for ZooKeeper.
+ZK_WATCH_NAME = "zookeeper"
+
+# Datastore version from AppScale 3.0.0 onwards.
+VERSION_1 = "1.0"
+
+# Success status
```
menivaitsi: Add period at the end.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+    db_ips: A list of database node IPs to start Cassandra on.
+    keyname: A string containing the deployment's keyname.
+  """
+  logging.info("Starting Cassandra...")
+  for ip in db_ips:
+    start_service_cmd = START_SERVICE_SCRIPT + CASSANDRA_WATCH_NAME
+    cmd_status = utils.ssh(ip, keyname, start_service_cmd)
+    start_cassandra_ip = START_CASSANDRA + ":" + ip
+    if not cmd_status == 0:
+      logging.error("Monit was unable to start Cassandra.")
```
menivaitsi: What happens if one or more of the other cassandra peers are up at this point? Should we bring them down before exiting?

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+      status_dict[start_cassandra_ip] = FAILURE
+      sys.exit(1)
+    logging.info("Successfully started Cassandra.")
+    status_dict[start_cassandra_ip] = SUCCESS
+
+def start_zookeeper(status_dict, zk_ips, keyname):
+  """ Creates a monit configuration file and prompts Monit to start Cassandra.
```
menivaitsi: "Cassandra" -> "Zookeeper"

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+      zk_ips: A list of zookeeper node IPs to start ZooKeeper on.
+      keyname: A string containing the deployment's keyname.
+    """
+  logging.info("Starting ZooKeeper...")
+  for ip in zk_ips:
+    start_service_cmd = START_SERVICE_SCRIPT + ZK_WATCH_NAME
+    cmd_status = utils.ssh(ip, keyname, start_service_cmd)
+    start_zookeeper_ip = START_ZOOKEEPER + ":" + ip
+    if not cmd_status == 0:
+      logging.error("Monit was unable to start ZooKeeper.")
```
menivaitsi: Same here. We should cleanup before exiting.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+      status_dict[start_zookeeper_ip] = FAILURE
+      sys.exit(1)
+    logging.info("Successfully started ZooKeeper.")
+    status_dict[start_zookeeper_ip] = SUCCESS
+
+def get_datastore():
+  """ Returns a reference to the batch datastore interface. Validates where
+  the <datastore>_interface.py is and adds that path to the system path.
```
menivaitsi: <datastore> -> <database>

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+
+def update_entity_in_table(key, validated_entity, datastore):
+  """ Updates the APP_ENTITY_TABLE with the valid entity.
+  Args:
+    key: A str representing the row key to update in the table.
+    validated_entity: A validated entity which needs to be updated in place of
+    the current entity.
+    datastore: A reference to the batch datastore interface.
+  """
+  datastore.batch_put_entity(APP_ENTITY_TABLE, [key], APP_ENTITY_SCHEMA, validated_entity[key])
```
menivaitsi: Does the datastore batch_put return a list of keys successfully stored or true/false?

menivaitsi: I think it's best to add a Returns section regardless for functions like this one.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+    status_dict: A dictionary to record the status of the executed process.
+    keyname: A string containing the deployment's keyname.
+  """
+  logging.info("Stopping Cassandra...")
+  for ip in db_ips:
+    stop_service_cmd = STOP_SERVICE_SCRIPT + CASSANDRA_WATCH_NAME
+    cmd_status = utils.ssh(ip, keyname, stop_service_cmd)
+    stop_cassandra_ip = STOP_CASSANDRA + ":" + ip
+    if not cmd_status == 0:
+      logging.error("Monit was unable to stop Cassandra.")
```
menivaitsi: Shouldn't we try to stop all Cassandra peers regardless of failure on any one of the db nodes?

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,422 @@
...
+    status_dict: A dictionary to record the status of the executed process.
+    keyname: A string containing the deployment's keyname.
+  """
+  logging.info("Stopping ZooKeeper...")
+  for ip in zk_ips:
+    stop_service_cmd = STOP_SERVICE_SCRIPT + ZK_WATCH_NAME
+    cmd_status = utils.ssh(ip, keyname, stop_service_cmd)
+    stop_zookeeper_ip = STOP_ZOO_KEEPER + ":" + ip
+    if not cmd_status == 0:
+      logging.error("Monit was unable to stop ZooKeeper.")
```
menivaitsi: Same here.

tmarballi: I changed this to stop all ZooKeeper nodes regardless of error in stopping any. It is not showing up on the diff because its two lines below.

scripts/datastore_upgrade.py
tmarballi: I changed this to stop all Cassandra nodes regardless of error in stopping any. It is not showing up on the diff because its two lines below this one.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,445 @@
...
+
+  # If validating and updating entities logged an error in the status dict,
+  # return from this script.
+  if VALIDATE_ENTITIES in status_dict:
+    return
+
+  status_dict[VALIDATE_ENTITIES] = SUCCESS
+  logging.info("Updated invalid entities and deleted tombstoned entities.")
+
+  # Create a new table if required to store App Entity data version.
```
menivaitsi: Please remove "App Entity".

scripts/monit_start_service.py
```diff
@@ -0,0 +1,68 @@
...
+
+# The location on the local file system where we write the process ID
+# that Cassandra runs on.
+PID_FILE = "/var/appscale/appscale-cassandra.pid"
+
+# The default port to connect to Cassandra.
+CASSANDRA_PORT = 9999
+
+# Monit watch name for Cassandra.
+CASSANDRA_WATCH_NAME = "cassandra"
```
menivaitsi: Please reuse the watch name constants defined in datastore_upgrade.py if possible.

scripts/monit_start_service.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Monit watch name for ZooKeeper.
+ZK_WATCH_NAME = "zookeeper"
```
menivaitsi: Same here.

scripts/monit_start_service.py
```diff
@@ -0,0 +1,68 @@
...
+
+if __name__ == "__main__":
+  args_length = len(sys.argv)
+  if args_length < 2:
+    sys.exit(1)
+
+  service_name = (str(sys.argv[1]))
+  start_service(service_name)
+
+
```
menivaitsi: Remove empty line.

scripts/monit_start_service.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
+#
```
menivaitsi: Remove # from empty line.

scripts/monit_stop_service.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+#
```
menivaitsi: Remove # from empty line.

scripts/monit_stop_service.py
```diff
@@ -0,0 +1,26 @@
...
+
+if __name__ == "__main__":
+  args_length = len(sys.argv)
+  if args_length < 2:
+    sys.exit(1)
+
+  service_name = (str(sys.argv[1]))
+  stop_service(service_name)
+
+
```
menivaitsi: Remove empty line.

scripts/upgrade.py
```diff
@@ -0,0 +1,71 @@
+import json
```
menivaitsi: Add missing docstring.

scripts/upgrade.py
```diff
@@ -0,0 +1,71 @@
...
+sys.path.append(os.path.join(os.path.dirname(__file__), '../AppDB'))
+from dbconstants import *
+
+# Version for which Datastore upgrade is needed.
+UPGRADE_NEEDED_VERSION = "3.0.0"
+
+# JSON file location to record the status of the processes.
+UPGRADE_JSON_FILE = '/var/log/appscale/upgrade-status-'
+
+# Data upgrade status key
```
menivaitsi: Add "." at the end.

scripts/upgrade.py
```diff
@@ -0,0 +1,71 @@
...
+DATA_UPGRADE = 'Data Upgrade'
+
+# .JSON file extention
```
menivaitsi: Same here.

scripts/upgrade.py
```diff
@@ -0,0 +1,71 @@
...
+JSON_FILE_EXTENTION = ".json"
+
+def is_data_upgrade_needed(version):
+  """Checks if for this version of AppScale datastore upgrade is needed."""
```
menivaitsi: Please add Args/Returns sections.

scripts/upgrade.py
```diff
@@ -0,0 +1,71 @@
...
+  if version == UPGRADE_NEEDED_VERSION:
+    return True
+  return False
+
+def write_to_json_file(data, timestamp):
+  upgrade_status_file = UPGRADE_JSON_FILE + timestamp + JSON_FILE_EXTENTION
```
menivaitsi: Add missing docstring.

bootstrap_force_upgrade.sh
```diff
@@ -0,0 +1,77 @@
...
+done
+
+echo "Copying database files needed from /etc/appscale to a temporary location /root/database-temp."
+mkdir -p /root/database-temp
+cp -r /etc/appscale/. /root/database-temp
+cd /root/database-temp ; rm -r 2.8.0; rm -r certs; rm environment.yaml; rm home; rm VERSION
+
+echo "Running the bootstrap script with a --force-upgrade"
+
+(cd; cd appscale; bash bootstrap.sh --force-upgrade)
```
obino: Instead of cd; cd appscale, you could do cd ~/appscale

bootstrap_force_upgrade.sh
```diff
@@ -0,0 +1,77 @@
...
+                shift
+                continue
+        fi
+        usage
+done
+
+echo "Copying database files needed from /etc/appscale to a temporary location /root/database-temp."
+mkdir -p /root/database-temp
+cp -r /etc/appscale/. /root/database-temp
+cd /root/database-temp ; rm -r 2.8.0; rm -r certs; rm environment.yaml; rm home; rm VERSION
```
obino: you could do it in a subshell, like you did few lines down (cd /root/database-temp; rm ...)

Also, you can also use mktemp (something like TEMP_DIR=mktemp and then use it instead of /root/database-temp)

bootstrap_force_upgrade.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "Running the bootstrap script with a --force-upgrade"
+
+(cd; cd appscale; bash bootstrap.sh --force-upgrade)
+if [ $? -gt 0 ]; then
+        echo "Running bootstrap script failed!"
+        exit 1
+fi
+
+echo "Coping database files back from the temporary location to /etc/appscale."
+cd; cp -r /root/database-temp/. /etc/appscale
```
obino: If you put the above cd /root/database-temp in a subshell, you do not need cd here.

bootstrap_force_upgrade.sh
```diff
@@ -0,0 +1,77 @@
...
+                index=0
+                declare -i index
+                if [ "$line" = "$LOCAL_IP" ]; then
+                        LOCAL_TOKEN=""
+                fi
+                index=index+1
+        done
+}
+
+get_local_token
```
obino: I think this is no doing anything, is that right?

bootstrap_force_upgrade.sh
```diff
@@ -0,0 +1,77 @@
...
+
+cd /root/appscale/; cd AppDB/cassandra_env/templates/
```
obino: This can be combined in a single cd.

obino:
:+1:

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,475 @@
...
+  # appscale is not up while running this script.
+  ensure_app_is_not_running()
+
+  # Start Cassandra and ZooKeeper.
+  start_cassandra(status_dict, db_ips, master_ip, keyname)
+  start_zookeeper(status_dict, zk_ips, master_ip, keyname)
+
+  if not all_services_started(status_dict):
+    stop_cassandra(db_ips, status_dict, keyname)
+    stop_zookeeper(zk_ips,status_dict, keyname)
```
cdonati: It looks like both functions require an additional `master_ip` argument.

scripts/upgrade.py
```diff
@@ -0,0 +1,87 @@
...
+      available_version = str(sys.argv[index])
+      continue
+    if index == 2:
+      keyname = str(sys.argv[index])
+      continue
+    if index == 3:
+      timestamp = str(sys.argv[index])
+      continue
+    if (str(sys.argv[index]) == ("--master")):
+      master_ip_arg = str(sys.argv[index+1])
```
cdonati: I would use the `argparse` library and make this a required parameter (along with the zk and db ips). Otherwise, they can be undefined when referenced below.

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,475 @@
...
+      start_service_cmd = START_SERVICE_SCRIPT + CASSANDRA_WATCH_NAME
+      cmd_status = utils.ssh(ip, keyname, start_service_cmd)
+
+    start_cassandra_ip = START_CASSANDRA + ":" + ip
+    if not cmd_status == 0:
+      logging.error("Monit was unable to start Cassandra.")
+      status_dict[start_cassandra_ip] = FAILURE
+      continue
+    logging.info("Successfully started Cassandra.")
+    status_dict[start_cassandra_ip] = SUCCESS
```
cdonati: This will overwrite the failure status if a subsequent node is successful. The function should probably throw an exception if any node fails to start.

cdonati: Nevermind, I missed the reference to the specific IP.

scripts/upgrade.py
```diff
@@ -45,43 +46,37 @@ def write_to_json_file(data, timestamp):
...
+def init_parser():
+  """ Initializes the command line argument parser.
+    Returns:
+      A parser object.
+  """
+  parser = argparse.ArgumentParser(
+    description='Checks if any upgrade is required and runs the script for the process.')
+  parser.add_argument('version', type=str, help="available upgrade version")
+  parser.add_argument('keyname', type=str, help="keyname")
+  parser.add_argument('timestamp', type=str, help="timestamp to attach to the status file")
```
menivaitsi: Use one type of quotes (single, double) in all of these statements please.

scripts/upgrade.py
```diff
@@ -8,6 +8,8 @@
 
 import datastore_upgrade
 
+LATEST_VERSION = 'Latest-version'
```
menivaitsi: What is this used for?

menivaitsi: Are we ever going to overwrite this value?

menivaitsi: If not, then why have this key?

menivaitsi: You can use this for checking the status in a dict like the following:

`{'status': 'success', 'version': LATEST_VERSION, 'message': 'AppScale is currently at its latest version'}
`
or
`{'status': 'warning', 'version': LATEST_VERSION, 'message': 'AppScale is currently at its latest version'}
`

scripts/datastore_upgrade.py
```diff
@@ -0,0 +1,476 @@
...
+  # appscale is not up while running this script.
+  ensure_app_is_not_running()
+
+  # Start Cassandra and ZooKeeper.
+  start_cassandra(status_dict, db_ips, master_ip, keyname)
+  start_zookeeper(status_dict, zk_ips, master_ip, keyname)
+
+
+  # Sleep time for Cassandra and ZooKeeper to be started.
+  time.sleep(SLEEP_TIME)
```
cdonati: Instead of sleeping an arbitrary length here, I think it would be more reliable to check Cassandra's readiness directly:
```
while subprocess.call([NODE_TOOL, 'status']) != 0:
  time.sleep(1)
```
You can import `NODE_TOOL` from `cassandra_interface`.
For zookeeper, you can do the same with `/usr/share/zookeeper/bin/zkServer.sh status`.

scripts/upgrade.py
```diff
@@ -0,0 +1,87 @@
...
+import os
+import sys
+import yaml
+
+import datastore_upgrade
+
+LATEST_VERSION = 'Latest-version'
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../AppDB'))
+from dbconstants import *
```
cdonati: This import isn't used.

bootstrap.sh
```diff
@@ -246,15 +252,29 @@ if [ -d appscale/.appscale/certs ]; then
...
                 echo "It is a force upgrade, so a git stash will be performed."
                 (cd appscale; git stash save -u)
                 (cd appscale-tools; git stash save -u)
         fi
 
         if [ -n "${GIT_TAG}" ]; then
+                set +e
                 (cd appscale; git checkout "$GIT_TAG")
+                if [ $? -gt 0 ]; then
+                    echo "Please stash your local unsaved changes."
```
menivaitsi: Change these three prints to:

echo "Please stash your local unsaved changes and checkout the version of AppScale you are currently using to fix this error."
echo "e.g.: `git stash save -u; git checkout <AppScale_version>`"


The last sentence is too specific for the bootstrap.

bootstrap.sh
```diff
@@ -246,15 +252,29 @@ if [ -d appscale/.appscale/certs ]; then
...
+                    echo "Checkout the tag of the AppScale repository you are currently using".
+                    echo "Re-run the upgrade command from the tools."
+                    exit 1
+                fi
                 (cd appscale-tools; git checkout "$GIT_TAG")
+                if [ $? -gt 0 ]; then
+                    echo "Please stash your local unsaved changes."
```
menivaitsi: Same here.

bootstrap.sh
```diff
@@ -252,46 +246,25 @@ if [ -d appscale/.appscale/certs ]; then
...
-
         if [ -n "${GIT_TAG}" ]; then
                 set +e
                 (cd appscale; git checkout "$GIT_TAG")
                 if [ $? -gt 0 ]; then
-                    echo "Please stash your local unsaved changes."
-                    echo "Checkout the tag of the AppScale repository you are currently using".
-                    echo "Re-run the upgrade command from the tools."
+                    echo "Please stash your local unsaved changes and checkout the version of AppScale you are currently using to fix this error."
+                    echo "e.g.: git stash; git checkout <AppScale-version>
```
menivaitsi: You're missing the closing quote. I forgot to put it in my comment most likely.

bootstrap.sh
```diff
@@ -252,46 +246,25 @@ if [ -d appscale/.appscale/certs ]; then
...
                     exit 1
                 fi
                 (cd appscale-tools; git checkout "$GIT_TAG")
                 if [ $? -gt 0 ]; then
-                    echo "Please stash your local unsaved changes."
-                    echo "Checkout the tag of the AppScale repository you are currently using".
-                    echo "Re-run the upgrade command from the tools."
+                    echo "Please stash your local unsaved changes and checkout the version of AppScale you are currently using to fix this error."
+                    echo "e.g.: git stash; git checkout <AppScale-version>
```
menivaitsi: Same here.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1583/

PR #2042, opened 2016-05-25
Remove references to keystore for Java AppServer

cdonati:
The keystore was previously used to establish encrypted connections
with the datastore.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1499/

PR #2043, opened 2016-05-26
Fetch Java SDK from S3 package mirror

cdonati:
This will speed up rebuilds and make it easier to fetch packages if the zip archive is unavailable in certain regions.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1512/

PR #2050, opened 2016-06-09
Build the Java SDK in a subshell

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1567/

debian/appscale_install_functions.sh
```diff
@@ -11,12 +11,12 @@ if [ -z "$APPSCALE_HOME_RUNTIME" ]; then
     export APPSCALE_HOME_RUNTIME=/opt/appscale
 fi
 
-if [ -z "$APPSCALE_PACKAGE_MIRROR" ]; then
+if [ -z "${APPSCALE_PACKAGE_MIRROR-}" ]; then
```
menivaitsi: What does the '-' do here?

cdonati: It's a parameter expansion which specifies that the empty string should be used as the default value. It's needed because of the `set -u` that I added to `appscale_install.sh`.

menivaitsi: 👍

PR #2052, opened 2016-06-10
Fix settings generation for AppDashboard in AppController.

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/1570/


AppController/lib/app_dashboard.rb
```diff
@@ -48,6 +47,17 @@ module AppDashboard
   #   true if the AppDashboard was started successfully, and false otherwise.
   def self.start(login_ip, uaserver_ip, public_ip, private_ip,
       persistent_storage, secret)
+
+    # Pass the secret key and our public IP address (needed to connect to the
+    # AppController) to the app.
+    Djinn.log_run("cp #{APPSCALE_HOME}/AppDashboard/lib/secret_key.py /tmp/secret_key.py")
+    Djinn.log_run("cp #{APPSCALE_HOME}/AppDashboard/lib/local_host.py /tmp/local_host.py")
+    Djinn.log_run("cp #{APPSCALE_HOME}/AppDashboard/lib/uaserver_host.py /tmp/uaserver_host.py")
```
cdonati: Could this be replaced with native Ruby `File` operations? (read the contents here and write later)?

menivaitsi: The value of these constants is the same in all three templates.
We can even just overwrite them with a constant after we're done tarring up the app dir.

Is that ok?

cdonati: Yeah, that works for me.

obino: The code was using app_location that is defined just below the 3 lines:
if you move them after the definition of app_location, there is no
hardcoded path for those templates.

menivaitsi: Actually we modified that variable a while ago and that's why the dashboard broke.
Please refer to https://github.com/AppScale/appscale/commit/c4170768ff1b84744edfa909210a20d15437a064#diff-3728529fdd148dba907df39fdadd6521L53

PR #2069, opened 2016-07-07
Allow retries while connecting to Cassandra

cdonati:
This should reduce some of the excessive controller logging and fix an issue with the upgrade script attempting to run before Cassandra is ready.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1640/

PR #2071, opened 2016-07-08
Upgrade script changes to sleep until ZooKeeper and enough Cassandra nodes are up

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1650/

### DIFF
--- a/lib/templates/monit_template_for_upgrade.conf
+++ b/lib/templates/monit_template_for_upgrade.conf
@@ -2,4 +2,4 @@ check process {watch}-{port} matching "{match}"
   group {watch}
   start program = "/bin/bash -c '{env} {start} >>/var/log/appscale/{watch}-{port}.log'"
   stop program = "{stop}"
-  
+
--- a/lib/templates/monit_template_for_upgrade.conf
+++ b/lib/templates/monit_template_for_upgrade.conf
@@ -2,5 +2,4 @@ check process {watch}-{port} matching "{match}"
   group {watch}
   start program = "/bin/bash -c '{env} {start} >>/var/log/appscale/{watch}-{port}.log'"
   stop program = "{stop}"
-  if totalmem > {memory} MB for 10 cycles then restart
   
--- a/scripts/datastore_upgrade.py
+++ b/scripts/datastore_upgrade.py
@@ -426,13 +426,6 @@ def run_datastore_upgrade(zk_ips, db_ips, master_ip, status_dict, keyname):
   start_cassandra(status_dict, db_ips, master_ip, keyname)
   start_zookeeper(status_dict, zk_ips, master_ip, keyname)
 
-  # Sleep time for Cassandra and ZooKeeper to be started.
-  while subprocess.call([cassandra_interface.NODE_TOOL, 'status']) !=0:
-    time.sleep(10)
-
-  while subprocess.call(['/usr/share/zookeeper/bin/zkServer.sh', 'status']) !=0:
-    time.sleep(10)
-
   if not all_services_started(status_dict):
     stop_cassandra(db_ips, master_ip, status_dict, keyname)
     stop_zookeeper(zk_ips, master_ip, status_dict, keyname)
--- a/scripts/datastore_upgrade.py
+++ b/scripts/datastore_upgrade.py
@@ -435,7 +435,7 @@ def ensure_cassandra_nodes_match_replication(keyname):
     logging.info("{0} nodes are up. {1} are needed.".format(nodes_ready, replication))
     if nodes_ready >= int(replication):
       break
-    time.sleep(5)
+    time.sleep(1)
 
 def run_datastore_upgrade(zk_ips, db_ips, master_ip, status_dict, keyname):
   """ Runs the data upgrade process of fetching, validating and updating data
--- a/scripts/datastore_upgrade.py
+++ b/scripts/datastore_upgrade.py
@@ -5,12 +5,14 @@ import os
 import subprocess
 import sys
 import time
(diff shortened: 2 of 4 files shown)


scripts/datastore_upgrade.py
```diff
@@ -409,6 +411,32 @@ def all_services_started(status_dict):
       return False
   return True
 
+def ensure_cassandra_nodes_match_replication(keyname):
```
cdonati: Add docstring

scripts/datastore_upgrade.py
```diff
@@ -409,6 +411,32 @@ def all_services_started(status_dict):
...
+  # Get the database master IP.
+  with open(MASTERS_FILE_LOC) as masters_file:
```
cdonati: You can use `master_ip` var from the parent function. I'm going to open a PR that ensures that `master_ip` is the private IP for the db master since it's needed for cassandra.yaml.

scripts/datastore_upgrade.py
```diff
@@ -409,6 +411,32 @@ def all_services_started(status_dict):
...
+    master_ip = masters_file.read().strip()
+
+  command = cassandra_interface.NODE_TOOL + " " + 'status'
+  key_file = '{}/{}.key'.format(utils.KEY_DIRECTORY, keyname)
+  ssh_cmd = ['ssh', '-i', key_file, master_ip, command]
+  cmd_output = subprocess.Popen(ssh_cmd, stdout=subprocess.PIPE)
```
cdonati: Just for future reference, you can use the `subprocess.check_output` convenience function when you need the output.

scripts/monit_start_service.py
```diff
@@ -26,19 +28,31 @@
 # The default port to connect to Cassandra.
 CASSANDRA_PORT = 9999
 
+# The service name for monit to start Cassandra.
+CASSANDRA_SERVICE_NAME = "cassandra"
```
menivaitsi: Are those constants also defined somewhere else? Can you use the existing ones if any?

scripts/datastore_upgrade.py
```diff
@@ -409,6 +411,32 @@ def all_services_started(status_dict):
...
+
+  nodes_ready = 0
+  # Get the replication factor from the database_info.yaml file.
+  with open(DB_INFO_LOC) as db_info_file:
+    db_info = yaml.load(db_info_file)
+  replication = db_info[':replication']
+
+  while True:
```
menivaitsi: Is there a function for this in cassandra_prime?

scripts/datastore_upgrade.py
menivaitsi: There's already a method for this in appscale/lib/appscale_info.py I believe.

lib/templates/monit_template_for_upgrade.conf
```diff
@@ -2,5 +2,4 @@ check process {watch}-{port} matching "{match}"
   group {watch}
   start program = "/bin/bash -c '{env} {start} >>/var/log/appscale/{watch}-{port}.log'"
   stop program = "{stop}"
-  if totalmem > {memory} MB for 10 cycles then restart
```
menivaitsi: Add newline at end of file.

scripts/monit_start_service.py
```diff
@@ -50,9 +64,19 @@ def start_service(service_name):
     logging.error("Monit was unable to start " + service_name)
     return 1
   else:
+    sleep_until_service_is_up(service_name)
     logging.info("Successfully started " + service_name)
     return 0
 
+def sleep_until_service_is_up(service_name):
```
menivaitsi: Can you pass the subprocess command and args as params to this function?
That way the body of this function could be simplified and look something like:

```
while subprocess.call([command, cmd_args]) != 0:
  time.sleep(10)
```

menivaitsi: Add a docstring here.

scripts/monit_start_service.py
```diff
@@ -26,19 +28,31 @@
 # The default port to connect to Cassandra.
 CASSANDRA_PORT = 9999
 
+# The service name for monit to start Cassandra.
+CASSANDRA_SERVICE_NAME = "cassandra"
+
+# The service name for monit to start Zookeeper.
+ZK_SERVICE_NAME = "zookeeper"
+
 def start_service(service_name):
```
menivaitsi: Add missing docstring.

scripts/datastore_upgrade.py
menivaitsi: Add docstring.

scripts/monit_start_service.py
tmarballi: I could do that but the code where the method is called is common for Zk and Cassandra. So it would need additional ifs there anyway to pass the respective command. I could do either way!

scripts/datastore_upgrade.py
tmarballi: Yes, Thanks there was an existing method in appscale_info for both getting the master ip here and replication factor below!

scripts/monit_start_service.py
```diff
@@ -50,9 +64,18 @@ def start_service(service_name):
...
     return 1
   else:
+    sleep_until_service_is_up(service_command)
     logging.info("Successfully started " + service_name)
     return 0
 
+def sleep_until_service_is_up(service_command):
+  """ Sleeps until the service is actually up after Monit has started it.
+    Args:
+      service_name: The name of the service to start.
```
menivaitsi: Please update the Args section.

PR #2083, opened 2016-07-22
Deleting a push task after completion/expiration/permanent failure.

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/1711/


AppTaskQueue/templates/task.py
```diff
@@ -55,6 +55,7 @@ def get_wait_time(retries, args):
       logger.error("Task %s with id %s has expired with expiration date %s" % \
                    (args['task_name'], QUEUE_NAME.request.id, args['expires']))
       celery.control.revoke(QUEUE_NAME.request.id)
```
cdonati: Is the task update needed right before deleting the task?

menivaitsi: Nop, thanks!

PR #2084, opened 2016-07-22
Check if state is for this deployment.

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1713/

AppController/djinn.rb
```diff
@@ -2962,6 +2960,11 @@ def restore_appcontroller_state()
     Djinn.log_info("Reload State : #{json_state}")
 
     @@secret = json_state['@@secret']
+    # Is the state we are reading current?
+    if @@secret != HelperFunctions.get_secret()
```
menivaitsi: Won't this condition be satisfied if I restart my AppScale deployment?

obino: It should be, since the secret is passed around by the headnode. If
this is part of the old deployment, then it doesn't matter, since the
state should still be valid. This test should fail only if the state
belongs to a different deployment: the node should then wait for the
headnode.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1756/

AppController/djinn.rb
```diff
@@ -1885,6 +1885,7 @@ def job_start(secret)
     write_zookeeper_locations()
     pick_zookeeper(@zookeeper_data)
     write_our_node_info()
+    backup_appcontroller_state() if my_node.is_shadow?
```
cdonati: What happens if ZooKeeper is not up yet at this point?

obino:
This won't cover the cases I am thinking about (volumes belonging to the wrong deployment) most of the time, so closing it.

PR #2088, opened 2016-07-23
Wait until a quorum can be achieved for every token

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1718/

AppController/djinn.rb
```diff
@@ -3609,12 +3609,14 @@ def start_api_services()
       threads << Thread.new {
         Djinn.log_info("Starting database services.")
         clear_datastore = @options['clear_datastore'].downcase == "true"
-        replication = Integer(@options['replication'])
+        db_nodes = @nodes.count{|node| node.is_db_master? or node.is_db_slave?}
+        needed_nodes = needed_for_quorum(db_nodes,
```
menivaitsi: Would this be a problem in migration cases where the replication factor increases or decreases?

cdonati: @menivaitsi We don't have an automated way of changing the replication factor yet. We assume that the replication in the configuration matches the keyspace.

Though the tools check to make sure that the replication factor doesn't exceed the total number of nodes, it's a good idea to add a check here as well. Otherwise, we could end up in a situation where it's waiting for less than 1 node to be ready.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1724/

AppController/djinn.rb
menivaitsi: @cdonati That makes sense. Thanks!

menivaitsi:
👍

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1729/

PR #2091, opened 2016-07-25
Use older version of httparty

cdonati:
Some Ruby code depends on httparty 0.13.7's behavior of allowing additional types to be passed to JSON.dump. This is intended as a quick fix until all of the JSON code can be inspected and fixed.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1727/

PR #2092, opened 2016-07-26
Bumping AppScale version to 3.0.0

menivaitsi:


RELEASE
```diff
@@ -5,6 +5,40 @@
...
+- Log monit service errors
+- Log the upgrade progress
+- Remove app archive upon appscale remove/undeploy
+- Removed unused code in AppTaskQueue
+- Delete push tasks after completion/expiration
+- Use fixed version of wstools
+- Fixed bug with updating cron upon app redeploy
+- Added logging of datastore results in debug mode
+- Removed confusing error about non-existing dir during bootstrap
+- Use fixed version of httparty
```
cdonati: I don't think we need entries for things that don't change anything.

cdonati:
In the future, I think we should try to order things roughly by impact rather than date committed.

PR #2094, opened 2016-07-28
Set num_tokens to 1

cdonati:
This is for compatibility with existing deployments. Cassandra's
behavior changed in 2.1.15. Firstly, it actually uses this value
if initial_token is set to an empty string. Secondly, "Specifying
initial_token will override this setting on the node's initial
start, on subsequent starts, this setting will apply even if
initial token is set."

### DIFF
--- a/scripts/datastore_upgrade.py
+++ b/scripts/datastore_upgrade.py
@@ -27,6 +27,7 @@ from datastore_server import ID_KEY_LENGTH
 from dbconstants import APP_ENTITY_SCHEMA
 from dbconstants import APP_ENTITY_TABLE
 from zkappscale import zktransaction as zk
+from zkappscale.zktransaction import ZK_SERVER_CMD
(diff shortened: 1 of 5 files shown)


scripts/datastore_upgrade.py
```diff
@@ -107,7 +113,13 @@ def start_cassandra(db_ips, db_master, keyname):
       logging.exception(message)
       raise dbconstants.AppScaleDBError(message)
 
-    logging.info("Successfully started Cassandra.")
+  logging.info('Waiting for Cassandra to be ready')
+  status_cmd = '{} status'.format(cassandra_interface.NODE_TOOL)
+  while (utils.ssh(db_master, keyname, status_cmd,
```
menivaitsi: Is vertical alignment slightly off here?

cdonati: Yeah, I fixed it in https://github.com/AppScale/appscale/pull/2096

PR #2098, opened 2016-07-29
Don't define initial token

cdonati:
Previous versions of Cassandra would change num_tokens
on restarts, so setting num_tokens to 256 without defining an
initial token is the most compatible configurations for existing
deployments.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1750/

PR #2107, opened 2016-08-09
Use unless instead of if !

obino:
Follow ruby practice.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -2727,7 +2727,7 @@ class Djinn
   def update_firewall()
     all_ips = []
     @nodes.each { |node|
-      all_ips << node.private_ip unless all_ips.include? node.private_ip
+      (all_ips << node.private_ip) unless all_ips.include? node.private_ip
     }
     all_ips << "\n"
     HelperFunctions.write_file("#{APPSCALE_CONFIG_DIR}/all_ips", all_ips.join("\n"))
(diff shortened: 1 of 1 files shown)


AppController/djinn.rb
```diff
@@ -5144,11 +5122,7 @@ def initialize_scaling_info_for_app(app_name, force=false)
     @current_req_rate[app_name] = 0
     @total_req_rate[app_name] = 0
     @last_sampling_time[app_name] = Time.now.to_i
-
-    if !@last_decision.has_key?(app_name)
-      @last_decision[app_name] = 0
-    end
-
+    @last_decision[app_name] = 0 unless @last_decision.has_key?(app_name)
```
cdonati: I know that this works, but the way it reads is ambiguous. I could see it as either
`(@last_decision[app_name] = 0) unless @last_decision.has_key?(app_name)`
or
``@last_decision[app_name] = (0 unless @last_decision.has_key?(app_name))`

AppController/djinn.rb
```diff
@@ -2745,9 +2727,7 @@ def write_database_info()
   def update_firewall()
     all_ips = []
     @nodes.each { |node|
-      if !all_ips.include? node.private_ip
-        all_ips << node.private_ip
-      end
+      all_ips << node.private_ip unless all_ips.include? node.private_ip
```
cdonati: I think this is harder to parse as a one-liner due to the ambiguity:
`(all_ips << node.private_ip) unless all_ips.include? node.private_ip` vs `all_ips << (node.private_ip unless all_ips.include? node.private_ip)`.

PR #2111, opened 2016-08-11
Add additional index to filter on tag existence

cdonati:
Since Cassandra can only do `=` queries on secondary indices, we need to specify the existence of a tag explicitly for `groupByTag=true,tag=None` queries.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1795/

PR #2114, opened 2016-08-11
Remove routing if no appserver is running

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1810/

cdonati:
:+1:

AppController/djinn.rb
```diff
@@ -983,13 +983,15 @@ def status(secret)
         next if app_name == "none"
         stats_str << "    Information for application: #{app_name}\n"
         stats_str << "        Language            : "
-        unless @app_info_map[app_name]['language'].nil?
-          stats_str << "#{@app_info_map[app_name]['language']}\n"
-        else
+        if @app_info_map[app_name]['language'].nil?
```
menivaitsi: Is this the best check to ensure app isn't running? Can we instead use just @app_info_map[app_name]?

menivaitsi: I understand that if the language is not set, we report stats unknown. Correct?

obino: This is just reporting, and here we just report the language of the application. This is set way before the appserver is running, so technically we can see a 'pending' appserver (that is the app is not running), but we can still have the language showing properly. The case (rare) is that this call is done after the update call has been done, so the app_info_map for the app is created, but the language has not been set yet.

menivaitsi: @obino Thanks for clarifying.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1820/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1827/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1834/

PR #2121, opened 2016-08-17
Don't start the dashbaord in a thread

obino:
This seems to create a race condition when restarting (ie when the state
already has the dashborard's appservers), where the tarball of the app
appears to be corrupted.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1833/

PR #2128, opened 2016-08-19
Check the return value from done_uploading

obino:
done_uploading adds the node to the list of 'app-hosters' in zookeeper.
This is used to find which nodes has a recent copy of the application when
it's time to refresh the local copy. It could happen that zookeeper is not
available at that time, so we need to retry few times and print an error
message if something goes wrong.

Reference:
https://ocd.appscale.com:8080/job/Blobstore/1149/artifact/appscale-logs/192.168.103.178/appscale/controller-17443.log

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -5438,8 +5438,16 @@ HOSTS
       @app_info_map[app]['language'] = "python27"
     else
       # Make sure we advertize that this node is hosting the app code.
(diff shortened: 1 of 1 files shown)


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1856/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1857/

PR #2139, opened 2016-08-29
Remove SOAPpy library from codebase

cdonati:
It's already installed via pip, so there's no need for it.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1911/

PR #2140, opened 2016-08-29
Add Azure agent to Infrastructure Manager

tmarballi:


InfrastructureManager/agents/factory.py
```diff
@@ -1,7 +1,9 @@
+from agents.azure_agent import AzureAgent
```
cdonati: Is this the same code as the tools have? Could you just do `from appscale.tools.agents.azure_agent import AzureAgent`, or is there a difference between the two agents?

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,559 @@
...
+    PARAM_TENANT_ID,
+    PARAM_ZONE
+  )
+
+  # The admin username needed to create an Azure VM instance.
+  ADMIN_USERNAME = 'azureuser'
+
+  # The file path for the authorized keys on the head node
+  # for an Azure VM.
+  AUTHORIZED_KEYS_FILE = "/home/{}/.ssh/authorized_keys"
```
obino: /home is not a good choce to hardcode since the homes directories could be mounted somewhere else. We could use ~ instaed.

InfrastructureManager/agents/factory.py
tmarballi: No sadly the agent code on the AppScale side is different from the Tools. Eventually I will have to look at what we need to just do the import but all the agents are used a little differently on the AppScale side.

InfrastructureManager/agents/azure_agent.py
tmarballi: I had tried giving it the '~' before and the sshconfig keypath where this file is used, needs for it to be /home/{ADMIN-USERNAME} and if I give it anything else, it comes back as invalid.

cdonati: You can use `os.path.expanduser`.

obino: hmm ... ~azureuser or ~root should work, but you need to make sure it's passed like ~ to the remote shell (ie not interpreted from the local shell)

AppController/djinn.rb
```diff
@@ -3963,6 +3964,20 @@ def copy_encryption_keys(dest_node)
     HelperFunctions.scp_file(gce_oauth, gce_oauth, ip, ssh_key)
   end
 
+  def enable_root_login(ip, ssh_key, user_name)
```
menivaitsi: Please add docstring to this method.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+  PARAM_CREDENTIALS = 'credentials'
+  PARAM_EXISTING_RG = 'does_exist'
+  PARAM_GROUP = 'group'
+  PARAM_INSTANCE_IDS = 'instance_ids'
+  PARAM_KEYNAME = 'keyname'
+  PARAM_IMAGE_ID = 'image_id'
+  PARAM_REGION = 'region'
+  PARAM_RESOURCE_GROUP = 'resource_group'
+  PARAM_STORAGE_ACCOUNT = 'storage_account'
+  PARAM_SUBCR_ID = 'subscription_id'
```
menivaitsi: I'd change that to PARAM_SUBSCRIPTION_ID to match the rest of them.
If that makes other usages too long, change to PARAM_SUB**S**CR_ID.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+                             parameters, vm_network_name):
+    """ Creates an Azure virtual machine using the network interface created.
+    Args:
+      credentials: A ServicePrincipalCredentials instance, that can be used to
+        access or create any resources.
+      network_client: A NetworkManagementClient instance.
+      network_id: The network id of the network interface created.
+      parameters: A dict, containing all the parameters necessary to
+        authenticate this user with Azure.
+      vm_network_name: The name of the Virtual machine to use.
```
menivaitsi: Virtual -> virtual

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+        format(storage_account, vm_network_name))
+
+    image_hd = VirtualHardDisk(uri=parameters[self.PARAM_IMAGE_ID])
+    os_type = OperatingSystemTypes.linux
+    os_disk = OSDisk(os_type=os_type, caching=CachingTypes.read_write,
+                     create_option=DiskCreateOptionTypes.from_image,
+                     name=vm_network_name, vhd=virtual_hd, image=image_hd)
+
+    compute_client.virtual_machines.create_or_update(
+      resource_group, vm_network_name, VirtualMachine(location=zone,
```
menivaitsi: Either move the location arg to the next line, or align vertically here.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+      if public_ip_address.ip_address:
+        utils.log('Azure VM is available at {}'.
+                  format(public_ip_address.ip_address))
+        break
+      utils.log("Waiting {} second(s) for IP address to be available".
+                format(self.SLEEP_TIME))
+      time.sleep(self.SLEEP_TIME)
+
+  def associate_static_ip(self, instance_id, static_ip):
+    """Associates the given static IP address with the given instance ID.
```
menivaitsi: Add space after opening """

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+
+  def sleep_until_delete_operation_done(self, result, resource_name):
+    """ Sleeps until the delete operation for the resource is completed
+    successfully.
+    Args:
+      result: An instance, of the AzureOperationPoller to poll for the status
+        of the operation being performed.
+      resource_name: The name of the resource being deleted.
+    """
+    while not result.done():
```
menivaitsi: Don't we need a timeout here in case something hangs?

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+    Args:
+      parameters: A dict that includes keys indicating the remote state
+        that should be deleted.
+    """
+    subscription_id = str(parameters[self.PARAM_SUBCR_ID])
+    resource_group = parameters[self.PARAM_RESOURCE_GROUP]
+    credentials = self.open_connection(parameters)
+    network_client = NetworkManagementClient(credentials, subscription_id)
+
+    utils.log("Deleting the Virtual Network, Public IP Address "
```
menivaitsi: Do this operations need to be sequential? Can we put them in separate threads?

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+      parameters: A dict, containing all the parameters necessary to authenticate
+        this user with Azure. We assume that the user has already authorized this
+        account by creating a Service Principal with the appropriate (Contributor)
+        role.
+    Returns:
+      A ServicePrincipalCredentials instance, that can be used to access or
+        create any resources.
+    """
+    app_id = parameters[self.PARAM_APP_ID]
+    app_secret_key = parameters[self.PARAM_APP_SECRET] + "="
```
menivaitsi: Why do we append '=' here?

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+    return credentials
+
+  def create_network_interface(self, network_client, interface_name, ip_name,
+                               subnet, parameters):
+    """ Creates the Public IP Address resource and uses that to create the
+    Network Interface.
+    Args:
+      network_client: A NetworkManagementClient instance.
+      interface_name: The name to use for the Network Interface.
+      ip_name: The name to use for the Public IP Address.
```
menivaitsi: Is this arg the name of the VM or the actual IP? Please make the var name and description less ambiguous.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+      location=region, public_ip_allocation_method=IPAllocationMethod.dynamic,
+      idle_timeout_in_minutes=4)
+    result = network_client.public_ip_addresses.create_or_update(
+      group_name, ip_name, ip_address)
+    self.sleep_until_update_operation_done(result, ip_name)
+    public_ip_address = network_client.public_ip_addresses.get(group_name, ip_name)
+
+    utils.log("Creating/Updating the Network Interface '{}'".format(interface_name))
+    network_interface_ip_conf = NetworkInterfaceIPConfiguration(
+      name='default', private_ip_allocation_method=IPAllocationMethod.dynamic,
```
menivaitsi: Could someone choose a different name from their account and maybe use that?
What if an existing default network had conflicting network rules with what AppScale needs to function?

menivaitsi: I'd change that to a variable and we can add an AppScalefile setting to configure this at a later point in time.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+      subnet=subnet, public_ip_address=PublicIPAddress(id=(public_ip_address.id)))
+
+    result = network_client.network_interfaces.create_or_update(group_name,
```
menivaitsi: Move group_name to the next line.

InfrastructureManager/agents/azure_agent.py
```diff
@@ -0,0 +1,454 @@
...
+
+  def sleep_until_update_operation_done(self, result, resource_name):
+    """ Sleeps until the create/update operation for the resource is completed
+      successfully.
+    Args:
+      result: An instance, of the AzureOperationPoller to poll for the status
+        of the operation being performed.
+      resource_name: The name of the resource being updated.
+    """
+    while not result.done():
```
menivaitsi: I'd argue there's a similar need for timeout here.

InfrastructureManager/agents/azure_agent.py
tmarballi: Yes, the operations need to be sequential because Network Interface uses the Public IP Address which uses the Virtual Network. So they have to be deleted in that order. But they are pretty quick operations unlike the create/delete VM so I believe it should be fine.

InfrastructureManager/agents/azure_agent.py
tmarballi: You are right, It needs a max timeout value. Will add that!

InfrastructureManager/agents/azure_agent.py
tmarballi: It actually is the name for Public IP Address resource which needs to be created to provide to the Network Interface. The actual IP is the id on that resource and can be retrieved using this name.

InfrastructureManager/agents/azure_agent.py
tmarballi: That makes sense! I will change it to use the same name as the network interface.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1958/

PR #2144, opened 2016-08-30
Return long queue form in leased tasks

menivaitsi:


menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/1932/

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -661,7 +666,7 @@ def _lease_task(self, index, new_eta):
     parameters = {'app': self.app, 'queue': self.name, 'id': task_id}
     result = self.db_access.session.execute(select_task, parameters)[0]
     task_info = {
-      'queueName': self.name,
+      'queueName': LONG_QUEUE_FORM.format(app=self.app, queue=self.name),
```
cdonati: I think the substitution should happen inside `rest_api.py` since this class is shared between the rest and protobuf servers.

menivaitsi: @cdonati It doesn't look like I have the appID available in that context. Any suggestions?

cdonati: `project` would be the relevant variable there.

menivaitsi: The REST API handler returns a JSON safe version of a task.
In the task context we don't always have the project available. 

I was hoping to avoid adding two new args in https://github.com/AppScale/appscale/blob/master/AppTaskQueue/appscale/taskqueue/task.py#L148 to call it with long_queue_form mode on and to give it an app/project ID.
I guess that's the only way though.

cdonati: I don't think you need new arguments to json_safe_dict. Yuu can just filter it afterwards.

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/1935/

cdonati:
At the end of the post method on the REST API handler, you can do `[task.update({'queueName': LONG_FORM_QUEUE.format(...)} for task in tasks]` to replace the task_list being serialized.

menivaitsi:
@cdonati Good point. I need a break :p I'll push a clean branch.

PR #2159, opened 2016-09-07
Fix dashboard app upload functionality

cdonati:


AppController/djinn.rb
```diff
@@ -111,6 +111,10 @@
 APPSCALE_APP_LOGROTATE = 'appscale-app-logrotate.conf'
 
 
+# The location of the appscale-upload-app script from appscale-tools.
+UPLOAD_APP_SCRIPT = '/usr/local/bin/appscale-upload-app'
```
menivaitsi: Can that path be dynamically discovered? What would be the case in other distros?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1989/

AppController/djinn.rb
cdonati: Unfortunately, /usr/local/bin is not on the path when djinnServer is started from the init script. It's certainly possible for that script to be installed elsewhere, but I'm not sure the best way to handle it. We could alter the path in the init script.

AppController/lib/app_controller_client.rb
```diff
@@ -71,7 +71,7 @@ def initialize(ip, secret)
     @conn.add_method("set_apps_to_restart", "apps_to_restart", "secret")
     @conn.add_method("status", "secret")
     @conn.add_method("get_stats", "secret")
-    @conn.add_method("upload_app", "app", "secret")
+    @conn.add_method("upload_app", "archived_file", "file_suffix", "email", "secret")
```
menivaitsi: Is this method only called for dashboard uploads?

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -168,15 +168,17 @@ def __init__(self):
     }
 
 
-  def get_appcontroller_client(self):
+  def get_appcontroller_client(self, server_ip=MY_PUBLIC_IP):
     """ Retrieves our saved AppController connection, creating a new one if none
     currently exist.
 
+    Args:
+      server_ip: An IP address specifying which machine to make AC calls to.
```
menivaitsi: AC -> AppController

Helps with debugging and searches.

AppController/lib/app_controller_client.rb
cdonati: Yes. The tools scp the file directly and then call `done_uploading`.

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -347,27 +350,32 @@ def upload_app(self, filename, upload_file):
...
-        if status == AppUploadStatuses.COMPLETE:
-          os.remove('{}.{}'.format(tgz_file.name, file_suffix))
-          return 'Application uploaded successfully. Please wait for the '\
-            'application to start running.'
-      os.remove('{}.{}'.format(tgz_file.name, file_suffix))
-      raise AppHelperException('Saw status {} when trying to upload app.'
-        .format(status))
+
+      # The local controller needs to SCP the tempfile to the shadow node.
+      acc = self.get_appcontroller_client(server_ip='127.0.0.1')
```
obino: This is ruby, remove server_ip=

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -347,27 +350,32 @@ def upload_app(self, filename, upload_file):
...
+
+      # The sandboxed version of tempfile does not support specifying a suffix.
+      with tempfile.NamedTemporaryFile(delete=False) as tgz_file:
```
obino: Same here, remove delete=

AppController/djinn.rb
```diff
@@ -1068,28 +1075,33 @@ def upload_app(archived_file, file_suffix, email, secret)
...
+        new_location = [archived_file, file_suffix].join('.')
+        Djinn.log_debug("Copying #{archived_file} to #{new_location}")
+        FileUtils.copy(archived_file, new_location)
+        archived_file = new_location
       end
 
       Djinn.log_debug("Uploading file at location #{archived_file}")
       keyname = @options['keyname']
-      command = "appscale-upload-app --file '#{archived_file}' " +
+      command = "#{UPLOAD_APP_SCRIPT} --file '#{archived_file}' " +
```
obino: The old command was without a full path: would that work? Don't we set the /usr/local/bin path when installing the tools?

AppDashboard/lib/app_dashboard_helper.py
cdonati: I'm pretty sure it's not Ruby. :)

AppController/djinn.rb
cdonati: It's been broken since we stopped modifying the path after installing the tools. It's available on the path (from a shell) automatically after installing with `python setup.py install`, but it's not available from the init script that starts the controller.

PR #2161, opened 2016-09-08
Don't use celery 4 yet

cdonati:
At least some code will need to be rewritten to be compatible with the new version of celery.


cdonati:
Precise: https://ocd.appscale.com:8080/job/Daily%20Build/2018/
Trusty: https://ocd.appscale.com:8080/job/Daily%20Build/2019/

PR #2165, opened 2016-09-10
Fix keyname on restore

obino:
If AC restarts, we need to make sure we use the keyname stored in zookeeper.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2010/

PR #2168, opened 2016-09-12
Don't fail when max_concurrent_requests is defined

cdonati:
Ignore the value for now. We can make use of it later when setting up the routing. This change was introduced in the 1.8.7 SDK.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2020/

PR #2171, opened 2016-09-13
Fix datastore viewer script 

tmarballi:


scripts/enable-datastore-viewer.sh
```diff
@@ -5,9 +5,10 @@
 # Enable the datastore viewer and reload nginx.
 
 ALLOW=""
-APPS_ID=""
+APP_ID=""
 IP="all"
 VIEWER_PORT="8099"
```
menivaitsi: AppScale opens up ports 8080-8099 for apps (https://github.com/AppScale/appscale/blob/master/firewall.conf#L25). 
That means that potentially all these ports can be occupied by running apps.

That block would have to be expanded for datastore viewer use.
Either with a static range 8100-8119 or dynamically after a datastore viewer is enabled.
In either case, the firewall.conf should be modified to allow this.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,32 +54,32 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
-                for x in $APPS_ID ; do
-                        if [ "$x" = "$app_id" ]; then
-                                found="yes"
-                                break
-                        fi
-                done
-                if [ "$found" = "no" ]; then
+APPENGINE_IP=""
+for x in $(cat /etc/appscale/all_ips); do
+        OUTPUT=$(ssh $x -i /etc/appscale/keys/cloud1/*.key 'ps ax | grep appserver | grep -Ev '(grep|appscaledashboard)' | grep -- "--admin_port" | sed "s;.*--admin_port \([0-9]*\).*/var/apps/\(.*\)/app .*;\1 \2;g" | sort -ru' | grep "$APP_ID")
```
menivaitsi: Please replace the `grep -Ev '(grep|appscaledashboard)'` part with the `grep "$APP_ID"` at the end.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,32 +54,32 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
+                        port=$i
+                        APPENGINE_IP=$x
+                        break
                 fi
-        fi
+        done
+done
 
-	let $((VIEWER_PORT += 1))
+let $((VIEWER_PORT += 1))
```
menivaitsi: This is not necessary if you start the VIEWER_PORT from 8100 at https://github.com/AppScale/appscale/pull/2171/files#diff-d438d935c16df0ab964b80d79ce47507L10

scripts/enable-datastore-viewer.sh
```diff
@@ -53,32 +54,32 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
 
-	# Do not apply if it's already there.
-	if grep "datastore_viewer" /etc/nginx/sites-enabled/* > /dev/null ; then
-		echo "Datastore viewer already enabled for $app_id."
-		continue
-	fi
+while [[ $(lsof -i :$VIEWER_PORT) ]]; do
+        let $((VIEWER_PORT += 1))
```
menivaitsi: Indentation is off here. Please make it uniform across the script.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,32 +54,32 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
+done
+
+while [[ $(lsof -i :$LOCAL_PORT) ]]; do
+        let $((LOCAL_PORT += 1))
```
menivaitsi: Same here.

scripts/enable-datastore-viewer.sh
```diff
@@ -96,12 +97,15 @@ server {
...
+		echo "$pippo" >> /etc/nginx/sites-enabled/${APP_ID}_datastore_viewer.conf
+		echo "Datastore viewer enabled for $APP_ID at http://$(cat /etc/appscale/my_public_ip):$VIEWER_PORT. Allowed IP: $IP."
 		service nginx reload
-	else
-		echo "Cannot find configuration for ${app_id}."
-	fi
-done
+		echo "Note: For a multi node deployment, you will need to forward the admin port from one of the AppServers on another node to the head node"
+        echo "You might need to type in the password for the app engine node."
+        echo "Run the SSH Tunnelling command: ssh -L ${LOCAL_PORT}:localhost:$port $APPENGINE_IP -N"
```
menivaitsi: It's good practice to use {} around variables when printing in Bash to avoid ambiguities when expanding the variable.

Please add for port and APPENGINE_IP here and throughout.

scripts/enable-datastore-viewer.sh
```diff
@@ -96,12 +97,15 @@ server {
     }
 }
 "
-	if [ -e /etc/nginx/sites-enabled/${app_id}.conf ]; then
-		cp /etc/nginx/sites-enabled/${app_id}.conf /tmp
-		echo "$pippo" >> /etc/nginx/sites-enabled/${app_id}.conf
-		echo "Datastore viewer enabled for $app_id at http://$(cat /etc/appscale/my_public_ip):$VIEWER_PORT. Allowed IP: $IP."
+
+if [ -e /etc/nginx/sites-enabled/${APP_ID}.conf ]; then
+		cp /etc/nginx/sites-enabled/${APP_ID}.conf /tmp
```
menivaitsi: Is the /tmp copy ever used for a rollback? If not, please remove.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,32 +54,32 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
+done
+
+while [[ $(lsof -i :$LOCAL_PORT) ]]; do
+        let $((LOCAL_PORT += 1))
+done
 
-	# Prepare the nginx config snippet.
-	pippo="
+# Prepare the nginx config snippet.
+pippo="
```
menivaitsi: I'd change this variable to a more meaningful name like: `config`

scripts/enable-datastore-viewer.sh
obino: Agreed: we don't touch the application nginx configuration, so this backup is not needed anylonger.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,34 +54,37 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
-                found="no"
-                for x in $APPS_ID ; do
-                        if [ "$x" = "$app_id" ]; then
-                                found="yes"
-                                break
-                        fi
-                done
-                if [ "$found" = "no" ]; then
+APPENGINE_IP=""
+for x in $(cat /etc/appscale/all_ips); do
```
cdonati: Give `x` a more descriptive name.

scripts/enable-datastore-viewer.sh
```diff
@@ -53,34 +54,37 @@ if [ ! -e /etc/nginx/sites-enabled ]; then
...
+        OUTPUT=$(ssh $x -i /etc/appscale/keys/cloud1/*.key 'ps ax | grep appserver | grep -E "(grep|$APP_ID)" | grep -- "--admin_port" | sed "s;.*--admin_port \([0-9]*\).*/var/apps/\(.*\)/app .*;\1 \2;g" | sort -ru')
```
cdonati: Break into multiple lines with a `\` to reduce width.

scripts/enable-datastore-viewer.sh
```diff
@@ -96,12 +100,17 @@ server {
...
-		service nginx reload
-	else
-		echo "Cannot find configuration for ${app_id}."
-	fi
-done
+
+if [ -e /etc/nginx/sites-enabled/${APP_ID}.conf ]; then
+        echo "$config" > /etc/nginx/sites-enabled/${APP_ID}_datastore_viewer.conf
+        echo "Datastore viewer enabled for ${APP_ID} at http://$(cat /etc/appscale/my_public_ip):${VIEWER_PORT}. Allowed IP: $IP."
+        service nginx reload
```
cdonati: You may want to print the success statement after the reload in case the reload fails.

scripts/enable-datastore-viewer.sh
```diff
@@ -96,12 +100,17 @@ server {
...
+        if [ "$NO_OF_IP_LINES" != "2" ]; then
+                echo "Note: For a multi node deployment, you will need to forward the admin port from one of the AppServers on another node to the head node."
+                echo "You might need to type in the password for the app engine node."
+                echo "Run the SSH Tunnelling command: ssh -L ${LOCAL_PORT}:localhost:${port} ${APPENGINE_IP} -N"
+                echo "Modify the appscale/firewall.conf to open ports for the datastore viewer."
```
menivaitsi: The order of the instructions here is a bit weird.
I'd do the Modify... line first, then the Run... and the password at the end.

Are we gonna add the ssh key in the tunnel command afterall?

cdonati:
:+1:

PR #2173, opened 2016-09-14
Don't add app if we have already marked it as loaded

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2025/

PR #2175, opened 2016-09-15
Java support for deleting Pull Queue tasks

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/2029/

### DIFF
--- a/AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
+++ b/AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
@@ -29,8 +29,6 @@ import com.google.apphosting.utils.remoteapi.RemoteApiPb.Request;
 import com.google.apphosting.utils.remoteapi.RemoteApiPb.Response;
 import com.google.apphosting.api.ApiProxy;
 import com.google.appengine.api.taskqueue.TaskQueuePb;
-import com.google.appengine.api.taskqueue.TaskQueuePb.TaskQueueAddRequest;
-import com.google.appengine.api.taskqueue.TaskQueuePb.TaskQueueAddRequest.Header;
 import com.google.apphosting.utils.config.AppEngineWebXml;
 
 public class AppScaleTaskQueueClient {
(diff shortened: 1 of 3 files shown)


AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
```diff
@@ -101,6 +99,24 @@ public AppScaleTaskQueueClient() {
...
+        deleteRequest.setAppId(getAppId());
+        Request request = new Request();
+        request.setMethod("Delete");
+        request.setServiceName(SERVICE_NAME);
+        request.setRequestAsBytes(deleteRequest.toByteArray());
+        Response response = sendRequest(request);
+        if (response.hasApplicationError()) {
+            throw new ApiProxy.ApplicationException(
+                response.getApplicationError().getCode(),
+                "TaskQueue Delete operation failed"
```
cdonati: Is there a way to use the protobuf detail field from the response instead of this string? We don't have any for delete calls right now, but we may in the future.

AppServer_Java/src/com/google/appengine/api/taskqueue/dev/DevPullQueue.java
```diff
@@ -77,8 +77,22 @@
...
+        logger.log(Level.FINE, "PullQueue: Sending DeleteRequest to TaskQueue server for " + queueName);
+        TaskQueuePb.TaskQueueDeleteResponse deleteResponse = client.delete(deleteRequest);
+
+        return deleteResponse;
+    }
+
+    // Abstract method needs to be overriden here.
     boolean deleteTask(String taskName) {
-        return this.taskMap.remove(taskName) != null;
+        throw new ApiProxy.ApplicationException(TaskQueuePb.TaskQueueServiceError.ErrorCode.INTERNAL_ERROR.getValue());
```
cdonati: Is this because we don't support this method yet? Can we include a message explaining that instead of a generic error?

menivaitsi: That is not used in the Pull Queues but since both DevPullQueue and DevPushQueue inherit from DevQueue it must be overridden here.

Unfortunately, there is no NOT_IMPLEMENTED error code in the TaskQueuePb class.
I can add a message. Thanks.

AppServer_Java/src/com/google/appengine/api/taskqueue/dev/AppScaleTaskQueueClient.java
menivaitsi: I couldn't find a way to get that with com.google.apphosting.utils.remoteapi.RemoteApiPb.Response.

We'd need to swap out the request exchange mechanism between the Java AppSever and the TaskQueue server to do that efficiently, which imho falls out of scope for this branch.

menivaitsi:
https://ocd.appscale.com:8080/job/Daily%20Build/2030/

PR #2180, opened 2016-09-21
Add Azure as a provider in the fast start script.

tmarballi:


scripts/fast-start.sh
```diff
@@ -100,6 +100,8 @@ fi
 # Let's try to detect the environment we are using.
 PUBLIC_IP=""
 PRIVATE_IP=""
+AZURE_METADATA=$(curl -s -o /dev/null -w "%{http_code}" http://169.254.169.254/metadata/v1/InstanceInfo)
```
menivaitsi: I'd move this up at the top where we define the Google Metadata service and keep it just as the endpoint.
You can do the curl command straight up in the check block. That way it follows the pattern for the rest of providers.

scripts/fast-start.sh
```diff
@@ -111,6 +113,8 @@ elif lspci | grep VirtualBox > /dev/null ; then
 elif ${CURL} -iLs metadata.google.internal |grep 'Metadata-Flavor: Google' > /dev/null ; then
     # As per https://cloud.google.com/compute/docs/metadata.
     PROVIDER="GCE"
+elif [ $AZURE_METADATA = "200" ]; then
```
menivaitsi: You can have the curl here.

PR #2182, opened 2016-09-30
Added a conntrack fallback in case that state is not availale.

christianbaun:
Some sources:
* http://unix.stackexchange.com/questions/108169/what-is-the-difference-between-m-conntrack-ctstate-and-m-state-state
* http://serverfault.com/questions/358996/iptables-whats-the-difference-between-m-state-and-m-conntrack


obino:
Thanks @christianbaun. Started and internal OCD run for this PR https://ocd.appscale.com:8080/job/Daily%20Build/2061/

cdonati:
:+1:

PR #2186, opened 2016-10-05
Upadte demo app, and use local app for faststart

obino:


scripts/fast-start.sh
```diff
@@ -19,8 +19,7 @@ IP="$(which ip)"
 APPSCALE_CMD="$(which appscale)"
 APPSCALE_UPLOAD="$(which appscale-upload-app)"
 GOOGLE_METADATA="http://169.254.169.254/computeMetadata/v1/instance/"
-GUESTBOOK_URL="http://www.appscale.com/wp-content/uploads/2014/07/guestbook.tar.gz"
-GUESTBOOK_APP="/root/guestbook.tar.gz"
+GUESTBOOK_APP="/root/appscale/AppServer/demos/appengine-guestbook-python"
```
menivaitsi: Can we have a $APPSCALE_HOME variable here instead of /root/appscale? 
I believe it's already available in an AppScale environment.

scripts/fast-start.sh
```diff
@@ -243,7 +237,7 @@ KEYNAME=$(grep keyname /root/AppScalefile | cut -f 2 -d ":")
 
 # Deploy sample app.
 [ -z "${ADMIN_EMAIL}" ] && ADMIN_EMAIL="a@a.com"
-[ -e ${GUESTBOOK_APP} ] && ${APPSCALE_UPLOAD} --keyname ${KEYNAME} --email ${ADMIN_EMAIL} --file ${GUESTBOOK_APP}
+[ -d ${GUESTBOOK_APP} ] && ${APPSCALE_UPLOAD} --keyname ${KEYNAME} --email ${ADMIN_EMAIL} --file ${GUESTBOOK_APP}
```
menivaitsi: What does the -d do here?

cdonati:
I would prefer not to introduce more diffs than necessary to the AppServer directory. Could we just include the fancy app on the disk image instead of putting it in the repo?

PR #2187, opened 2016-10-05
Validate payload when adding pull tasks with REST API

cdonati:
In the future, we will likely want to store the payload as a blob. For now, however, this fix gives us the same behavior as Google's API.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2072/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2073/

PR #2191, opened 2016-10-07
Use URL-safe base64 decoding with the REST API

cdonati:
Google does not accept '/' or '+' in the payloadBase64 field.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2078/

PR #2198, opened 2016-10-13
Include the cursor as part of the queryResult

cdonati:
While the compiled cursor is used by the datastore, the (other) cursor in this case is used by the API Proxy to keep track of which LiveQuery object to use.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2099/

PR #2199, opened 2016-10-13
AppDashboard Redesign

whoarethebritons:
loading from JSON string to templates in dashboard


AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +651,65 @@ def can_upload_apps(self):
...
+  def get_dash_layout_settings(self, email=None):
+    """ Queries the UserAppServer to see what settings the user has saved
+    for customizing the UI of the Dashboard
+
+    Returns:
+      A JSON string describing the customization layout
+    """
+    default_dash_layout = '''
+        {"nav" :
+              {"App Management":{"True":
```
cdonati: It's a good idea to indicate what each of the values mean in the data structure. There's nothing attached to "True" here to hint at what it's supposed to indicate.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +651,65 @@ def can_upload_apps(self):
...
+                {"Upload App":"/apps/new",
+                "Delete App":"/apps/delete"}},
+              "AppScale Management":{"True":
+                {"Cloud Status":"/status/cloud",
+                "Manage Users":"/authorize"}},
+              "Debugging/Monitoring":{"True":
+                {"Monit":"temp1",
+                "TaskQueue":"temp2",
+                "Logging":"/logs"}}}
```
cdonati: Since you would like these to be ordered, I would recommend the following (Python code):
```
{'nav': [
    {'heading': 'App Management',
     'visible': True,  # Is this needed? Could the section just be removed from the structure before it's sent to the client?
     'buttons': {'Upload App': '/apps/new', 'Delete App': '/apps/delete'}},
    {'heading': 'AppScale Management',
     'visible': True,
     'buttons': {'Cloud Status': '/status/cloud'}},
    ...
]}
```

AppDashboard/static/test.html
```diff
@@ -0,0 +1,608 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en"><!-- FILE:templates/layouts/main.html --><head>
```
cdonati: Is this file being used?

AppDashboard/templates/layouts/panel.html
```diff
@@ -0,0 +1,26 @@
+<div class="panel panel-default" id="panel{{ panel_count }}">
```
cdonati: I would recommend basing the panel names on their heading name rather than their count. You could include a attribute-friendly ID in the JSON structure.

AppDashboard/templates/shared/navigation.html
```diff
@@ -37,30 +46,42 @@
...
+                            {% for permission_val, title in sublist.items() %}
+                                {% if permission_val == "True" %}
+                                        <div id="lists" class="panel panel-default">
+                                            <div class="panel-heading">
+                                                <h4 class="panel-title">
+                                                    <a data-toggle="collapse" href="#collapse{{ count }}"> {{ main_list_title }}</a>
+                                                </h4>
+                                            </div>
+                                        </div><div>
+                                        <div id="collapse{{ count }}" class="in">
```
cdonati: These should also have a more friendly and stable ID.

AppDashboard/templates/apps/database.html
```diff
@@ -0,0 +1,88 @@
...
+                <h4><span class="icon12 icomoon-icon-database"></span>Current Database:</h4>
+                <!-- DB Hook -->
+                <ul class="unstyled">
+                    <li id="db-img"><span class="icon16 icomoon-icon-arrow-right-2"></span><strong>Powered by
+                        {% if dbinfo['table'] == 'cassandra' %}
+                        <img id="Cassandra" src="/images/cassandra.jpg" alt="Cassandra"/>
+                        {% elif dbinfo['table'] == 'hbase' %}
+                        <img id="Hbase" src="/images/hbase.png" alt="Hbase"/>
+                        {% elif dbinfo['table'] == 'hypertable' %}
+                        <img id="Hypertable" src="/images/hypertable.jpg" alt="Hypertable">
```
cdonati: I know you didn't add these, but the references and assets for Hbase and Hypertable can be removed.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +651,65 @@ def can_upload_apps(self):
...
       return False
+
+  def get_dash_layout_settings(self, email=None):
+    """ Queries the UserAppServer to see what settings the user has saved
+    for customizing the UI of the Dashboard
+
+    Returns:
+      A JSON string describing the customization layout
+    """
+    default_dash_layout = '''
```
whoarethebritons: my reasoning for making the json string was just proof that I could load from one, the json string will never be constructed in python it will actually be constructed in javascript and sent to the python server to be stored in the datastore

AppDashboard/dashboard.py
```diff
@@ -121,6 +123,7 @@ def render_template(self, template_file, values=None):
       'is_user_cloud_admin': self.dstore.is_user_cloud_admin(),
       'can_upload_apps': self.dstore.can_upload_apps(),
       'apps_user_is_admin_on': owned_apps,
+      'user_layout_pref': json.loads(self.dstore.get_dash_layout_settings(), object_pairs_hook=OrderedDict),
```
menivaitsi: Please limit all code lines to 80 characters.
In this case you could modify the line to be:

```
'user_layout_pref': json.loads(self.dstore.get_dash_layout_settings(), 
                               object_pairs_hook=OrderedDict),
```

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: because I was planning on loading from json, I didn't see a way to make python load it as a list... it said it would be a dictionary

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +651,65 @@ def can_upload_apps(self):
     except Exception as err:
       logging.exception(err)
       return False
+
+  def get_dash_layout_settings(self, email=None):
+    """ Queries the UserAppServer to see what settings the user has saved
+    for customizing the UI of the Dashboard
```
menivaitsi: Please add a period at the end of the sentence.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +651,65 @@ def can_upload_apps(self):
...
+
```
menivaitsi: Add Args section to describe the method arguments (email in this case).

AppDashboard/static/css/custom.css
```diff
@@ -1921,4 +1907,31 @@ footer {
...
+}
+.mainnav .sortable {
+    width:100%;
+    overflow:auto;
+}
+.mainnav li a.glyphicon.btn {
+    padding: 4px;
+    float:right;
+    width: 10%;
+}.mainnav li a{
```
menivaitsi: Add new line before next css rule.

AppDashboard/static/js/dash.js
```diff
@@ -0,0 +1,6 @@
+$(document).ready(function(){
+    //in case there's any later special implementations with multiple areas that can be dragged
+    //$( ".selector" ).droppable({accept: ".special"});
+
+    $(".sortable").sortable({cursorAt:{top:3}, cursor:"move", helper:"clone"});
+});
```
menivaitsi: Add new line at end of file.

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: for right now "True" is a placeholder so I don't have to be logged in, other than that its meant to be whether the user can see that category

AppDashboard/templates/shared/navigation.html
```diff
@@ -5,6 +5,15 @@
                 <a id="appscale_logo" href="/"><img src="/images/appscale-logo.png?v=2"/></a>
               <div class="container-fluid">
                 <div class="nav-no-collapse">
+                    <a href="mailto:support@appscale.com" class="btn btn-primary btn-large">Send Feedback</a>
+                    &nbsp;
+                    <a href="https://github.com/AppScale/appscale/issues" class="btn btn-info btn-large" target="_blank">Submit Issue</a>
```
menivaitsi: This button should link to: https://www.appscale.com/resources/tutorials/redmine

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: so that "True" is eventually is_user_cloud_admin or can_upload_apps or different combinations of those two. If I manage it in the template I have to test which heading I have to see which variables I need to use.

AppDashboard/lib/app_dashboard_data.py
cdonati: The syntax for JSON and Python code is almost identical. The JSON string `{}` will be loaded as a Python dict whereas the JSON string `[]` will be loaded as a Python list.

AppDashboard/lib/app_dashboard_data.py
cdonati: Right. I'm not suggesting managing it in the template. You can remove it (or include it) by testing it against those variables before rendering the template. In my opinion, it's best to keep logic out of the template whenever possible.

whoarethebritons: So... test if each of those keys are in the list... if they aren't allowed to see them remove them?

AppDashboard/static/test.html
whoarethebritons: its just a mockup it will be removed by the final pull request

AppDashboard/templates/layouts/panel.html
whoarethebritons: I just realized I forgot to do this one, however I'm having difficulty having that and having panels being added (since those won't have the correct id) unless I make the dictionary have another item for the nav area that has the id that it can use to look up in the look up dictionary... thoughts?

whoarethebritons: I changed it don't worry about it.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -170,6 +172,75 @@ def __init__(self, helper=None):
     """
     self.helper = helper or AppDashboardHelper()
 
+  def build_dict(self):
+    LOOKUP_DICT = {}
+    LOOKUP_DICT['cloud_stats'] = {"title":"Cloud Status",
```
cdonati: I think this function would be more readable if the keys were included in the dict literal definition:
```
LOOKUP_DICT = {
  'cloud_status': {'title': 'Cloud Status', ...},
  'database_stats': {'title: 'Database', ...},
  ...
}
```

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -170,6 +172,75 @@ def __init__(self, helper=None):
...
+                                   "template":"authorize/cloud.html"}
+    LOOKUP_DICT['logging'] = {"title":"Logs",
+                              "link":"/logs",
+                              "template":"logs/main.html"}
+    LOOKUP_DICT['monit'] = {"title":"Monit",
+                             "link":self.get_monit_url()}
+    LOOKUP_DICT['taskqueue'] = {"title":"TaskQueue",
+                             "link":self.get_flower_url()}
+
+    LOOKUP_DICT['app_management'] = [self.helper.can_upload_apps(),
```
cdonati: Does this key need to be included in the return dict at all if `self.helper.can_upload_apps()` evaluates to `False`?
It would make much more sense to me if it looked like
```
if self.helper.can_upload_apps():
  LOOKUP_DICT['app_management'] = {'title': 'App Mangement' ...}
```

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -170,6 +172,75 @@ def __init__(self, helper=None):
...
+                                     {"App Management":
+                                        [{"upload_app": LOOKUP_DICT[
+                                          'upload_app']},
+                                         {"delete_app": LOOKUP_DICT[
+                                           'delete_app']}]}]
+    # TODO: remove, used temporarily so user does not have to log in
+    LOOKUP_DICT['app_management'] = ["True",
```
cdonati: When dealing with boolean values, use Python's `True` and `False`. You can then test it directly with `if value` rather than a string equality like `if value == 'True'`.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +721,45 @@ def can_upload_apps(self):
...
+      A JSON string describing the customization layout
+    """
+    LOOKUP_DICT = self.build_dict()
+    #base admin template
+    default_dash_layout ='''{
+        "nav":["app_management","appscale_management","debugging_monitoring"],
+        "panel":["cloud_stats","database_stats","memcache_stats"]
+        }'''
+    temp_dict = json.loads(default_dash_layout)
+    temp_dict['nav'][:] = [{key:LOOKUP_DICT.get(str(key))[1]} for key in
```
cdonati: The slice syntax is only necessary if you need to modify the list itself (eg. if some other variable is referencing it directly and you need to change that reference). Since you are creating a whole new list for this key here, you can just set it directly with `temp_dict['nav'] = ...`.

All of the keys are strings anyway, so there's no need for the `str` function.

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: ok, but for the navigation parts it should stay as is right?

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: ok, I thought that would cause issues for some reason, I'll change it

AppDashboard/lib/app_dashboard_data.py
cdonati: You can change the conditional part of your list comprehension from `if LOOKUP_DICT.get(str(key))[0] == "True"` to something like `if key in LOOKUP_DICT` so that it doesn't even try to retrieve the key.

AppDashboard/dashboard.py
```diff
@@ -1128,12 +1204,57 @@ def get(self):
...
+
+class AjaxSaveLayoutSettings(AppDashboard):
+  """ Class that stores dashboard layout settings in the Datastore. """
+
+  def post(self):
+    """ sets the dashboard layout settings """
+
+    nav = self.request.get("nav")
+    panel = self.request.get("panel")
+    json_string = '{"nav":'+nav+',"panel":'+panel+"}"
```
cdonati: I would recommend dealing with python-native structures as soon as you can on the server side. For example, `set_dash_layout_settings` should probably only accept a dict for `values`, not a string.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+                            key in LOOKUP_DICT]
+
+        temp_dict['panel'] = [{key: LOOKUP_DICT.get(key)} for key in
+                              temp_dict.get('panel') if
+                              key in LOOKUP_DICT]
+        user_info.dash_layout_settings = temp_dict
+        user_info.put()
+      return
+    except Exception as err:
+      logging.exception(err)
```
cdonati: This is too generic. The calling function will have no idea that something went wrong.

AppDashboard/dashboard.py
```diff
@@ -1157,7 +1279,10 @@ def get(self):
   ('/logs/(.+)', LogServicePage),
   ('/gather-logs', LogDownloader),
   ('/groomer', RunGroomer),
-  ('/change-password', ChangePasswordPage)
+  ('/change-password', ChangePasswordPage),
+  ('/ajax/render/panel', AjaxRenderPanel),
+  ('/ajax/save/layout', AjaxSaveLayoutSettings),
+  ('/ajax/reset/layout', AjaxResetLayoutSettings)
```
cdonati: URLs should usually be constructed in increasing specificity. In this case, `/ajax/layout/reset` and `/ajax/layout/save` would make more sense.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -171,6 +173,77 @@ def __init__(self, helper=None):
...
+          LOOKUP_DICT["debugging_monitoring"] = {"Debugging/Monitoring":
+                                                [{"monit": LOOKUP_DICT["monit"]},
+                                                 {"taskqueue": LOOKUP_DICT[
+                                                   "taskqueue"]},
+                                                 {"logging": LOOKUP_DICT[
+                                                   "logging"]},
+                                                 {"app_console": LOOKUP_DICT[
+                                                   "app_console"]}]}
+        return LOOKUP_DICT
+    except Exception as err:
```
cdonati: Which exceptions is this catching? It's usually better to avoid catching `Exception` when possible.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -555,14 +628,22 @@ def update_users(self):
           is_user_cloud_admin = self.helper.is_user_cloud_admin(email)
           can_upload_apps = self.helper.can_upload_apps(email)
           owned_apps = self.helper.get_owned_apps(email)
+          dash_layout_settings = self.get_dash_layout_settings(email)
+
+          dash_change = (dash_layout_settings.get("nav") !=
+                         user_info.dash_layout_settings.get("nav")) and \
```
cdonati: It seems strange to me that `dash_change` would only be true if _both_ have changed.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+          temp_dict = json.loads(values)
+        except TypeError as err:
+          #base admin template
+          default_value = '''{
+            "nav":["app_management", "appscale_management",
+            "debugging_monitoring"],
+            "panel":["app_console","upload_app","cloud_stats","database_stats",
+            "memcache_stats"]
+            }'''
+          temp_dict = json.loads(default_value)
```
cdonati: Just create the dict directly.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+    user = users.get_current_user()
+    if not user:
+      return
+    email = user.email()
+    try:
+      user_info = self.get_by_id(UserInfo, email)
+      if user_info:
+        try:
+          temp_dict = json.loads(values)
+        except TypeError as err:
```
cdonati: In addition to catching invalid JSON, you should also verify that there are no unexpected keys or values. If a user without appropriate permissions posted values that only a cloud admin should be able to see, it probably shouldn't be written to the database even though you filter it when retrieving that data.

It would make sense to me to have the validation happen in the post handler, but it could also go here instead.

As a side note, I think it would make more sense for this function to raise an exception on invalid input rather than continue on with default values as if nothing went wrong.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+            values = user_info.dash_layout_settings
+            default_nav = ["app_management","appscale_management",
+                           "debugging_monitoring"]
+
+            nav_list=[]
+            for key_dict in values.get('nav'):
+              for temp_key in key_dict:
+                nav_list.append(temp_key)
+
+            print(nav_list)
```
cdonati: This should probably be removed or replaced with a logging call if you need it in production.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+            if set(nav_list)!=set(default_nav):
+              for key in default_nav:
+                if nav_list.count(key)==0:
```
cdonati: `if key not in nav_list` is more idiomatic.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+      email: A str that indicates the e-mail address of the administrator of
+        this application, or None if the currently logged-in user is the admin.
+
+    Returns:
+      A dictionary containing the customization layout
+    """
+    if email is None:
+      user = users.get_current_user()
+      if not user:
+        return
```
cdonati: Some places that call this function assume the output is a dictionary. Returning `None` will cause problems.

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -395,6 +395,35 @@ def upload_app(self, filename, upload_file):
...
+    """ Relocates a Google App Engine application to different ports.
+
+      Args:
+        appid: The application to be relocated
+        http_port: The HTTP Port to relocate the application to
+        https_port: The HTTPS Port to relocate the application to
+      Returns:
+        A str indicating that the application was relocated successfully.
+      Raises:
+        AppHelperException: If the application was not uploaded successfully.
```
cdonati: Left over from a copy/paste?

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -395,6 +395,35 @@ def upload_app(self, filename, upload_file):
...
+    acc = self.get_appcontroller_client()
+    try:
+      relocate_info = acc.relocate_app(appid, http_port, https_port)
+      logging.info(relocate_info)
+      # Returns:
+      #"OK" if the relocation occurred successfully, and a String containing
+      # the reason why the relocation failed in all other cases.
+      if(relocate_info!="OK"):
+        logging.error("AppController returned: {0}".format(relocate_info))
+        return "Error attempting to relocate Application: {0}"\
```
cdonati: This should probably raise an exception so that the calling function knows there was an error. Otherwise it would need to parse the return string.

AppDashboard/static/plugins/forms/tiny_mce/plugins/spellchecker/editor_plugin_src.js
```diff
@@ -167,7 +167,7 @@
 		},
 
 		_getSeparators : function() {
-			var re = '', i, str = this.editor.getParam('spellchecker_word_separator_chars', '\\s!"#$%&()*+,-./:;<=>?@[\]^_{|}���������������\u201d\u201c');
+			var re = '', i, str = this.editor.getParam('spellchecker_word_separator_chars', '\\s!"#$%&()*+,-./:;<=>?@[\]^_{|}����������������\u201d\u201c');
```
cdonati: Not sure what this change is doing. Did your editor auto-replace these characters?

AppDashboard/static/plugins/forms/tiny_mce/plugins/table/js/table.js
```diff
@@ -242,16 +242,16 @@ function insertTable() {
 	} else
 		inst.execCommand('mceInsertContent', false, html);
 
-	tinymce.each(dom.select('table[data-mce-new]'), function(node) {
+	tinymce.each(dom.select('table[data-mce-new]'), function(node) {
```
cdonati: Try to avoid committing changes like this unless you are reducing diffs with an external library.

AppDashboard/templates/apps/console.html
```diff
@@ -1,27 +1,82 @@
...
+                            } else {
+                                apps_html += '<a class="btn" href="' + apps[key][0] + '" target="_blank">' + key + '</a> ';
+                                apps_html += '<a class="btn" href="' + apps[key][1] + '" target="_blank">' +'https-'+ key + '</a>';
+                            }
+                            apps_html += '</li>';
+                        }
+                    }
+                    $('#application_info ul').html(apps_html);
+                });
+            }, 10000);
```
cdonati: Handle errors when making http requests.

AppDashboard/templates/layouts/panel.html
```diff
@@ -0,0 +1,29 @@
...
+    </h2>
+
+  </div><!--end .panel-heading-->
+  <div id="{{ id }}" class="panel-collapse collapse in">
+    <div class="panel-body">
+      <div class="content">
+        {% include page_info.template %}
+        <script type="text/javascript">
+          $(document).ready(function(){
+              $(".well").removeClass("well");
```
cdonati: Could this class be removed before rendering the page, or is the template included in other places where it's needed?

AppDashboard/templates/logs/main.html
```diff
@@ -2,14 +2,17 @@
 <!--Body content-->
 <div id="content" class="clearfix">
     <div class="contentwrapper"><!--Content wrapper-->
-
+        {% set app_page_title="Log Viewer" %}
```
cdonati: Could this be passed in instead of defined here? Using the string directly below would also be fine with me.

AppServer/lib/django-0.96/django/conf/locale/es/LC_MESSAGES/django.po
```diff
@@ -2362,7 +2362,7 @@ msgstr "El %(verbose_name)s se ha creado correctamente."
 #: views/generic/create_update.py:117
 #, python-format
 msgid "The %(verbose_name)s was updated successfully."
-msgstr "Se actualiz� con �xito el %(verbose_name)s."
+msgstr "Se actualiz� con �xito el %(verbose_name)s."
```
cdonati: lol. Please undo all the diffs on this file.

AppDashboard/static/plugins/forms/tiny_mce/plugins/table/js/table.js
whoarethebritons: so what actually happened with this was my computer keyboard was lagging and somehow selected the plugins folder to commit. I tried to undo it and apparently I didn't. I made no changes with the plugins folder... how can I remove it

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +729,125 @@ def can_upload_apps(self):
...
+      pass
+
+
+  def rebuild_dash_layout_settings_dict(self, email=None):
+    """ Rebuilds the user's layout settings in case there is an update
+      to the lookup dictionary
+
+    Args:
+      email: A str that indicates the e-mail address of the administrator of
+        this application, or None if the currently logged-in user is the admin.
```
whoarethebritons: it should be None if the user is currently logged in. email doesn't get passed most of the time.

AppDashboard/static/plugins/forms/tiny_mce/plugins/table/js/table.js
cdonati: In this case, you can do `git diff HEAD master AppDashboard/static/plugins/forms/tiny_mce/plugins > plugins.diff` and then `git apply plugins.diff`. You can remove the diff file after. That will undo any changes your branch introduced from master for that directory.

If you notice before pushing the branch publicly, you can undo a commit with `git reset HEAD~1`.

AppDashboard/lib/app_dashboard_data.py
cdonati: Okay, the docstring makes is sound like I need to explicitly pass in None in cases where the user is the admin. If I understand correctly, you were trying to indicate that the email parameter is not required if the user is logged in (since it will be auto-populated). I think you can just omit the "or None" clause.

AppDashboard/dashboard.py
```diff
@@ -105,16 +105,21 @@ def render_template(self, template_file, values=None):
     if values is None:
       values = {}
 
-    owned_apps = self.dstore.get_owned_apps()
-    self.helper.update_cookie_app_list(owned_apps, self.request, self.response)
+    is_cloud_admin = self.helper.is_user_cloud_admin()
+    if is_cloud_admin:
+      apps_user_is_admin_on = self.dstore.get_application_info()
+    else:
+      apps_user_is_admin_on = self.helper.get_owned_apps()
```
whoarethebritons: apps_user_is_admin on is a dictionary of the applications that is used in relocate, delete, and app stats. I'm fine with a rename but the appids are needed in all of those.

AppDashboard/dashboard.py
```diff
@@ -105,16 +105,21 @@ def render_template(self, template_file, values=None):
...
+    self.helper.update_cookie_app_list(apps_user_is_admin_on, self.request, self.response)
```
whoarethebritons: Ah! I thought I had anticipated all of them that would need a list instead, I'll add .keys() and it'll be good.

AppDashboard/lib/app_dashboard_data.py
whoarethebritons: Its catching the user_info exception. Or at least I assume so, every other function that made the first if statements and tried to retrieve user_info catch an exception. This method needs to check user_info variables otherwise it will not have any of the navigation keys. That was the bug I chased down last week.

AppDashboard/templates/logs/main.html
whoarethebritons: The title of the page? Its used as the panel title or the app_page title if app_page sets standalone_app_page to True. Rather than doing jQuery calls to grab the inner HTML of the heading, I decided to set a variable that both of those templates use. The reason for the if statement is simply that I can't access app_page_title in the app_page template where I would need to.

AppDashboard/dashboard.py
cdonati: You're right. I misread that. You can ignore my comment.

AppDashboard/dashboard.py
cdonati: In this case, I think it's better to fail fast and let the client know that their request was invalid. Otherwise, if for some reason there's a JavaScript bug somewhere down the line, it could look like the post is succeeding but having no effect. It would be clearer if the response from the post returned a 400 or something.

AppDashboard/lib/app_dashboard_data.py
cdonati: I'm not sure what you mean by the user_info exception. Do you have an example stacktrace that occurs when you remove that except block?

AppDashboard/templates/logs/main.html
cdonati: Cool. That works. I didn't know it was used elsewhere.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2209/

AppDashboard/lib/app_dashboard_helper.py
whoarethebritons: This error string is displayed to the user.

AppDashboard/templates/layouts/panel.html
whoarethebritons: The class isn't necessarily needed, its just a styling thing. I could definitely remove it in the templates. The class is currently used when the url is used rather than the panel.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2220/

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +735,152 @@ def can_upload_apps(self):
...
+            if set(panel_list)!= set(default_panel):
+              for key in default_panel:
+                if panel_list.count(key)==0:
+                  panel_list.append(key)
+
+            values['nav'] = [{key: LOOKUP_DICT.get(key)} for key in
+                                nav_list if
+                                key in LOOKUP_DICT]
+
+            values['panel'] = [{key: LOOKUP_DICT.get(key)} for key in
```
menivaitsi: Please rewrite as follows if you want to keep the list comprehension:

```
            is_admin_panel = LOOKUP_DICT.get(key).get('is_admin_panel')
            values['panel'] = [
              {key: LOOKUP_DICT.get(key)} for key in panel_list
              if key in LOOKUP_DICT and (not is_admin_panel or \
                 is_admin_panel == user_info.is_user_cloud_admin)
            ]
```

menivaitsi: Don't be afraid to use local variables for large accessors like `LOOKUP_DICT.get(key).get('is_admin_panel')`.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +735,152 @@ def can_upload_apps(self):
...
+    except Exception as err:
+      logging.exception(err)
+
+
+  def get_dash_layout_settings(self, user_info=None):
+    """ Queries the UserAppServer to see what settings the user has saved
+    for customizing the UI of the Dashboard.
+
+    Args:
+      user_info: The current user
```
menivaitsi: Add missing period.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -647,3 +735,152 @@ def can_upload_apps(self):
...
+
+    Returns:
+      A dictionary containing the customization layout
```
menivaitsi: Same here.

AppDashboard/lib/app_dashboard_helper.py
```diff
@@ -395,6 +395,35 @@ def upload_app(self, filename, upload_file):
...
+        AppHelperException: If the application was not uploaded successfully.
+      """
+    acc = self.get_appcontroller_client()
+    try:
+      relocate_info = acc.relocate_app(appid, http_port, https_port)
+      logging.info(relocate_info)
+      # Returns:
+      #"OK" if the relocation occurred successfully, and a String containing
+      # the reason why the relocation failed in all other cases.
+      if(relocate_info!="OK"):
```
menivaitsi: Add space around `!=`.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2232/

AppDashboard/dashboard.py
```diff
@@ -1157,7 +1290,10 @@ def get(self):
   ('/logs/(.+)', LogServicePage),
   ('/gather-logs', LogDownloader),
   ('/groomer', RunGroomer),
-  ('/change-password', ChangePasswordPage)
+  ('/change-password', ChangePasswordPage),
+  ('/ajax/render/panel', AjaxRenderPanel),
```
cdonati: For consistency, use `/ajax/panel/render`.

AppDashboard/static/js/dash.js
```diff
@@ -0,0 +1,86 @@
...
+    /*enable tooltips*/
+    $('[data-toggle="tooltip"]').tooltip();
+
+    /*add panel functionality*/
+    $(".add-panel").click(function() {
+        var newPanelID = "#" + $(this).attr("data-target");
+        if(!$(newPanelID).length){
+            $.ajax({
+                method: "post",
+                url: "/ajax/render/panel",
```
cdonati: /ajax/panel/render

AppDashboard/static/js/dash.js
```diff
@@ -0,0 +1,86 @@
...
+        $(".nav-heading-collapse").each(function(){
+            nav_array.push($(this).attr("href").match("\#(.*)")[1]);
+        });
+        var panel_array = [];
+        $("#dash-panels .panel-collapse").each(function(){
+            panel_array.push($(this).attr("id"));
+        });
+        $.ajax({
+            method: "post",
+            url:"/ajax/save/layout",
```
cdonati: /ajax/layout/save

AppDashboard/static/js/dash.js
```diff
@@ -0,0 +1,86 @@
...
+            error: function (result) {
+                $("#save-layout").html(result);
+            }
+        });
+    });
+    /*reset layout functionality*/
+    $("#reset-layout").click(function() {
+        $.ajax({
+            method:"post",
+            url:"/ajax/reset/layout",
```
cdonati: /ajax/layout/reset

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2248/

PR #2201, opened 2016-10-14
Prefix nginx entries

cdonati:
This allows us to remove the files that we've added without removing other files in that directory.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2104/

PR #2202, opened 2016-10-16
Fix Go unit test

cdonati:
This patch was taken from go-review.googlesource.com/#/c/29995/.
See also github.com/golang/go/issues/17276.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2115/

PR #2205, opened 2016-10-18
Use json instead of simplejson

cdonati:
Our build broke because simplejson was not installed.

Apparently, we had it installed by accident for one of our dependencies. Now that it's no longer installed, I think it's preferable to use the native json library. The only downside is that Python < 2.6 will not be supported, but we require that anyway for the tools.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2126/
https://ocd.appscale.com:8080/job/Euca-Faststart/283/

PR #2216, opened 2016-11-02
Download custom client jars during build

cdonati:
This will speed up the build process. I've linked to the source repos so that building them is reproducible.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2188/

PR #2219, opened 2016-11-07
Replace appscale-taskqueue script with entry point

cdonati:
The main reason for this pull request is that `pip install`ing a package after it was installed as an egg (eg. after a `python setup.py install` like we used to do) was not updating the script in /usr/local/bin. Another possibility would have been to add the `--egg` flag, but that has been deprecated as of pip 9.0.0.

This fix has the added bonus of being the recommended way to package console scripts: https://packaging.python.org/distributing/#scripts


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2208/

cdonati:
Successful tools-test run: https://ocd.appscale.com:8080/job/Tools-test/1110/

PR #2220, opened 2016-11-08
Give the Java AppServer a max heap size

cdonati:
The max heap size is set to 140MB less than the app's max memory. In practice, the process tends to use about 120MB more (according to monit) than the max heap size given to Java.

Previously, the default heap size was used, which was normally ~1GB. This change should prevent monit from killing the Java AppServer processes. Instead, the app will encounter an OutOfMemoryError if it tries to allocate more than the max heap size allows.

### DIFF
--- a/AppManager/app_manager_server.py
+++ b/AppManager/app_manager_server.py
@@ -226,10 +226,18 @@ def start_app(config):
     copy_successful = copy_modified_jars(config['app_name'])
     if not copy_successful:
       return BAD_PID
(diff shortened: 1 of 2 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2222/

cdonati:
Successful tools-test: https://ocd.appscale.com:8080/job/Tools-test/1119/

PR #2221, opened 2016-11-08
Replace batch writes to the transaction table

cdonati:
As the transaction table is just temporary ongoing transaction mutation information, writes to it do not need to be atomic. This branch also makes large batch writes concurrent, which results in a slight increase in performance.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2223/

PR #2229, opened 2016-11-18
Increase padding for max heap size

cdonati:
140MB was not enough padding. This was most apparent when fetching a large amount of data from the datastore.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2267/

PR #2239, opened 2016-12-01
Makes all rabbitmq slaves RAM nodes.

obino:
Not ready to merge.


obino:
Not good for now: we need to rework the way the tq requests move around. At this time the load is balanced across nodes, so we will look into this at a later time.

menivaitsi:
We'll reduce the number of RabbitMQ nodes we're clustering. This is not relevant anymore.

PR #2241, opened 2016-12-01
Appscale dashboard cron page

RaiaN:
Hi!
I implemented basic system to view/run cron jobs from Appscale dashboard. It looks like this:
cron_console
cron_tasks

What do you think about it?

I still have one issue about how to give an access to /etc/cron.d/ and /var/apps/ directories for appscaledashboard application. I did this through by modifying the file  
`AppServer/google/appengine/tools/devappserver2/python/sandbox.py`. 

Peter


RaiaN:
I will squash all commits to a few to avoid bad history

PR #2247, opened 2016-12-04
Ensure we have the correct path in monit

obino:
if/when the appcontroller crash or is killed, the path in monit doesn't
consider /usr/local/bin, creating the possibility of some of the scripts
to be unreachable.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2332/

PR #2248, opened 2016-12-04
Set property

obino:
This PR enables the ability to change properties set in the AppScalefile (for example verbose) at run time. At this time some of these changes won't be in effect (for example changing the flower_password would require a restart of flower, and we don't handle that yet), but the basics ones works.

We also enabled an old flag not in function: autoscale. By default is on, but if turned off, the AC won't add or remove appservers anylonger.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -3590,7 +3590,7 @@ class Djinn
         if my_node.is_zookeeper?
           configure_zookeeper(@nodes, @my_index)
           begin
-            start_zookeeper(@options['clear_datastore'].downcase == "true")
+            start_zookeeper(false)
           rescue FailedZooKeeperOperationException
             @state = "Couldn't start Zookeeper."
             HelperFunctions.log_and_crash(@state, WAIT_TO_CRASH)
@@ -3616,15 +3616,14 @@ class Djinn
 
       threads << Thread.new {
         Djinn.log_info("Starting database services.")
-        clear_datastore = @options['clear_datastore'].downcase == "true"
         db_nodes = @nodes.count{|node| node.is_db_master? or node.is_db_slave?}
         needed_nodes = needed_for_quorum(db_nodes,
                                          Integer(@options['replication']))
         if my_node.is_db_master?
-          start_db_master(clear_datastore, needed_nodes, db_nodes)
+          start_db_master(false, needed_nodes, db_nodes)
           prime_database
         else
-          start_db_slave(clear_datastore, needed_nodes, db_nodes)
+          start_db_slave(false, needed_nodes, db_nodes)
         end
       }
     end
@@ -3807,13 +3806,12 @@ class Djinn
   end
 
   def start_search_role()
-    Search.start_master(@options['clear_datastore'].downcase == "true")
+    Search.start_master(false)
   end
 
   def start_taskqueue_master()
-    clear_datastore = @options['clear_datastore'].downcase == "true"
     verbose = @options['verbose'].downcase == "true"
-    TaskQueue.start_master(clear_datastore, verbose)
+    TaskQueue.start_master(false, verbose)
     return true
   end
 
@@ -3825,9 +3823,8 @@ class Djinn
       master_ip = node.private_ip if node.is_taskqueue_master?
     }
 
-    clear_datastore = @options['clear_datastore'].downcase == "true"
     verbose = @options['verbose'].downcase == "true"
-    TaskQueue.start_slave(master_ip, clear_datastore, verbose)
+    TaskQueue.start_slave(master_ip, false, verbose)
     return true
   end
 
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -1405,12 +1405,10 @@ class Djinn
     return BAD_SECRET_MSG unless valid_secret?(secret)
 
     properties = {}
-    instance_variables.each { |name|
-      name_without_at_sign = name[1..name.length-1]
+    @options.keys{ |key|
       begin
-        if name_without_at_sign =~ /\A#{property_regex}\Z/
-          value = instance_variable_get(name)
-          properties[name_without_at_sign] = value
+        if key =~ /\A#{property_regex}\Z/
+          properties[key] = @options[key]
         end
       rescue RegexpError
         Djinn.log_warn("get_property: got invalid regex (#{property_regex}).")
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -458,7 +458,6 @@ class Djinn
     'controller_logs_to_dashboard' => [ TrueClass, 'False' ],
     'appengine' => [ Fixnum, '2' ],
     'autoscale' => [ TrueClass, 'True' ],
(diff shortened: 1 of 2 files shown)


obino:
OCD https://ocd.appscale.com:8080/job/Daily%20Build/2333/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2336/

AppController/djinn.rb
```diff
@@ -938,19 +932,100 @@ def set_parameters(layout, options, secret)
...
+    @@log.level = new_level if @@log.level != new_level
+
+    # Make sure flower is running with the proper password.
+    TaskQueue.stop_flower
+    TaskQueue.start_flower(@options['flower_password']) if my_node.is_shadow?
+  end
+
+  # Validates and sets the instance variables that Djinn needs before it can
+  # begin configuring and deploying services on a given node (and if it is the
+  # first Djinn, starting up the other Djinns).
```
menivaitsi: Replace Djinn with AppController please.

AppController/djinn.rb
```diff
@@ -938,19 +932,100 @@ def set_parameters(layout, options, secret)
...
+  #
+  # Args:
+  #   layout: this is a JSON structure containing the nodes
+  #     informations (IPs, roles, instance ID etc...). These are the nodes
```
menivaitsi: informations -> information

AppController/djinn.rb
```diff
@@ -938,19 +932,100 @@ def set_parameters(layout, options, secret)
...
+    TaskQueue.stop_flower
+    TaskQueue.start_flower(@options['flower_password']) if my_node.is_shadow?
+  end
+
+  # Validates and sets the instance variables that Djinn needs before it can
+  # begin configuring and deploying services on a given node (and if it is the
+  # first Djinn, starting up the other Djinns).
+  #
+  # Args:
+  #   layout: this is a JSON structure containing the nodes
```
menivaitsi: nodes -> node

AppController/djinn.rb
```diff
@@ -938,19 +932,100 @@ def set_parameters(layout, options, secret)
...
+    elsif opts.class != Hash
+      msg = "Error: options is not a Hash."
+      Djinn.log_error(msg)
+      return msg
+    else
+      @options = check_options(opts)
+    end
+
+    # Let's validate we have the needed options defined.
+    ['keyname', 'login', 'table'].each { |key|
```
menivaitsi: Is 'table' really needed? Don't we have it set to cassandra everywhere?

AppController/djinn.rb
```diff
@@ -1372,21 +1457,70 @@ def get_property(property_regex, secret)
...
+      # We give some extra information to the user about some properties.
+      if key == "keyname"
+        Djinn.log_warn("Changing keyname can break your deployment!")
+      end
+      if key == "max_memory"
+        Djinn.log_warn("max_memory will be enforced on new appservers only.")
+      end
+      if key == "min_images"
+        unless is_cloud?
+          Djinn.log_warn("min_images is not used in non-cloud infrastructures.")
```
menivaitsi: Since cloud/cluster distinction is going away, I suggest we change this to something like:

`"[...] not used in agent-managed infrastructures"`
or
`"[...] not used in managed infrastructures"`

AppController/djinn.rb
```diff
@@ -1372,21 +1457,70 @@ def get_property(property_regex, secret)
...
+        end
+      end
+      if key == "max_images"
+        unless is_cloud?
+          Djinn.log_warn("max_images is not used in non-cloud infrastructures.")
```
menivaitsi: Same here.

AppController/djinn.rb
```diff
@@ -2966,7 +3105,10 @@ def restore_appcontroller_state()
 
     # Now that we've restored our state, update the pointer that indicates
     # which node in @nodes is ours
-    find_me_in_locations()
+    find_me_in_locations
```
menivaitsi: I thought we'd agreed to keep `()` around for readability.

AppController/djinn.rb
```diff
@@ -5359,7 +5461,8 @@ def get_scaling_info_for_app(app_name, update_dashboard=true)
     # Let's make sure we have the minimum number of AppServers running.
     Djinn.log_debug("Evaluating app #{app_name} for scaling.")
     if (@app_info_map[app_name]['appengine'].nil? ||
```
menivaitsi: Can we assign a variable to `@app_info_map[app_name]['appengine']` so that the length of this if decreases?

AppController/djinn.rb
obino: we do, but I think is still valuable to leave it for community member wanting to work on a different database. Moreover changing it is a bit more complicated at this point.

AppController/djinn.rb
obino: I see what you mean. I was thinking about cloud infrastructures: I was tempted to use public, but we do support also private cloud as well, making it more complicated. I think that here is not confusing, since it's not our configuration, but relates to the infrasrtucture this deployment is using. Managed infrastrucuture seems odd to me. Perhaps 'used only on supported cloud platforms'?

AppController/djinn.rb
obino: oh ... sorry I must have missed the memo. I was following the ruby guidelines. I can re-add them though.

AppController/djinn.rb
obino: we could, but we have it anyway just checking at the length of the array. Isn't it better to avoid another variable that we need to keep synchronized?

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2343/

AppController/djinn.rb
menivaitsi: Well, "supported cloud platforms" is misleading because it insinuates that you cannot run in SoftLayer, for example, while you can.

You can leave it as is.

AppController/djinn.rb
menivaitsi: @obino I just meant rewrite these three lines like so:
```
current_appengines = @app_info_map[app_name]['appengine']
min_appengines = Integer(@options['appengine'])
if (current_appengines.nil? || current_appengines.length < min_appengines)
  ...
```

AppController/djinn.rb
cdonati: We are very inconsistent about this. When I first joined, the policy was to include the parentheses. However, the only style guides I was able to find seem to lean towards omitting them. I don't have a strong opinion, so I'm fine with omitting them unless someone disagrees.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2350/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2351/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2352/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2354/

PR #2254, opened 2016-12-13
Ensure monit stays up

obino:
Since we rely on Monit for most of our AppServers and Services, we need
to make the best effort to keep it running in spite of OOM and similar
conditions.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2386/

PR #2259, opened 2016-12-15
If rebalance fails, try next largest diff

cdonati:
Nodetool is sometimes unable to collect a sample of keys on the node that we want to move load from. This branch allows the rebalance script to try moving data from another machine instead of doing nothing.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2401/

PR #2261, opened 2016-12-16
Fix git status check for Java AppServer

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2405/

PR #2263, opened 2016-12-21
Check status code when fetching public IP

cdonati:
This was previously using the 404 template returned by Google as the IP address when trying to fetch AWS's metadata URL.


obino:
There is a constant in HelperFunctions called IP_OR_FQDN that may be used to validate the String as an IP or name. We could add them to the returned values.

PR #2268, opened 2017-01-05
Include options and facility arguments for syslog

cdonati:
The interpreter will still handle the function call without the additional arguments, but since the official documentation specifies 3 required arguments, some tools complain about it.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2456/

PR #2272, opened 2017-01-06
Use load balancers

obino:
This PR puts the file we write for the various services into a single method (AC related). Quite a few services uses to write the address of their members (say the taskqueues nodes) into a file for the stubs to find. This PR avoid the blindly re-write of all of those files each duty cycle, but write them only upon changes.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -4157,7 +4157,6 @@ class Djinn
     my_public = my_node.public_ip
     my_private = my_node.private_ip
 
-
     # Populate the appropriate list.
     @nodes.each { |node|
       all_ips << node.private_ip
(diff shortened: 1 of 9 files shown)


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2463/

AppController/djinn.rb
```diff
@@ -4151,148 +4139,97 @@ def rsync_files(dest_node)
...
+    taskqueue_ips = []
 
     my_public = my_node.public_ip
-    HelperFunctions.write_file("#{APPSCALE_CONFIG_DIR}/my_public_ip", "#{my_public}\n")
-
     my_private = my_node.private_ip
-    HelperFunctions.write_file("#{APPSCALE_CONFIG_DIR}/my_private_ip", "#{my_private}\n")
-
-    head_node_ip = get_shadow.public_ip
-    HelperFunctions.write_file("#{APPSCALE_CONFIG_DIR}/head_node_ip", "#{head_node_ip}\n")
```
menivaitsi: This is used by the dashboard. Without it appscale up hangs.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2530/

AppController/djinn.rb
```diff
@@ -4155,148 +4143,97 @@ def rsync_files(dest_node)
...
+    master_content = master_ips.join("\n") + "\n"
+    search_content = slave_ips.join("\n") + "\n"
+    slaves_content = slave_ips.join("\n") + "\n"
+
+    new_content = all_ips_content + login_content + load_balancer_content +
+      master_content + memcache_content + my_public + my_private +
+      num_of_nodes + taskqueue_content + search_content + slaves_content
+
+    # If nothing changed since last time we wrote locations file(s), skip it.
+    if new_content != @locations_content
```
whoarethebritons: This is still going to be true because the locations_content will have the shuffled taskqueue_ips... If you sorted both taskqueue lists and added it as an "and" to the condition that would test for that. I'm not really concerned about it too much its just not saving us anything right now.

obino: Good point. I addressed it with saving the content locally before we shuffle it, so that we check the 'freshness' of the cache on the un-shuffled TQ list to make ir reproducible.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2546/

PR #2280, opened 2017-01-19
Handle Java chunk size parameter

cdonati:
The Java datastore API allows the app to specify a chunk size when fetching results for a query. Previously, we were ignoring this parameter and always fetching all the results, which is not efficient when using an iterator.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2510/

PR #2283, opened 2017-01-20
Dashboard uses different mins

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2533/ 

One of the  hawkeye runs, failed because one of the instances got 'lost' by livirt on n30.

PR #2288, opened 2017-01-24
Put /usr/local/bin first on the PATH

cdonati:
This ensures that the controller uses the updated pip binary instead of the one that was installed by the package manager.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2575/

PR #2291, opened 2017-01-25
Catch HTTP errors when executing push tasks

cdonati:
This increases the likelihood that failed push tasks will be retried.

### DIFF
--- a/AppTaskQueue/appscale/taskqueue/templates/header.py
+++ b/AppTaskQueue/appscale/taskqueue/templates/header.py
@@ -8,6 +8,7 @@
 """
 import datetime
 import httplib
+import logging
 import os
 import sys
 import yaml
@@ -54,6 +55,7 @@ celery = Celery(module_name, broker=rabbitmq.get_connection_string(),
 celery.config_from_object('CELERY_CONFIGURATION')
 
 logger = get_task_logger(__name__)
+logger.setLevel(logging.INFO)
 
 master_db_ip = appscale_info.get_db_master_ip()
 connection_str = master_db_ip + ":" + str(constants.DB_SERVER_PORT)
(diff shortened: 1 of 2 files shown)


AppTaskQueue/appscale/taskqueue/templates/header.py
```diff
@@ -52,6 +55,7 @@ def setup_environment():
 celery.config_from_object('CELERY_CONFIGURATION')
 
 logger = get_task_logger(__name__)
+logger.setLevel(logging.INFO)
```
menivaitsi: Is there a way to pass in the log level?

cdonati: We could set an environment variable in the monit config when we start celery.

AppTaskQueue/appscale/taskqueue/templates/task.py
```diff
@@ -90,42 +96,51 @@ def get_wait_time(retries, args):
...
       logger.info("Task %s asked us to redirect to %s, so retrying there." % (
         args['task_name'], redirect_url))
       url = urlparse(redirect_url)
       if redirects_left == 0:
-        raise QUEUE_NAME.retry(countdown=get_wait_time(retries, args))
+        raise QUEUE_NAME.retry(countdown=wait_time)
       redirects_left -= 1
     else:
       # Fail
       # TODO: Update the database with the failed status
```
obino: If I understand correctly, we don't keep track of failed status on purpose (since we do a retry). Is this TODO here for a reason then?

menivaitsi: I believe it's leftover from a different approach. It can be removed.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2612/

PR #2292, opened 2017-01-26
Implement inequality check for Queue objects

cdonati:
This also removes a redundant call to monit.


obino:
Tested with removing monit config and restating, and all worked as expected. @menivaitsi I was not able to reproduce what we have seen yesterday.

PR #2294, opened 2017-01-28
Use non-blocking seed source for Java AppServer

cdonati:
When the entropy pool is depleted, the Java AppServer hangs before it can start serving. Using /dev/urandom will prevent this from happening.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2610/

PR #2299, opened 2017-01-31
Partial implementation of ModuleService in AppServer_Java

ochre:


obino:
@ochre as discussed, we won't be able to use this partial implementation, but we will wait for the modules full implementation. Thanks.

PR #2300, opened 2017-01-31
Added support for Host header in Task

ochre:


cdonati:
I'm curious about the use case for this change. Is this so you can enqueue a task on a different deployment? If so, I would recommend using urlfetch.

This change could lead to unexpected behavior if someone wants to set the host header for a task and they don't want to change the url. For example, they may have nginx on the load balancer set up to listen for a custom host, but they may not have dns configured for that host on the machine running the celery worker. In that case, the task would fail.

obino:
Thanks for submitting the PR: since we didn't get an answer on the use case and the age of the PR (it is getting conflicts) I am going to close it for now. Please contact us if you want to work on this.

PR #2307, opened 2017-02-02
Fix identifying search ip as an array instead of string

khaledkbadr:
Search ip was identified as a string that caused that error
```
/root/appscale/AppController/djinn.rb:4216:in `write_locations': undefined method `join' for "192.168.103.250":String (NoMethodError)
        from /root/appscale/AppController/djinn.rb:2949:in `update_firewall'
        from /root/appscale/AppController/djinn.rb:2071:in `job_start'
        from /root/appscale/AppController/djinnServer.rb:168:in `<main>'
```


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2671/

PR #2309, opened 2017-02-02
Ensure groups have a defined namespace

cdonati:
Clients do not always include the namespace as part of entity keys. This change ensures that groups with undefined namespaces are stored with an empty namespace string so that they will not be treated as a different group.


cdonati:
Build: https://ocd.appscale.com:8080/job/Daily%20Build/2658/
https://ocd.appscale.com:8080/job/Hawkeye/9847/
https://ocd.appscale.com:8080/job/Tools-test/1404/
https://ocd.appscale.com:8080/job/Blobstore/1694/

PR #2311, opened 2017-02-03
Consider user props with the same email as equal

cdonati:
When validating indexes generated from user properties, the datastore only has the email address.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2778/

PR #2315, opened 2017-02-07
Limit LogServer log retention

menivaitsi:


LogService/twisted/plugins/appscale_logserver_plugin.py
```diff
@@ -12,7 +12,7 @@
 class Options(usage.Options):
   optParameters = [["port", "p", 7422, "The port number to listen on."],
                    ["path", "a", "/opt/appscale/logserver", "Path where logs are stored."],
-                   ["size", "s", 25, "Size in GiB of retention of logs."],
+                   ["size", "s", 2, "Size in GiB of retention of logs."],
```
whoarethebritons: Won't this make it so we can't query back as far as we'd like to?

menivaitsi: That is correct. Currently GAE provides up to 2 GBs of logs.
A user can customize this value based on their from let's say the AppScalefile.

The root filesystem on any supported public cloud given will not be set up to hold that much data unless there are external volumes attached. This is a liability ATM.

obino: And is this per application or total?

menivaitsi: It is not straightforward to differentiate per app at that level.

PR #2316, opened 2017-02-07
Don't retry delete entry from zookeeper

obino:
Since the headnode may have already cleared the app-hosters list in
zookeeper, there is no need to retry deleting this node from the
app-hosters.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2703/

PR #2317, opened 2017-02-08
Don't start an extra AppServer to reach min

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2710/

PR #2330, opened 2017-02-14
Use transaction ID to determine write time

cdonati:
Though the group lock prevents parallel writes, local clock skews can cause linearizability problems. Since we are not using LWTs, the write with the latest timestamp is the one that gets read from Cassandra. If the coordinator node's clock is far enough behind on a later transaction, the write can get buried under an earlier transaction. This change uses the transaction ID as the basis for the write time to prevent this possibility.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2760/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2798/

PR #2331, opened 2017-02-14
Decrease gc_grace_seconds on ephemeral data

cdonati:
The downside of decreasing this value is that it's possible for data to be (unintentionally) undeleted if a node goes offline and the tombstones were not replicated.

For ephemeral tables like transactions and pull_queue_leases (used for stats), this downside is not relevant since all data is inserted with a TTL.

For pull_queue_tasks_index, undeleted index entries will not affect correctness, and it will barely affect performance.

The major benefit is reduced disk space usage (for the transactions table in particular) and increased performance for pull queue index lookups since tombstones will be deleted earlier.

### DIFF
--- a/AppDB/appscale/datastore/cassandra_env/schema.py
+++ b/AppDB/appscale/datastore/cassandra_env/schema.py
@@ -143,7 +143,7 @@ def create_transactions_table(session):
       entity blob,
       task blob,
       PRIMARY KEY (txid_hash, operation, namespace, path)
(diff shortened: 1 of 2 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2762/

cdonati:
Migration: https://ocd.appscale.com:8080/job/Migration/1032/

PR #2334, opened 2017-02-18
Logquery reverse option

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2790/


whoarethebritons:
@cdonati pointed out we should test whether this would affect the appengine side, please do not merge until that is determined.

PR #2344, opened 2017-02-25
Lease multiple tasks at the same time

cdonati:


AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -773,84 +781,103 @@ def _get_earliest_tag(self):
       return None
     return tag
 
-  def _lease_task(self, index, new_eta):
-    """ Acquires a lease on task in the queue.
+  def _increment_count_async(self, task):
+    """ Update retry count for for a task.
```
menivaitsi: Remove double "for".

PR #2345, opened 2017-02-25
Prevent multiple large batches for a transaction

cdonati:
Each transaction should be atomic.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2817/

PR #2346, opened 2017-02-25
Avoid parsing string representation of LogOffset

cdonati:
Instead, reference request_id property directly.

### DIFF
--- a/AppServer/google/appengine/api/logservice/logservice_stub.py
+++ b/AppServer/google/appengine/api/logservice/logservice_stub.py
@@ -166,7 +166,7 @@ class LogServiceStub(apiproxy_stub.APIProxyStub):
         fh.close()
       self._release_logserver_connection(key, log_server)
     except socket.error, e:
-      cleanup_logserver_connection(log_server)
+      _cleanup_logserver_connection(log_server)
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2848/

AppServer/google/appengine/api/logservice/logservice_stub.py
```diff
@@ -309,8 +308,8 @@ def _Dynamic_Read(self, request, response, request_id):
         _fill_request_log(requestLog, log, request.include_app_logs())
         result_count += 1
 
-      if result_count == count:
-        response.mutable_offset().set_request_id(requestLog.offset)
+        if result_count == count:
+          response.mutable_offset().set_request_id(requestLog.offset)
```
obino: Shouldn't we return if reach the requested count?

cdonati: The count is included in the query, so it shouldn't exceed the count. This is just to make sure a cursor is included in the response if we satisfy the count (indicating that there are probably more results available if given the new offset).

PR #2348, opened 2017-02-27
Re-added the add-instances capability

obino:
This PR reconciles the changes made into the autoscale, with the manual
scaling done with appscale-add-instances.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2820/

AppController/djinn.rb
```diff
@@ -5418,25 +5317,93 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+        Djinn.log_info("We need #{vms_to_spawn} more VMs.")
+
+        if Time.now.to_i - @last_scaling_time < (SCALEUP_THRESHOLD *
+                SCALE_TIME_MULTIPLIER * DUTY_CYCLE)
+          Djinn.log_info("Not scaling up right now, as we recently scaled " +
+            "up or down.")
+          return
+        end
+
+        allowed_vms = Integer(@options['max_images']) - @nodes.length
```
cdonati: Since `@nodes` is not changed in this same sync block, it seems like this still has the potential for exceeding `max_images`. Is there some global state somewhere that keeps track of machines that are still starting?

AppController/test/tc_djinn.rb
```diff
@@ -469,368 +467,6 @@ def test_start_roles_on_nodes_bad_input
...
-
-    djinn = Djinn.new()
-    djinn.nodes = [original_node]
-    djinn.my_index = 0
-    djinn.options = options
-    flexmock(djinn).should_receive(:add_nodes).and_return()
-    actual = djinn.start_new_roles_on_nodes_in_cloud(ips_to_roles)
-    assert_equal(true, actual.include?(node1_info))
-    assert_equal(true, actual.include?(node2_info))
-  end
```
cdonati: If these unit tests are no longer useful, should they be replaced with one that is?

AppController/djinn.rb
```diff
@@ -2506,208 +2496,116 @@ def is_app_running(appname, secret)
...
-      nodes_info << {
-        "public_ip" => ip,
-        "private_ip" => ip,
-        "jobs" => roles,
-        "disk" => nil
+      ips.each { |ip_or_node|
+        begin
+          # Convert (or check) if we have an IP address or we have a node
+          # we need to start, then we add this role to the node.
+          ip = HelperFunctions.convert_fqdn_to_ip(ip_or_node)
```
obino: Not really. The idea is that the rescue block is a name that cannot be converted (so we hope that node-x cannot be converted), which means that is a node we need to add by scratch. If we were to force IP address or node-x (instead of IP/hostname or node-x) we could get by without this conversion.

AppController/djinn.rb
```diff
@@ -5418,25 +5317,93 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+    if needed_appservers > 0
+      Djinn.log_debug("Need to start VMs for #{needed_appservers} more AppServers.")
+    end
+
+    # Here we cound the number of machines we need to spawn, and the roles
+    # we need.
+    vms_to_spawn = 0
+    roles_needed = {}
+    if needed_appservers > 0
+      Integer(needed_appservers/3).downto(0) {
```
obino: Yes, eventually we want to combine that with the memory used by the specific app and the memory available in each instance type.

AppController/djinn.rb
```diff
@@ -2506,208 +2496,116 @@ def is_app_running(appname, secret)
...
-        "private_ip" => ip,
-        "jobs" => roles,
-        "disk" => nil
+      ips.each { |ip_or_node|
+        begin
+          # Convert (or check) if we have an IP address or we have a node
+          # we need to start, then we add this role to the node.
+          ip = HelperFunctions.convert_fqdn_to_ip(ip_or_node)
+          found = false
+          node_roles.each { |node|
```
obino: No it doesn't need to be ordered. Originally they were been used in a different way: what we can pass directly to the spawn is just a list of jobs, while whatever we got with an IP, we need to compare it with @nodes so I was trying to keep the structure similar. Not sure if it still makes sense.

AppController/djinn.rb
```diff
@@ -2506,208 +2496,116 @@ def is_app_running(appname, secret)
...
+          }
+          unless found
+            node_roles << {
+              "public_ip" => ip,
+              "private_ip" => ip,
+              "jobs" => role,
+              "disk" => nil
+            }
+          end
+        rescue AppScaleException
```
obino: Are you suggesting to get the unless found out of the rescue block? Would that be more readable?

AppController/djinn.rb
```diff
@@ -2506,208 +2496,116 @@ def is_app_running(appname, secret)
...
-
-      # Initialize them and integrate them with our nodes.
-      Djinn.log_debug("info about new nodes is " +
-        "[#{new_nodes_info.join(', ')}]")
-      add_nodes(new_nodes_info)
     end
+    Djinn.log_debug("We spawned VMs for these roles #{new_nodes_info}.")
+    Djinn.log_debug("We wil use the following nodes #{node_roles}.")
+
+    # If we have an already running nodes with the same IP, we changes
```
obino: lol thanks, 2 typo per line ...

AppController/djinn.rb
obino: good point: we should have @nodes changes under its lock. I will add it. In general in this part of the code, we shouldn't run over the max_images allowed.

AppController/test/tc_djinn.rb
obino: yes :)

AppController/djinn.rb
cdonati: One idea that might make this more readable would be to split `ips` to something like `nodes_without_ips` and `ips` before this loop. It's a little confusing that `ips` contains some things that aren't IP addresses.

AppController/djinn.rb
```diff
@@ -5418,25 +5364,95 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+
+        if Time.now.to_i - @last_scaling_time < (SCALEUP_THRESHOLD *
+                SCALE_TIME_MULTIPLIER * DUTY_CYCLE)
+          Djinn.log_info("Not scaling up right now, as we recently scaled " +
+            "up or down.")
+          return
+        end
+
+        @state_change_lock.synchronize {
+          allowed_vms = Integer(@options['max_images']) - @nodes.length
```
cdonati: As far as I can tell, adding locks around other places that deal with `@nodes` does not address the problem. My concern was that @nodes was not being changed in this _current_ lock. Maybe start_roles_on_nodes changes @nodes, I'm not sure. I think the change should be more explicit if it does.

The last time I tested autoscaling, it triggered another scaling event before the new nodes were added to @nodes, so it started more nodes even though the total resulting nodes exceeded `max_images`. If changing @nodes is not feasible in this block, it seems like allowed_vms should take the nodes that are still starting into account.

AppController/djinn.rb
```diff
@@ -2506,208 +2496,119 @@ def is_app_running(appname, secret)
...
-    end
-
-    Djinn.log_info("Received a request to acquire nodes with roles " +
-      "#{nodes_needed.join(', ')}, with instance type #{instance_type}.")
-
-    vms_to_use = []
-
-    # We look for 'open' nodes first.
+    # Let's count the open nodes we can use before having to create new
+    # VMS.
```
menivaitsi: VMS -> VMs

AppController/djinn.rb
```diff
@@ -2506,208 +2496,119 @@ def is_app_running(appname, secret)
...
-
-      # Initialize them and integrate them with our nodes.
-      Djinn.log_debug("info about new nodes is " +
-        "[#{new_nodes_info.join(', ')}]")
-      add_nodes(new_nodes_info)
     end
+    Djinn.log_debug("We spawned VMs for these roles #{new_nodes_info}.")
+    Djinn.log_debug("We will use the following nodes #{node_roles}.")
+
+    # If we have an already running node with the same IP, we change it's
```
menivaitsi: it's -> its

AppController/djinn.rb
```diff
@@ -3153,8 +3071,11 @@ def backup_appcontroller_state()
     APPS_LOCK.synchronize {
       local_state = {'@@secret' => @@secret }
       DEPLOYMENT_STATE.each { |var|
+        v = nil
```
menivaitsi: I suggest we abstain from single letter var names. It's harder to read and maintain.

AppController/djinn.rb
```diff
@@ -3613,30 +3540,34 @@ def find_me_in_locations()
...
-      end
-      if node.public_ip == public_ip
-        HelperFunctions.log_and_crash("Found my public ip (#{public_ip}) " +
-            "in @nodes but my private ip is not matching! @nodes=#{@nodes}.")
-      end
+    @state_change_lock.synchronize {
+      @nodes.each { |node|
+        if node.private_ip == public_ip
+          HelperFunctions.log_and_crash("Found my public ip (#{public_ip}) " +
+              "but not my private ip in @nodes. Please correct it. @nodes=#{@nodes}")
```
menivaitsi: Please use 2-space indentation here.

AppController/djinn.rb
```diff
@@ -3613,30 +3540,34 @@ def find_me_in_locations()
...
+        end
+        if node.public_ip == public_ip
+          HelperFunctions.log_and_crash("Found my public ip (#{public_ip}) " +
+              "in @nodes but my private ip is not matching! @nodes=#{@nodes}.")
```
menivaitsi: Same here.

AppController/djinn.rb
```diff
@@ -3679,14 +3610,19 @@ def start_stop_api_services
...
+        @nodes.each { |node|
+          db_master = node.private_ip if node.jobs.include?('db_master')
+        }
       }
       setup_db_config_files(db_master)
 
       threads << Thread.new {
         Djinn.log_info("Starting database services.")
-        db_nodes = @nodes.count{|node| node.is_db_master? or node.is_db_slave?}
+        db_nodes = nil
```
menivaitsi: Is this needed? Is there an action if it's not set?

AppController/djinn.rb
```diff
@@ -5418,25 +5364,95 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+  # services.
+  def scale_deployment
+    needed_appservers = scale_appservers
+    if needed_appservers > 0
+      Djinn.log_debug("Need to start VMs for #{needed_appservers} more AppServers.")
+    end
+
+    # Here we cound the number of machines we need to spawn, and the roles
+    # we need.
+    vms_to_spawn = 0
```
menivaitsi: Can we rename this to something like: _vm_id_? It's a bit confusing as is.

AppController/djinn.rb
```diff
@@ -5418,25 +5364,95 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+      Djinn.log_debug("Another thread is already working with the InfrastructureManager.")
+      return
+    end
+
+    Thread.new {
+      SCALE_LOCK.synchronize {
+        Djinn.log_info("We need #{vms_to_spawn} more VMs.")
+
+        if Time.now.to_i - @last_scaling_time < (SCALEUP_THRESHOLD *
+                SCALE_TIME_MULTIPLIER * DUTY_CYCLE)
```
menivaitsi: Please use 4-space indentation for continuation lines.

AppController/lib/infrastructure_manager_client.rb
```diff
@@ -175,7 +175,20 @@ def terminate_instances(options, instance_ids)
...
  
   
-  def spawn_vms(num_vms, options, job, disks)
+  # Create new VMs.
+  #
+  # Args:
+  #   num_vms: the number of VMs to create.
+  #   options: a hash containing information needed by the agent
+  #     (credentials etc ...).
+  # jobs: an Array containing the roles for each VM to be created.
```
menivaitsi: Indent these two args.

AppController/lib/infrastructure_manager_client.rb
```diff
@@ -175,7 +175,20 @@ def terminate_instances(options, instance_ids)
...
+  # disks: an Array specifying the disks to be associated with the VMs (if
+  #   any, it can be nil).
+  #
+  # Returns
+  #   an Array containing the nodes information, suitable to be converted
```
menivaitsi: an -> An

AppController/djinn.rb
obino: I'm afraid about the scope of db_nodes, that now could be in an inner loop ... I will have to test more with ruby.

AppController/djinn.rb
obino: Hmm ... I'm not quite sure: those are the numbers of new VMs we need to create. vm_id doesn't seem to specify anything useful here: we don't have any ID, and actually we don' t have the VMs yet. Perhaps vm_to_create? or num_vm_to_spawn/create? Or vms_desired/required?

AppController/djinn.rb
obino: For the ips part, isn't the new code a bit more clear? We split the usage right away into new_node (to create) vs the ones that we need to check if they exists.

obino: Got it: the nodes are added by add_node, at the end of start_roles_on_nodes. Since we have now a lock around it (few line before with SCALE_LOCK) there shouldn't be a way to have autoscaling tripping over itself.

AppController/djinn.rb
obino: Ok, it is required, otherwise the db_nodes variable won't be accessible outside the synchronized scope.

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/3096/

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3097/

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3098/

AppController/djinn.rb
```diff
@@ -5357,25 +5301,95 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+          Djinn.log_info("Not scaling up right now, as we recently scaled " +
+            "up or down.")
+          return
+        end
+
+        @state_change_lock.synchronize {
+          allowed_vms = Integer(@options['max_images']) - @nodes.length
+          if allowed_vms < vms_to_spawn
+            Djinn.log_warn("Requested #{vms_to_spawn} VMs, but we can only" +
+              " start #{allowed_vms}, so not spawning more nodes.")
```
tmarballi: So I was testing autoscaling with a 2 node deployment and I set max images to 3 and appengine to 15. It needed to spawn 5 VMs to accommodate 15 appservers, that makes vms_to_spawn = 5 and allowed_vms = 1 (according to this calculation). Shouldn't it still spawn that one VM which is still in the max_images limit? It would not do it in this condition because allowed_vms is less than the vms_to_spawn.

AppController/djinn.rb
```diff
@@ -5357,25 +5301,95 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+  # Adds or removes AppServers and/or nodes to the deployment, depending
+  # on the statistics of the application and the loads of the various
+  # services.
+  def scale_deployment
+    needed_appservers = scale_appservers
+    if needed_appservers > 0
+      Djinn.log_debug("Need to start VMs for #{needed_appservers} more AppServers.")
+    end
+
+    # Here we cound the number of machines we need to spawn, and the roles
```
tmarballi: Change cound => count

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3115/

tmarballi:
https://ocd.appscale.com:8080/job/Tools-test/

AppController/lib/infrastructure_manager_client.rb
```diff
@@ -176,7 +176,20 @@ def terminate_instances(options, instance_ids)
...
   
-  def spawn_vms(num_vms, options, job, disks)
+  # Create new VMs.
+  #
+  # Args:
+  #   num_vms: the number of VMs to create.
+  #   options: a hash containing information needed by the agent
+  #     (credentials etc ...).
+  #   jobs: an Array containing the roles for each VM to be created.
+  #     disks: an Array specifying the disks to be associated with the VMs
```
jeanleonov: Two white spaces before `disk:` can be removed

AppController/djinn.rb
```diff
@@ -2516,200 +2503,112 @@ def start_roles_on_nodes(ips_hash, secret)
...
+    # We spawn new nodes if we need to (and can do so) here.
+    new_nodes_info = []
+    if new_nodes_roles.length > open_nodes
+      num_of_vms = new_nodes_roles.length - open_nodes
       unless is_cloud?
-        Djinn.log_error("Still need #{vms_to_spawn} more nodes, but we " +
+        Djinn.log_warn("Still need #{num_of_vms} more nodes, but we " +
         "aren't in a cloud environment, so we can't acquire more nodes - " +
         "failing the caller's request.")
         return NOT_ENOUGH_OPEN_NODES
```
jeanleonov: Wouldn't it better to raise an exception here?
It's not merge stopper.. but if you address other comments somewhere here...

obino: I'm not sure the exception is carried well across SOAP, and also that the tools handles it. I would say that if we go for it, we may want to wait for the Admin API and have this case handled there.

AppController/djinn.rb
```diff
@@ -4287,24 +4227,26 @@ def write_locations
...
-      load_balancer_ips << node.private_ip if node.is_load_balancer?
-      master_ips << node.private_ip if node.is_db_master?
-      memcache_ips << node.private_ip if node.is_memcache?
-      search_ips << node.private_ip if node.is_search?
-      slave_ips << node.private_ip if node.is_db_slave?
-      taskqueue_ips << node.private_ip if node.is_taskqueue_master? ||
-        node.is_taskqueue_slave?
+    num_of_nodes = 0
+    @state_change_lock.synchronize {
+      num_of_nodes = @nodes.length.to_s
```
whoarethebritons: Why do we make num_of_nodes a string?

AppController/djinn.rb
```diff
@@ -5358,25 +5302,92 @@ def find_lowest_free_port(starting_port, ending_port=0, appid="")
...
+      Djinn.log_debug("Need to start VMs for #{needed_appservers} more AppServers.")
+    end
+
+    # Here we count the number of machines we need to spawn, and the roles
+    # we need.
+    vms_to_spawn = 0
+    roles_needed = {}
+    vm_scaleup_capacity = Integer(@options['max_images']) - @nodes.length
+    if needed_appservers > 0
+      Integer(needed_appservers/3).downto(0) {
```
whoarethebritons: Magic number, should this be a constant?

AppController/djinn.rb
```diff
@@ -5648,37 +5658,48 @@ def try_to_scale_up(app_name, delta_appservers)
...
+
+    # Let's pick the latest appengine node hosting the application and
+    # remove the AppServer there, so we can try to reclaim it once it's
+    # unloaded.
+    get_all_appengine_nodes.reverse { |node_ip|
+      @app_info_map[app_name]['appengine'].each { |location|
+        host, port = location.split(":")
+        if host == node_ip
+          @app_info_map[app_name]['appengine'].delete(location)
+          @last_decision[app_name] = Time.now.to_i
```
whoarethebritons: Does this need to be in the state lock?

AppController/djinn.rb
obino: because it will be used a string, so I thought it was clear to make it a string here.

AppController/djinn.rb
obino: this is just temporary: it will be removed once we move to capacity triggers. We could a TODO I guess?

AppController/djinn.rb
obino: to be safe, I tried to add a lock around any changes of @app_info_map since it's a global structure and it's used in threads extensively.

AppController/djinn.rb
whoarethebritons: I see. I saw the +'s below and assumed it was a number

AppController/djinn.rb
whoarethebritons: @obino its not in the state lock is what I meant

obino: you are rigth, scale-deployment ended up outside the syncrhonize.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3117/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3119/

PR #2350, opened 2017-02-28
Don't log info-level message after successful call

cdonati:
When not in verbose mode, this message dominates the controller log.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2851/

PR #2352, opened 2017-03-01
Use argparse for the datastore options

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2859/

jeanleonov:
@cdonati , I see - you made parsing of command line arguments quite better.
But wouldn't you like to try [docopt](https://github.com/docopt/docopt)?
It's modern lib for parsing arguments which uses POSIX help message for command as configuration.
I applied it in [hawkeye](https://github.com/AppScale/hawkeye/pull/109)

cdonati:
@jeanleonov I like that argparse is in the standard library. I'm certainly not opposed to using docopt. It looks like a great library. However, I haven't used it before, so I would need to learn the syntax.

PR #2355, opened 2017-03-03
Create apps dir in moutn_persistent_storage

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2846/

obino:
@cdonati yeah I see what you mean. I didn't like to have a single mkdir afterward though, and since I wasn't sure in which function to add it, that was the best candidate. Do you think a single mkdir afterward is better?

cdonati:
I think this is fine.

PR #2356, opened 2017-03-03
Don't log sensitive data

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2847/

obino:
@jeanleonov well the information written there were already in the logs at the start of the process when we call set_paramters, which effectively makes them redundant.  Same for nodes: we print them when we change them, so that was somewhat redundant. If you still think we should have the controller state been logged, I will ping you and ask help to get the patched to_s method :)

AppController/djinn.rb
```diff
@@ -1069,8 +1069,7 @@ def set_parameters(layout, options, secret)
       @options['ec2_url'] = @options['EC2_URL']
     end
 
-    Djinn.log_debug("set_parameters: set @options to #{@options}.")
-    Djinn.log_debug("set_parameters: set @nodes to #{@nodes}.")
+    Djinn.log_run("mkdir -p #{PERSISTENT_MOUNT_POINT}/apps")
```
cdonati: You probably need to delete this line.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2880/

PR #2357, opened 2017-03-03
Change message and severity of log message

obino:
Do not warn if we cannot find our old IP address files, since this is
the normal operation during the first 'up'.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2849/

PR #2359, opened 2017-03-05
Indicate that the index is clean after a scrub

cdonati:
This will prevent the groomer from cleaning indexes unnecessarily.

### DIFF
--- a/AppDB/appscale/datastore/groomer.py
+++ b/AppDB/appscale/datastore/groomer.py
@@ -12,6 +12,7 @@ import dbconstants
 import utils
 
 from appscale.taskqueue.distributed_tq import TaskName
+from .cassandra_env import cassandra_interface
 from .datastore_distributed import DatastoreDistributed
 from .unpackaged import APPSCALE_LIB_DIR
 from .unpackaged import APPSCALE_PYTHON_APPSERVER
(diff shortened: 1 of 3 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2854/

AppDB/appscale/datastore/groomer.py
```diff
@@ -556,6 +557,15 @@ def clean_up_indexes(self, direction):
...
     end_key = dbconstants.TERMINATING_STRING
 
+    # Indicate that an index scrub has started after the journal was removed.
+    if direction == datastore_pb.Query_Order.ASCENDING and not start_key:
+      index_state = self.db_access.get_metadata(
+        cassandra_interface.INDEX_STATE_KEY)
+      if index_state is None:
+        self.db_access.set_metadata(
+          cassandra_interface.INDEX_STATE_KEY,
+          cassandra_interface.IndexStates.SCRUB_IN_PROGRESS)
```
obino: Is there a chance if a groomer dies (say out of memory) to have an index that will stay in this state forever?

cdonati: There's no behavioral difference for the deployment between an unset key and `SCRUB_IN_PROGRESS`, so staying in that state for awhile shouldn't be a problem.

Once the groomer run again and gets to the end of scrubbing the kind index, this index will be marked as clean.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2905/
The first 2 hawkeye failures happened because the app was stopped before it could be marked as enabled. The last hawkeye failure was due to a euca 503.

PR #2360, opened 2017-03-05
Remove API proxy lock for datastore and taskqueue

cdonati:
The state kept the API proxy for the datastore service includes an index cache, query cursors, and transactional tasks. The index cache is built when the AppServer starts and never altered afterwards. All of the operations on the query cursors and transactional tasks are atomic.

It's possible for another client to mess with a request's cursor and tasks, but the API lock does not help with that.

There is no state kept in the taskqueue service stub.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2856/

PR #2364, opened 2017-03-07
Reduce taskqueue multiplier

cdonati:
This gives single-node deployments more memory to work with.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2861/

PR #2367, opened 2017-03-07
Cluster stats refactoring

jeanleonov:
AppController and InfrustructureManager seems to be done.
Dashboard is in progress yet.

It makes sense to do a demo first. But we were discussing these changes, so you're welcome to leave some comments before the demo.

Is not mergable yet.


AppController/djinn.rb
```diff
@@ -505,7 +571,7 @@ def initialize()
     @@logs_buffer = []
 
     @@log = Logger.new(STDOUT)
-    @@log.level = Logger::INFO
+    @@log.level = Logger::DEBUG
```
obino: The default should be INFO.

AppController/djinn.rb
```diff
@@ -1260,22 +1250,22 @@ def get_app_upload_status(reservation_id, secret)
   #   secret: A string with the shared key for authentication.
   # Returns:
   #   A JSON string with the statistics of the nodes.
-  def get_stats_json(secret)
+  def get_cluster_stats_json(secret)
```
obino: I personally prefer the name get_cluster_stats (without _json) since this is the only one to collect stats (that is, there is no the non-json form).

AppController/djinn.rb
```diff
@@ -4915,7 +4861,9 @@ def prep_app_dashboard()
           'min_appengines' => 3
         }
       end
-      @app_names << AppDashboard::APP_NAME
+      if @app_names.include? AppDashboard::APP_NAME
```
obino: What is the purpose of this check?

AppController/djinn.rb
```diff
@@ -2226,7 +2172,7 @@ def job_start(secret)
         if my_node.is_shadow? && @options['autoscale'].downcase != "true"
           Djinn.log_info("--- This deployment has autoscale disabled.")
         end
-        stats = JSON.parse(get_all_stats(secret))
+        stats = get_node_stats(secret)
```
obino: I am having trouble to locate this function get_node_stats: is that needed?

AppController/djinn.rb
jeanleonov: This specific function does have only one form and it is JSON. But other functions of AppContoller not always uses JSON. So I'd say it's sensible to leave `_json` suffix.

jeanleonov:
[Build #2906](https://ocd.appscale.com:8080/job/Daily%20Build/2906/) uses this branch of appscale repo and branch with updated status of tools repo.

AppController/djinn.rb
```diff
@@ -2644,7 +2594,7 @@ def start_new_roles_on_nodes(nodes_needed, instance_type, secret)
     return BAD_SECRET_MSG unless valid_secret?(secret)
 
     # TODO: we should validate the roles type, to ensure they are valid.
-    if nodes_needed.class != Array || instance_type.class != String
+    if nodes_needed.class != Array || instance_type.class != String && is_cloud
```
cdonati: I would recommend adding parentheses for clarity. Though there is no ambiguity for the interpreter, a reader without a firm grasp of operator precedence in Ruby might wonder if the above represents
`if (nodes_needed.class != Array || instance_type.class != String) && is_cloud`
or
`if nodes_needed.class != Array || (instance_type.class != String && is_cloud)`

AppController/djinn.rb
```diff
@@ -5927,47 +5807,27 @@ def remove_appserver_process(app_id, port)
...
     Djinn.log_debug("Sending a log with request rate #{app_id}, timestamp " +
-      "#{timestamp}, request rate #{request_rate}")
+      "#{@last_sampling_time[app_id]}, request rate " +
+      "#{@average_req_rate[app_id]}")
     encoded_request_info = JSON.dump({
-      'timestamp' => timestamp,
-      'request_rate' => request_rate
+      'timestamp' => @last_sampling_time[app_id],
+      'avg_request_rate' => @average_req_rate[app_id],
+      'num_of_requests' => @total_req_rate[app_id]
```
cdonati: I realize this was named this way before your branch, but we may want to change it at some point since "request rate" sounds like it involves two units (eg. req/sec), but "num_of_requests" sounds like it represents just one unit.

AppController/djinn.rb
```diff
@@ -6324,15 +6184,23 @@ def get_all_stats(secret)
       }
     end
 
-    all_stats["public_ip"] = my_node.public_ip
-    all_stats["private_ip"] = my_node.private_ip
-    all_stats["roles"] = my_node.jobs or ["none"]
-    Djinn.log_debug("All stats: #{all_stats}")
+    node_stats["cloud"] = my_node.cloud
```
cdonati: Does this value identify which cloud the node is in? We may wish to deprecate this value until we can properly support multi-cloud deployments.

InfrastructureManager/system_manager.py
```diff
@@ -160,6 +174,38 @@ def get_swap_usage(self, secret):
...
+    Returns:
+      A dictionary containing average load for last 1, 5 and 15 minutes,
+      and information about running and scheduled entities,
+      and PID of the most recently added process.
+    """
+    if self.secret != secret:
+      return self.__generate_response(False,
+        InfrastructureManager.REASON_BAD_SECRET)
+
+    loadavg = open("/proc/loadavg").read().split()
```
cdonati: I would recommend using a with statement here to ensure the descriptor is closed.

InfrastructureManager/system_manager.py
```diff
@@ -160,6 +174,38 @@ def get_swap_usage(self, secret):
...
+    kernel_entities = loadavg[3].split("/")
+    loadavg_stat = { JSONTags.LOADAVG :
+      {
+        JSONTags.LAST_1_MIN : float(loadavg[0]),
+        JSONTags.LAST_5_MIN : float(loadavg[1]),
+        JSONTags.LAST_15_MIN : float(loadavg[2]),
+        JSONTags.RUNNABLE_ENTITIES : int(kernel_entities[0]),
+        JSONTags.SCHEDULING_ENTITIES : int(kernel_entities[1]),
+        JSONTags.NEWEST_PID : int(loadavg[4])
```
cdonati: I'm not sure how we would make use of this value.

AppController/djinn.rb
jeanleonov: @cdonati , yes, I guess - I saw `cloud1` as its value. I checked usages of this and found that such refactoring may affect number of tests and probably some functionality. So it may require a couple of extra builds which would find all hidden dependencies. So I wouldn't remove `cloud` in this branch.

jeanleonov:
[Build #2966](https://ocd.appscale.com:8080/job/Daily%20Build/2966/) - single failure in okay tests was manually verified and is irrelevant to this PR

PR #2370, opened 2017-03-09
Dashboard uses new stats and "poll" model

whoarethebritons:
@jeanleonov Do not merge until it has been reviewed and comments have been addressed :)


InfrastructureManager/system_manager.py
```diff
@@ -111,11 +111,13 @@ def get_memory_usage(self, secret):
         InfrastructureManager.REASON_BAD_SECRET)
 
     mem_stats = psutil.virtual_memory()
+    # Conversion divisor for kB to MB.
+    conversion = 1024 * 1024
```
cdonati: It makes more sense to me to deal in bytes at the API level.

AppDashboard/lib/app_dashboard_data.py
```diff
@@ -351,91 +284,43 @@ def update_head_node_ip(self):
...
+      lastHourDateTime = timestamp - datetime.timedelta(hours=1)
+      old_requests_query = RequestInfo.query(RequestInfo.timestamp <
+                                             lastHourDateTime)
+      old_requests = []
+      for key in old_requests_query.iter(keys_only=True):
+        old_requests.append(key)
+      ndb.delete_multi(old_requests)
+      reversed_time = (2 ** 34 - int(request_info.get('timestamp'))) * 1000000
+      request_stats = RequestInfo(
+                      id=app_id + str(reversed_time),  # puts entities time desc.
```
cdonati: You can let the datastore assign an ID since we aren't getting it directly anywhere.

whoarethebritons: @cdonati ok I'll change that, that was how it was previously done so I had kept it.

AppController/djinn.rb
```diff
@@ -1289,10 +1289,15 @@ def update_node_info_cache()
...
           end
         end
-        new_stats << node_stats
+        Thread.current[:node_stats] = node_stats
       }
-      @cluster_stats = new_stats
     }
+    new_stats = []
+    threads.each do |t|
+      t.join
```
obino: The join here will make the AC on the shadow to wait for the response, and if one of the node is down, it could also incur in some timeouts. If you do so, I would suggest to add a lock at the beginning, and skip the update if a previous update is running.

AppController/djinn.rb
```diff
@@ -5593,8 +5529,8 @@ def try_to_scale_up(app_name, delta_appservers)
       @cluster_stats.each { |node|
         next if node['private_ip'] != host
 
-	    # Convert total memory to MB
-        total = node['memory']['total']/1024/1024
+	      # Convert total memory to MB
```
obino: Comment doesn't seems to line up: perhaps there is a tab? Sentence should finish with a period.

AppController/djinn.rb
```diff
@@ -5606,7 +5542,7 @@ def try_to_scale_up(app_name, delta_appservers)
 
         # Now we do a similar calculation but for the current amount of
         # available memory on this node. First convert bytes to MB
-        host_available_mem = node['memory']['available']/1024/1024
+        host_available_mem = Float(node['memory']['available']/1024/1024)
```
obino: I noticed that 1024/1024 is used few times: perhaps we should use a contant MB? Or something similar?

AppController/djinn.rb
```diff
@@ -5878,47 +5814,27 @@ def remove_appserver_process(app_id, port)
...
-  #     request rate for the given application.
-  #   request_rate: An Integer that indicates how many requests were served for
-  #     the given application in the last second since we queried it.
+  #   secret: A String that authenticates callers.
   # Returns:
-  #   true if the request info was successfully sent, and false otherwise.
-  def send_request_info_to_dashboard(app_id, timestamp, request_rate)
+  #   A JSON string containing the average request rate, timestamp, and total
+  # requests seen for the given application.
+  def get_request_info(secret, app_id)
```
obino: secret is usually the last parameter in the call, at least for most of the functions in djinn.

AppController/djinn.rb
whoarethebritons: @obino I am also fine with removing it I added it for time improvement. What do you think?

PR #2371, opened 2017-03-09
Single monit start for multiple ports

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2892/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2893/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2895/

AppController/lib/monit_interface.rb
```diff
@@ -116,27 +124,39 @@ def self.write_monit_config(watch, start_cmd, stop_cmd, port,
...
-    logfile = "/var/log/appscale/#{watch}-#{port}.log"
+    if port.nil?
+      new_start_cmd = "#{start_cmd}"
+      new_stop_cmd = "#{stop_cmd}"
+      new_match_cmd = "#{match_cmd}"
+      suffix = ""
+    else
+      new_start_cmd = "#{start_cmd} -p #{port}"
+      new_stop_cmd = "#{stop_cmd} #{port}"
+      new_match_cmd = "#{new_start_cmd}" if match_cmd == start_cmd
```
cdonati: What's the if statement for here? Wouldn't it be problematic if this was not defined?

obino: @cdonati it is defined on both sides of the if so it will be defined afterward.

obino:
@jeanleonov good question: since we don't have a discovery mechanism, we do indeed rely on default port for quite a few services. So this branch shouldn't break it, but makes the use of default port less explicit. I assume that the real fix would be a registration mechanism for each service so that we can use dynamic ports, or something like zeroconf or avahi to discover services.

AppController/lib/monit_interface.rb
cdonati: If `port` is not nil, and `match_cmd` != `start_cmd`, `new_match_cmd` will not be defined.

obino: duh! thanks @cdonati

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2959/

obino:
OCD migration job https://ocd.appscale.com:8080/job/Migration/1200/

PR #2372, opened 2017-03-09
enable application when user update an app

obino:


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2896/

obino:
OCD tool test https://ocd.appscale.com:8080/job/Tools-test/1589

AppController/djinn.rb
```diff
@@ -1955,6 +1955,25 @@ def update(apps, secret)
...
+        begin
+          result = uac.enable_app(appid)
+          Djinn.log_debug("enable_app returned #{result}.")
+          break
+        rescue FailedNodeException
+          # Failed to talk to the UserAppServer: let's try again.
+          Djinn.log_debug("Failed to talk to UserAppServer for #{appid}.")
+        end
+        if tries == 0
+          Djinn.log_warn("Couldn't enable application #{appid}!")
```
cdonati: Wouldn't it be better to return a failure message to the client and prevent the AppController from trying to continue setting up the app?

obino: The problem is that right now update receive a list of apps, so I'm not sure we should work as in a trasaction that is to fail them all atomically or not. I left the old logic for now, which try to start them independently. If we change update to only accept one app at a atime, then returning an error here would make more sense to me.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2900/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2904/

obino:
Got few extra runs with migration, since it stresses more the restore of the application:

https://ocd.appscale.com:8080/job/Migration/1151/
https://ocd.appscale.com:8080/job/Migration/1150/
https://ocd.appscale.com:8080/job/Migration/1149/

AppController/djinn.rb
```diff
@@ -1949,14 +1949,35 @@ def update(apps, secret)
...
+      uac = UserAppClient.new(my_node.private_ip, @@secret)
+      RETRIES.downto(0) { |tries|
+        begin
+          result = uac.enable_app(appid)
+          Djinn.log_debug("enable_app returned #{result}.")
+          APPS_LOCK.synchronize {
+            setup_app_dir(appid, true)
+          }
+          break
+        rescue FailedNodeException
```
cdonati: Please only rescue the code that can throw the exception. Otherwise, it's more difficult to track down the problem when we encounter the error.

AppController/djinn.rb
```diff
@@ -1949,14 +1949,35 @@ def update(apps, secret)
...
+          # Failed to talk to the UserAppServer: let's try again.
+          Djinn.log_debug("Failed to talk to UserAppServer for #{appid}.")
+        end
+        if tries == 0
```
cdonati: The last try might have succeeded.

obino: if it did we wouldn't be here since we have a break above.

cdonati: The break just breaks out of the retries loop.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2911/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2915/

PR #2375, opened 2017-03-10
Use client-defined value when allocating IDs

cdonati:
Previously, we had a check-and-set problem where we would first modify the counter and then make another call to ZooKeeper to retrieve the (now possibly changed) value to use for an ID.

### DIFF
--- a/AppDB/appscale/datastore/zkappscale/inspectable_counter.py
+++ b/AppDB/appscale/datastore/zkappscale/inspectable_counter.py
@@ -1,0 +1,98 @@
+""" A ZooKeeper counter that returns the new value when incremented. """
+
+from kazoo.exceptions import BadVersionError
+from kazoo.retry import ForceRetryError
+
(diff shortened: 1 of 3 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2903/

AppDB/appscale/datastore/zkappscale/inspectable_counter.py
```diff
@@ -0,0 +1,98 @@
...
+    self._ensure_node()
+    old, stat = self.client.get(self.path)
+    old = old.decode('ascii') if old != b'' else self.default
+    version = stat.version
+    data = self.default_type(old)
+    return data, version
+
+  @property
+  def value(self):
+    """ Retrive the current value from ZooKeeper.
```
menivaitsi: Retrive -> Retrieve

PR #2377, opened 2017-03-13
Use query value when cursor has multiple values

cdonati:
When a cursor contains an entity that has multiple values with the same property name, the start_row for the index scan should fall within the range given by the query filters.

This fixes single-property cursor queries that have an equality filter on a multi-valued property.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2922/

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -1002,10 +1003,21 @@ def __get_start_key(self, prefix, prop_name, order, last_result):
...
 
     val = None
-    for p in plist:
-      if p.name() == prop_name:
-        val = str(encode_index_pb(p.value()))
-        break
+
+    # If the entity provided by the cursor has multiple values for the given
+    # property and the query has information about what the value should be,
+    # use the value from the query.
```
jeanleonov: Wouldn't be more accurate to say:
> Use the value from the query if possible.

This if statement doesn't actually check if entity has multiple values for the given property.

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -1002,10 +1003,21 @@ def __get_start_key(self, prefix, prop_name, order, last_result):
...
+    if (query is not None and query.filter_size() == 1 and
+        query.filter(0).op() == datastore_pb.Query_Filter.EQUAL):
```
jeanleonov: What if we have more complex query with 2 filters or with one but not EQUAL?

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -1733,7 +1745,7 @@ def __single_property_query(self, query, filter_info, order_info):
       cursor = appscale_stub_util.ListCursor(query)
       last_result = cursor._GetLastResult()
       startrow = self.__get_start_key(prefix, property_name, direction, 
-        last_result)
+        last_result, query=query)
```
jeanleonov: Shouldn't other 4 usages of `__get_start_key` pass query parameter too?

AppDB/appscale/datastore/datastore_distributed.py
cdonati: Good point. I will update that comment.

AppDB/appscale/datastore/datastore_distributed.py
cdonati: We will need to do more work to behave properly in those scenarios. The value of the cursor entity can be ambiguous when the filter is not EQUAL.

The GAE docs mention that using a cursor with an inequality filter for a property with multiple values gives unexpected results: https://cloud.google.com/appengine/docs/standard/python/datastore/query-cursors#limitations_of_cursors, but we need to make sure we do the same thing (which probably does not involve an infinite loop).

AppDB/appscale/datastore/datastore_distributed.py
cdonati: Eventually, I would like to cover all the edge cases. We will probably need to refactor the logic of `__get_start_key` to behave more like a filter.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3149/
The first hawkeye failure was a rabbitmq issue. The second was a connection timeout.
https://ocd.appscale.com:8080/job/Blobstore/2077/

cdonati:
Hawkeye Test:
  master: https://ocd.appscale.com:8080/job/Hawkeye/11095/
  branch: https://ocd.appscale.com:8080/job/Hawkeye/11097/

PR #2382, opened 2017-03-14
Only use RealRPC in runtime stubs

cdonati:
The throttling in the bulkloader is not compatible with the threads created by RealRPC calls.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2935/
The hawkeye failure was due to RabbitMQ failing to come up.

PR #2384, opened 2017-03-17
AppServer restarts after new code is deployed

obino:
This PR fixes a race condition where AppServers could have been
restarted before the new code was deployed on each App Engine node.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/2954

PR #2385, opened 2017-03-21
Allocate entity IDs in blocks

cdonati:
This introduces several changes:
- It replaces the ZooKeeper counter Cassandra LWTs.
- It uses a new range for automatically generated IDs that matches the scattered ID allocator in GAE.
- It allocates large blocks of IDs at a time to improve performance.


AppDB/appscale/datastore/cassandra_env/entity_id_allocator.py
```diff
@@ -0,0 +1,243 @@
...
+
+  def next(self):
+    """ Generates a new entity ID.
+
+    Returns:
+      An integer specifying an entity ID.
+    """
+    # This function will require a tornado lock when made asynchronous.
+    if self.start_id is None or self.start_id > self.end_id:
+      self.start_id, self.end_id = self.allocate_size(DEFAULT_RESERVATION_SIZE)
```
jeanleonov: Does this mean that each start of AppDB instance costs at least 100000 reserved IDs?

cdonati: It doesn't allocate allocate a block until the first time it needs a scattered ID, but in practice, yes, each start will likely reserve that many IDs for each active project.

It's definitely a trade-off between allocation frequency and efficiency of ID usage. This size allows around 22 billion starts assuming the servers allocate an ID automatically during their life time.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3158/

PR #2386, opened 2017-03-21
Set timeout on opener if possible

cdonati:
This prevents the global socket timeout value from getting changed when the opener supports a timeout parameter.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2969/
https://ocd.appscale.com:8080/job/Euca-Faststart/867/

AppServer/google/appengine/tools/appengine_rpc.py
```diff
@@ -521,8 +522,10 @@ def Send(self, request_path, payload="",
     if 'AUTH_DOMAIN' in os.environ:
       auth_domain = os.environ['AUTH_DOMAIN'].lower()
 
+    opener_has_timeout = 'timeout' in inspect.getargspec(self.opener.open).args
     old_timeout = socket.getdefaulttimeout()
```
jeanleonov: Initialization of `old_timeout` could be moved to if statement.

AppServer/google/appengine/tools/appengine_rpc.py
```diff
@@ -521,8 +522,10 @@ def Send(self, request_path, payload="",
     if 'AUTH_DOMAIN' in os.environ:
       auth_domain = os.environ['AUTH_DOMAIN'].lower()
 
+    opener_has_timeout = 'timeout' in inspect.getargspec(self.opener.open).args
```
jeanleonov: Cool! I haven't ever used this feature of `inspect`.

AppServer/google/appengine/tools/appengine_rpc.py
cdonati: I agree. Unfortunately, PyCharm complains that `old_timeout` might be referenced before assignment in the finally block. I don't know the best way to organize it without adding a warning.

AppServer/google/appengine/tools/appengine_rpc.py
```diff
@@ -611,7 +617,8 @@ def Send(self, request_path, payload="",
           else:
             raise
     finally:
-      socket.setdefaulttimeout(old_timeout)
+      if not opener_has_timeout:
+        socket.setdefaulttimeout(old_timeout)
```
whoarethebritons: @cdonati @jeanleonov could old_timeout be initialized down here since its not used before this point?

jeanleonov: @whoarethebritons , it couldn't, because it should be initialized by original defaulttimeout which is rewrote in the beginning of this method

AppServer/google/appengine/tools/appengine_rpc.py
jeanleonov: Okay) let's make PyCharm happy.
Then the branch is ready to be merged.

cdonati:
Unfortunately, this change did not fix the socket timeout issue.

PR #2387, opened 2017-03-22
Use eventlet pool for Celery workers

cdonati:


cdonati:
@jeanleonov It's a similar concept to tornado in that it allows concurrency without threads. The process maintains its own "green threads", which are basically the same as tornado's coroutines.

The main difference is that the context switch is handled automatically whenever the patched `httplib` starts an io-bound operation. In tornado, this switch needs to be done manually with `yield`.

Though my preference is for the way tornado does it, celery has built-in support for eventlet.

cdonati:
I'm going to resubmit this as a new PR since the worker module changes have been merged.

PR #2390, opened 2017-03-23
Hermes Stats Framework

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2995/


Hermes/hermes.py
```diff
@@ -248,18 +276,39 @@ def main():
   logging.info("Hermes is up and listening on port: {0}.".
     format(options.port))
 
-  # Periodically check with the portal for new tasks.
-  # Note: Currently, any active handlers from the tornado app will block
-  # polling until they complete.
-  PeriodicCallback(poll, hermes_constants.POLLING_INTERVAL).start()
+  master = appscale_info.get_private_ip() in \
+           appscale_info.get_load_balancer_ips()
```
cdonati: I don't remember how our discussion concluded, but I'm not sure we need multiple machines performing stats collection for the entire cluster.

Hermes/hermes.py
```diff
@@ -235,6 +259,10 @@ def main():
   app = tornado.web.Application([
     (MainHandler.PATH, MainHandler),
     (TaskHandler.PATH, TaskHandler),
+    (NodeStatsHandler.PATH, NodeStatsHandler, {'STATS': STATS,
+                                         'secret': appscale_info.get_secret()}),
```
cdonati: Reading from the filesystem for each request seems excessive.

Hermes/helper.py
```diff
@@ -262,6 +267,46 @@ def get_cluster_stats():
...
+  """ Collects stats from this node.
+
+  Returns:
+    A dictionary containing all the monitoring stats.
+  Raises:
+    AppControllerException if we cannot talk to our AppController,
+    InfrastructureManagerException if we cannot talk to the
+    InfrastructureManagerService.
+  """
+  acc = appscale_info.get_appcontroller_client(head_node=False)
```
cdonati: This client probably does not need to be re-instantiated each loop.

Hermes/hermes.py
```diff
@@ -37,8 +39,17 @@
...
 import appscale_info
 import appscale_utils
+
+sys.path.append(os.path.dirname(__file__) + '/lib')
+from custom_exceptions import InfrastructureManagerException
 # Tornado web server options.
 define("port", default=hermes_constants.HERMES_PORT, type=int)
+# Determines whether this node is a master Hermes node. A slave Hermes node
+# will only collect its node's statistics.
+define("master", default=False, type=bool)
```
cdonati: Where is this option used?

Hermes/hermes.py
```diff
@@ -100,6 +111,19 @@ def poll():
   # The poller can move forward without waiting for a response here.
   helper.urlfetch_async(request)
 
+def cache_node_stats():
+  try:
+    STATS['node'] = helper.update_node_cache()
+  except (AppControllerException, InfrastructureManagerException) as e:
```
cdonati: Pylint complains about single-letter variable names. I'm curious about the reason for this wrapper function instead of just catching the exceptions where they can be raised.

Hermes/hermes.py
```diff
@@ -100,6 +111,19 @@ def poll():
...
+    logging.error("Saw exception while trying to update node statistics cache: "
```
cdonati: logging.exception is usually cleaner and more helpful in these cases.

Hermes/hermes.py
```diff
@@ -248,18 +276,39 @@ def main():
...
-  # Periodically check with the portal for new tasks.
-  # Note: Currently, any active handlers from the tornado app will block
-  # polling until they complete.
-  PeriodicCallback(poll, hermes_constants.POLLING_INTERVAL).start()
+  master = appscale_info.get_private_ip() in \
+           appscale_info.get_load_balancer_ips()
+
+  # Cache this node's statistics immediately so that the master Hermes nodes
+  # can use it as soon as possible.
+  cache_node_stats()
```
cdonati: PeriodicCallback schedules the first run immediately, so this is not needed.

Hermes/hermes.py
```diff
@@ -100,6 +111,19 @@ def poll():
   # The poller can move forward without waiting for a response here.
   helper.urlfetch_async(request)
 
+def cache_node_stats():
```
cdonati: Please write docstrings for all new functions.

Hermes/lib/infrastructure_manager_client.py
```diff
@@ -0,0 +1,118 @@
...
+  # The message that an AppController can return if callers do not authenticate
+  # themselves correctly.
+  BAD_SECRET_MESSAGE = {'success': False, 'reason': 'bad secret'}
+
+  # The number of times we should retry SOAP calls in case of failures.
+  DEFAULT_NUM_RETRIES = 5
+
+
+  def __init__(self, host, secret):
+    """Creates a new AppControllerClient.
```
cdonati: Please replace all references to AppController in this module.

AppController/djinn.rb
```diff
@@ -1273,26 +1276,21 @@ def get_cluster_stats_json(secret)
...
-          rescue FailedNodeException
-            Djinn.log_warn("Failed to get status update from node at #{ip}, so " +
-              "not adding it to our cached info.")
-            next
-          end
-        end
-        new_stats << node_stats
+      ip = get_shadow.private_ip
+      uri = URI("http://#{ip}:#{HermesService.getport()}/cluster_stats")
+      string_stats = (Net::HTTP.post_form(uri, 'secret' => @@secret))
```
whoarethebritons: @cdonati I had been using a get but then it complained that the response body should be empty (if I remember correctly). I would be happy to change back to get because I felt like that was the correct thing to use as well.

Hermes/handlers.py
```diff
@@ -143,3 +146,43 @@ def post(self):
...
+class NodeStatsHandler(RequestHandler):
+  """ Main handler class. """
+
+  # The path for this handler.
+  PATH = "/node_stats"
+
+  def initialize(self, STATS, secret):
+    self.STATS = STATS
+    self.secret = secret
+    self.BAD_SECRET_MESSAGE = {'success': False, 'reason': 'bad secret'}
```
whoarethebritons: @cdonati the message part I'll change, but the STATS attribute is so that it can use the same STATS variable that's being used in hermes, that was the way I could use the variable across files supposedly.

Hermes/hermes.py
whoarethebritons: @cdonati if I caught them in the helper I couldn't log anything about them, or that was the experience I had with it.

PR #2397, opened 2017-03-31
Remove unnecessary db request while adding task

cdonati:
Using the timestamp assigned by the coordinator node does not provide any benefit over using a client-provided timestamp.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3024/

PR #2398, opened 2017-03-31
Log exceptions encountered on API server

cdonati:
This will help pinpoint gateway timeouts and other issues that the API server can experience.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3025/
The Hawkeye failure was due to a RabbitMQ restart.

cdonati:
We decided that the extra verbosity was not worth it as long as the api proxy process is logging to the same file as the runtime.

PR #2401, opened 2017-04-02
Improve lease performance

cdonati:
On my machine, this cut lease times in half when leasing in batches of 500.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3034/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3047/

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -125,6 +125,7 @@ def __init__(self, queue_info, app):
         self.task_retry_limit = retry_params['task_retry_limit']
 
     self.validate_config()
+    self.prepared_statements = {}
```
jeanleonov: How long does Queue object usually live? Couldn't this dict be cause of a memory leak? If it can live quite long, then it makes sense to limit size of the dict. As an option third party lib can be used (https://pypi.python.org/pypi/expiringdict/1.1.2) or we could implement our own.

cdonati: It usually lives forever. However, it should have at most about 10 items in it since prepared statements do not have values bound to them yet.

jeanleonov: @cdonati , do you mean it should be collected by garbage collector? Or how it explains that there is no risk of memory leak? What I see is that this dict is only extended through the time. Every time new statement is prepared it gets new item. And as this dict is referenced by Queue object and since the dict holds references to prepared statements, it seems that none of them will be collected by garbage collector.

cdonati: It shouldn't be garbage collected. It's trading a bit of memory usage for better CPU performance. It would definitely be a leak if the items in this dict are no longer useful. However, there can only ever be 5 items in this dictionary (because only 5 prepared statements are ever cached per queue), and they are used any time the relevant operation is performed.

The keys are basically templates for common operations, and if the the statement (template) has already been prepared, the function can use that instead of adding a new item to the dict.

jeanleonov: Ohhh.. Sorry, I missed the fact that all these statements are string constants=( After the first look at these I was sure that they're format strings which get different parameters every time and so will be different for each call of function. Now I see that parameters of statements get its values after statement are prepared.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3145/
The hawkeye failure was due to a connection timeout.

PR #2407, opened 2017-04-05
Continue build when repo update is not required

cdonati:
This allows users to re-run "upgrade" if the first time fails.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -43,8 +43,6 @@ if [ "$HOME" != "/root" ]; then
 fi
 echo "Success"
 
-set -e
-
 # Let's get the  command line arguments.
 while [ $# -gt 0 ]; do
         if [ "${1}" = "--repo" ]; then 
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -252,7 +255,7 @@ if [ -d /etc/appscale/certs ]; then
 
 
         # Let's upgrade the repository: if GIT_TAG is empty we are on HEAD.
-        if [ -n "${GIT_TAG}" ]; then
+        if [ -n "${GIT_TAG}" ] && [ "${UPDATE_REPO}" = "Y" ]; then
```
obino: as an FYI you could have a single [] and have '-a' in between conditions

cdonati: Some style guides prefer avoiding `-a` for clarity: https://github.com/koalaman/shellcheck/wiki/SC2166

obino: interesting: good catch.

obino:
OCD run for creating image: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/3981/

PR #2408, opened 2017-04-06
fix upgrade script to handle changes since 3.1.0

whoarethebritons:


scripts/datastore_upgrade.py
```diff
@@ -77,19 +77,22 @@ def ensure_app_is_not_running():
...
 
 
-def start_cassandra(db_ips, db_master, keyname):
+def start_cassandra(db_ips, db_master, keyname, zookeeper):
   """ Creates a monit configuration file and prompts Monit to start Cassandra.
   Args:
     db_ips: A list of database node IPs to start Cassandra on.
     db_master: The IP address of the DB master.
     keyname: A string containing the deployment's keyname.
+    zookeeper: The IP addresses of the Zookeeper nodes.
```
tmarballi: Could you rename this to be 'zookeeper_ips' so that its more specific?

scripts/datastore_upgrade.py
```diff
@@ -77,19 +77,22 @@ def ensure_app_is_not_running():
...
   Raises:
     AppScaleDBError if unable to start Cassandra.
   """
   logging.info("Starting Cassandra...")
   for ip in db_ips:
-    init_config = '{script} --local-ip {ip} --master-ip {db_master}'.format(
-      script=SETUP_CASSANDRA_SCRIPT, ip=ip, db_master=db_master)
+    init_config = '{script} --local-ip {ip} --master-ip {db_master} ' \
+                  '--zk-locations {zk_locations}'.format(
+      script=SETUP_CASSANDRA_SCRIPT, ip=ip, db_master=db_master,
```
tmarballi: Could you fix the indentation here to be in line with the above?

PR #2411, opened 2017-04-07
Manually retry LWT statements during queue ops

cdonati:
Previously, the session's default retry policy would throw an exception if the statement encountered a timeout.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3073/

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -440,30 +421,18 @@ def update_task(self, task, new_lease_seconds):
...
     parameters = {
       'app': self.app,
       'queue': self.name,
       'id': task.id,
-      'new_eta': new_eta
+      'new_eta': new_eta,
+      'current_time': datetime.datetime.utcnow()
     }
-    if check_lease:
+    if hasattr(task, 'leaseTimestamp'):
```
jeanleonov: @cdonati , it seems that the same problem (processing of possible absence of an attribute) is solved in different way previously in this file. I wouldn't strongly recommend to use try/except, but it seems that try/except is more pythonic way: https://docs.python.org/2/glossary.html#term-eafp. Anyway it makes sense to keep style the same inside a file.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -304,9 +299,7 @@ def add_task(self, task):
     except AttributeError:
       parameters['lease_expires'] = 0
 
-    result = self.db_access.session.execute(insert_task, parameters)[0]
-    if not result.applied:
-      raise InvalidTaskInfo('Task name already taken: {}'.format(task.id))
+    self._insert_task(parameters)
```
jeanleonov: @cdonati , I guess `retries` should be passed to the `_insert_task`. Shouldn't it?

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -675,6 +644,109 @@ def to_json(self, include_stats=False, fields=None):
...
+        'Task name already taken: {}'.format(parameters['id']))
+
+  def _update_lease(self, parameters, check_lease=True, retries=5):
+    """ Update lease expiration on a task entry.
+
+    Args:
+      parameters: A dictionary specifying the new parameters.
+      check_lease: A boolean specifying that the old lease_expires field must
+        match the one provided.
+      retries: The number of times to try the update.
```
jeanleonov: @cdonati , it makes sense to extend `update_lease` and `update_task` methods with this parameter. And, as it should be optional for those, I'd suggest to make it required on the level of private method, and specify default values on the level of API methods. It's safer to have a single default instead of multiple across different methods. The same for insert task.

AppTaskQueue/appscale/taskqueue/queue.py
cdonati: Yes. I agree with this principle.

AppTaskQueue/appscale/taskqueue/queue.py
cdonati: Good point.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3106/
The hawkeye failures were connection timeouts.
https://ocd.appscale.com:8080/job/Euca-Faststart/970/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3123/

PR #2413, opened 2017-04-12
Add UAClient

cdonati:
UAClient is a wrapper that makes UAServer calls. It keeps the soap client and secret together in the same object.


common/appscale/common/ua_client.py
```diff
@@ -0,0 +1,179 @@
...
+      email: A string specifying the user's email address.
+    Raises:
+      UAException if the operation was not successful.
+    """
+    response = self.server.disable_user(email, self.secret)
+    if response.lower() != 'true':
+      raise UAException(response)
+
+  def does_app_exist(self, app_id):
+    """ Checks if a user exists.
```
whoarethebritons: user -> app

PR #2414, opened 2017-04-18
Add AdminServer

cdonati:
This is a new server meant to implement the Admin API, handling application deployments.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3103/
https://ocd.appscale.com:8080/job/Migration/1337/
Please see https://github.com/AppScale/appscale-tools/pull/614

AdminServer/appscale/admin/utils.py
```diff
@@ -0,0 +1,115 @@
...
+    output['error'] = operation['error']
+
+  if 'response' in operation:
+    output['response'] = operation['response']
+
+  return output
+
+
+def format_version(operation, status, create_time, http_port):
+  """ Formmats a version for the client.
```
obino: Typo in Formats.

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+      try:
+        details.update(error.kwargs)
+      except AttributeError:
+        pass
+
+    self.finish(json_encode({'error': details}))
+
+
+class VersionsHandler(BaseHandler):
+  """ Manages service versions. """
```
cdonati: No, it's version as in App Engine project services and versions. Though we only support one version (default) and one service (default), it makes sense to me to start using the structure and terminology of the API, which supports multiple services and modules.

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+    except UAException:
+      message = 'Unable to retrieve project metadata'
+      logging.exception(message)
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR, message=message)
+
+    if 'owner' not in project_metadata:
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
+                            message='Project owner not defined')
+
+    if project_metadata['owner'] != user:
```
cdonati: As of now, the UAServer only supports one owner per project.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3130/
Migration: https://ocd.appscale.com:8080/job/Migration/1359/

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+@gen.coroutine
+def wait_for_port_assignment(operation_id, deadline, acc):
+  """ Waits until port is assigned for version.
+
+  Args:
+    operation_id: A string specifying an operation ID.
+    deadline: A float containing a unix timestamp.
+    acc: An AppControllerClient.
+  Raises:
+    OperationTimeout if the deadline is exceeded.
```
jeanleonov: It also may raise `KeyError` if operation with `operation_id` was removed from cache. It seems to be regular behavior of the function.

jeanleonov: Or probability of removing active operations is significantly low?

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+
+  Args:
+    http_port: An integer specifying the version's port number.
+    operation_id: A string specifying an operation ID.
+    deadline: A float containing a unix timestamp.
+  Raises:
+    OperationTimeout if the deadline is exceeded.
+  """
+  logging.debug('Waiting for {} to open'.format(http_port))
+  operation = operations[operation_id]
```
jeanleonov: The same here (is KeyError when accessing operations cache considerably possible)

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+
+@gen.coroutine
+def wait_for_deploy(operation_id, acc):
+  """ Tracks the progress of a deployment.
+
+  Args:
+    operation_id: A string specifying the operation ID.
+    acc: An AppControllerClient instance.
+  """
+  operation = operations[operation_id]
```
jeanleonov: And here

AdminServer/appscale/admin/constants.py
```diff
@@ -0,0 +1,58 @@
...
+  JAVA,
+  GO,
+  PHP
+)
+from tornado.web import HTTPError
+
+
+class CustomHTTPError(HTTPError):
+  """ An HTTPError that keeps track of keyword arguments. """
+  def __init__(self, status_code=500, **kwargs):
```
jeanleonov: `CustomHTTPError` is usually used with `message`argument. But `tornado.HTTPError` supports `reason` keyword argument. I think it makes sense to not limit basic functionality of `HTTPError` and pass all three argument which `HTTPError` handles: `status_code, log_message, reason`.

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+    source_path = version['deployment']['zip']['sourceUrl']
+    if source_path.startswith(proto_prefix):
+      source_path = source_path[len(proto_prefix):]
+
+    self.begin_deploy(project_id, source_path)
+
+    operation_id = str(uuid.uuid4())
+    insert_time = datetime.datetime.utcnow()
+
+    operations[operation_id] = {
```
jeanleonov: operation entity has strict structure and it's used many times. Doesn't it worth to create a model for this? It would reduce amount of representative string constants..

AdminServer/appscale/admin/__init__.py
cdonati: It would be unusual if operation was not in the cache since this function gets called immediately after the operation is added. However, I agree that that possibility should be dealt with. For now, I converted the KeyError to OperationTimeout since that seems more reasonable to catch by the caller.

AdminServer/setup.py
```diff
@@ -0,0 +1,28 @@
...
+  name='appscale-admin',
+  version='0.0.1',
+  description='An implementation of the Google App Engine Admin API',
+  author='AppScale Systems, Inc.',
+  url='https://github.com/AppScale/appscale',
+  license='Apache License 2.0',
+  keywords='appscale google-app-engine python',
+  platforms='Posix',
+  install_requires=[
+    'appscale-common',
```
jeanleonov: Where this package can be found?

AdminServer/appscale/__init__.py
```diff
@@ -0,0 +1 @@
+__import__('pkg_resources').declare_namespace(__name__)
```
jeanleonov: I like to see something what I didn't know about python=)
It forces me to google and learn more.
But... why do you use `__import__`? [Documentation](https://docs.python.org/2/library/functions.html#__import__) says that it should be used only in case you want to import a module whose name is only known at runtime.

I'd suggest to replace it with
```
import pkg_resources
pkg_resources.declare_namespace(__name__)
```

AdminServer/setup.py
jeanleonov: I see..

AdminServer/appscale/__init__.py
cdonati: The setuptools docs claim that line is required for namespace packages: http://setuptools.readthedocs.io/en/latest/setuptools.html#namespace-packages

`import` probably works too, but I think it would cause `appscale.pkg_resources` to be defined.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3146/
Migration: https://ocd.appscale.com:8080/job/Migration/1376/ (see https://github.com/AppScale/appscale-testing/pull/221)

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,400 @@
...
+
+    try:
+      info_map = acc.get_app_info_map()
+    except AppControllerException as error:
+      logging.warning('Unable to fetch info map: {}'.format(error))
+      yield gen.sleep(1)
+      continue
+
+    try:
+      raise gen.Return(info_map[project_id]['nginx'])
```
tmarballi: This might be my Python shortcomings, but is there a reason we have to do a 'raise' for returning the port?

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,400 @@
...
+  try:
+    operation = operations[operation_id]
+  except KeyError:
+    raise OperationTimeout('Operation no longer in cache')
+
+  start_time = time.time()
+  deadline = start_time + MAX_DEPLOY_TIME
+
+  http_port = yield wait_for_port_assignment(operation_id, deadline, acc)
+  yield wait_for_port_to_open(http_port, operation_id, deadline)
```
tmarballi: I learned something new with 'yield' today 👍

AdminServer/appscale/admin/__init__.py
```diff
@@ -0,0 +1,400 @@
...
+from google.appengine.api.appcontroller_client import AppControllerException
+
+
+# The state of each operation.
+operations = OperationsCache()
+
+
+@gen.coroutine
+def wait_for_port_assignment(operation_id, deadline, acc):
+  """ Waits until port is assigned for version.
```
tmarballi: We don't really have concept of versions yet right, so is this so that we use that terminology or we made some specific changes for that?

AppDB/appscale/datastore/backup/cassandra_backup.py
```diff
@@ -238,4 +239,21 @@ def restore_data(path, keyname, force=False):
     utils.ssh(db_ip, keyname,
       'monit start {}'.format(CASSANDRA_MONIT_WATCH_NAME))
 
+  logging.info('Waiting for all nodes in cluster to start')
```
tmarballi: Is this change dependent on the Admin Server changes?

AdminServer/appscale/admin/__init__.py
cdonati: It's just a conceptual change. The idea now is that we will start by supporting just the default service/version. Once that conceptual change is there, it will be much easier to support multiple services and versions.

AppDB/appscale/datastore/backup/cassandra_backup.py
cdonati: This was a fix for the Migration job. In master, the tools persistently retry UAServer-related calls indefinitely, but the AdminServer just returns an error when unable to deploy. With https://github.com/AppScale/appscale-testing/pull/221, I don't think this change is needed, so I will remove it from this pull request.

AdminServer/appscale/admin/__init__.py
cdonati: Python 2 does not allow you to return values from a generator. This is kind of a hack required by tornado for coroutines. http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.coroutine

tmarballi: That makes sense, Thanks for clarification @cdonati

AdminServer/appscale/admin/utils.py
```diff
@@ -0,0 +1,51 @@
...
+      return
+
+    prefix += '.{}'.format(field_name)
+    return missing_field(prefix, group, resource_part[field_name])
+
+  missing_fields = []
+  for group in required_fields:
+    field = missing_field(resource_name, group.split('.'), resource)
+    if field is not None:
+      missing_fields.append(missing_field)
```
whoarethebritons: Not sure if I missed something, but should it be `missing_fields.append(field)` ?

common/appscale/common/ua_client.py
```diff
@@ -36,6 +36,23 @@ def add_admin_for_app(self, email, app_id):
...
+
+    Args:
+      app_id: A string specifying an application ID.
+      email: A string specifying the user's email address.
+      language: A string specifying the project's language.
+    """
+    response = self.server.commit_new_app(app_id, email, language, self.secret)
+
+    # If the project already exists, consider the operation a success.
+    if response == 'Error: appname already exist':
```
whoarethebritons: Could this be a constant that they both use so that if we ever want to change it we only have to change it in one place?

AdminServer/appscale/admin/utils.py
cdonati: Yes, good catch.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3152/

cdonati:
This must go in with https://github.com/AppScale/appscale-tools/pull/614 and https://github.com/AppScale/appscale-testing/pull/221

PR #2417, opened 2017-04-25
Exit controller service with error when not configured

cdonati:
Otherwise, the init system will not know that the start failed.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3133/
The Hawkeye failure was due to a rabbitmq cookie issue.

PR #2418, opened 2017-04-25
Disable ssl verification for SOAPPy

cdonati:
This is a temporary measure until SOAPpy can be replaced.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3125/
The Hawkeye error was a connection timeout.

PR #2419, opened 2017-04-25
Remove groomer log file check

cdonati:
This was originally added as a fallback for when the groomer froze, but that issue is fixed now that the groomer does not run the ZooKeeper GC thread.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3131/
The first hawkeye failure was due to Cassandra connection issues. The second one was the usual proxy connection timeout.

PR #2421, opened 2017-04-26
Only start one process per BRServer

cdonati:
Otherwise, tornado will start one process per core, making it more difficult to monitor with a pidfile.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3138/

PR #2424, opened 2017-04-28
New hermes stats

jeanleonov:
Here you can see changes to Hermes only (at least AppController need to be changed to make appscale work - will do next week).

The structure stats modules is following:
```
hermes
  stats
    producers
       cluster_stats        gets stats from other nodes (reads using timestamp as cursor)
       node_stats           gets nodes stats
       processes_stats     gets processes stats for all monit processes
       proxies_stats      gets proxies stats from haproxy
    subscribers
       cache             receives different kinds of stats and writes it to cache (limited size and TTL are implemented
       portal          will send stats to portal
       profile         will write stats to csv tables
    stats_app        module responsible for configuring all pubsubs and api handlers (process stats can be disabled, proxies stats is available on LB nodes only, cluster stats is available only on master node)
    handlers          handlers to provide access to stats caches, current stats and to cluster stats
    pubsub_base
    unified_service_names  
```
Review volunteers are welcome, but I'll write more docs later, so it makes to wait unless you're want to see what's going on with stats now. 

p.s.: have to go.. will try to add more comments and respond on your comments during weekend and my day-offs


Hermes/appscale/hermes/helper.py
```diff
@@ -33,6 +30,11 @@
 REPORT_TASKS = ['backup', 'restore']
 
 
+class MissingRequestArgs(Exception):
+  """ An exception when HTTP Request parameters are missing. """
+  pass
```
jeanleonov: is this used?

Hermes/appscale/hermes/stats/handlers.py
```diff
@@ -0,0 +1,92 @@
...
+
+from appscale.hermes.constants import SECRET_HEADER
+
+
+class Respond404Handler(RequestHandler):
+  def initialize(self, reason):
+    self.reason = reason
+
+  def get(self):
+    self.set_status(404, self.reason)
```
jeanleonov: Move to global handlers

Hermes/appscale/hermes/stats/handlers.py
```diff
@@ -0,0 +1,92 @@
...
+    if self.request.headers.get(SECRET_HEADER) != options.secret:
+      logging.warn("Received bad secret from {client}"
+                   .format(client=self.request.remote_ip))
+      self.set_status(401, "Bad secret")
+      return
+    payload = json.loads(self.request.body)
+    last_utc_timestamp = payload.get('last_utc_timestamp')
+    limit = payload.get('limit')
+    include_lists = payload.get('include_lists')
+    fetch_latest_only = payload.get('fetch_latest_only')
```
jeanleonov: fetch_latest_only should be passed to stats_cache

Hermes/appscale/hermes/stats/producers/cluster_stats.py
```diff
@@ -0,0 +1,315 @@
...
+    url=url, method='GET', body=json.dumps(arguments), headers=headers,
+    validate_cert=False, request_timeout=REQUEST_TIMEOUT,
+    allow_nonstandard_methods=True
+  )
+  async_client = httpclient.AsyncHTTPClient()
+
+  try:
+    # Send Future object to coroutine and suspend till result is ready
+    response = yield async_client.fetch(request)
+  except Exception as e:
```
jeanleonov: Specify exact exceptions..

common/appscale/common/appscale_info.py
```diff
@@ -24,16 +24,16 @@ def read_file_contents(path):
...
+  if head_node:
+    ips = get_load_balancer_ips()
+    acc_ip = ips[0]
+  else:
+    acc_ip = get_private_ip()
   secret_file = '/etc/appscale/secret.key'
   secret = read_file_contents(secret_file)
-
-  return AppControllerClient(head_node, secret)
+  return AppControllerClient(acc_ip, secret)
```
jeanleonov: check if you need this

Hermes/appscale/hermes/stats/producers/cluster_stats.py
```diff
@@ -0,0 +1,315 @@
...
+    arguments['limit'] = limit
+  if include_lists is not None:
+    arguments['include_lists'] = include_lists
+  if fetch_latest_only:
+    arguments['fetch_latest_only'] = True
+
+  url = "http://{ip}:{port}/{path}".format(
+    ip=node_ip, port=options.port, path=method_path)
+  request = httpclient.HTTPRequest(
+    url=url, method='GET', body=json.dumps(arguments), headers=headers,
```
cdonati: Is there a reason these arguments need to be in the body instead of as standard url variables?

Hermes/appscale/hermes/stats/pubsub_base.py
```diff
@@ -0,0 +1,147 @@
...
+        subscriber.receive(stats)
+      except Exception:
+        logging.error(
+          "Failed to send {stats} to {subscriber}"
+          .format(stats=self._stats_source.stats_name, subscriber=name),
+          exc_info=sys.exc_info()
+        )
+
+  def start(self):
+    self._read_and_publish()
```
cdonati: PeriodicCallback adds the task to the stack to be executed without an initial delay, so I don't think this initial call is needed.

Hermes/appscale/hermes/stats/subscribers/cache.py
```diff
@@ -0,0 +1,153 @@
...
+  It takes care about storing snapshots in limited cache and provides
+  reading method with acknowledgment mechanism.
+  Each node has local stats cache for each kind of stats it collects
+  (node stats, processes stats and haproxy stats for LB nodes).
+  It is used as a temporary storage for stats which wasn't read by master yet.
+  """
+
+  def __init__(self, snapshots_cache_size, ttl=None):
+    self._snapshots_cache = []
+    self._cache_lock = threading.Lock()
```
cdonati: You shouldn't need a threading lock in async code as long as you haven't manually created other threads that operate on the same resources. If you are worried about a yield causing ordering issues, you can use `tornado.locks.Lock`. Since a threading lock doesn't yield, it's easy to end up with a deadlock if you use a threading lock in async code.

Hermes/appscale/hermes/stats/pubsub_base.py
```diff
@@ -0,0 +1,147 @@
...
+
+  def receive(self, stats):
+    """ Handlers produced stats entity
+    Args:
+      stats: an object containing stats of some kind (node, processes, ...)
+    """
+    raise NotImplementedError()
+
+
+class StatsPublisher(object):
```
cdonati: In my opinion, the "Publish" and "Subscribe" terminology is a little bit misleading in this context. Most of the time I've seen these terms, they indicate that a subscriber is a client that maintains a persistent connection to a service that pushes content selectively. Since stats are obtained by clients using standard http requests, this might mislead a reader.

In this case, it appears that it is more like an abstraction over filtering what gets cached. Instead of "read and publish", it seems more like "collect and cache".

Hermes/appscale/hermes/__init__.py
```diff
@@ -12,30 +9,31 @@
...
-
 from tornado.ioloop import IOLoop
 from tornado.ioloop import PeriodicCallback
-from tornado.options import define
 from tornado.options import options
-from tornado.options import parse_command_line
 
-sys.path.append(os.path.join(os.path.dirname(__file__), '../AppServer'))
+import helper
+from appscale.hermes.constants import DEFAULT_HERMES_PORT
```
jeanleonov: [PEP8#imports](https://www.python.org/dev/peps/pep-0008/#imports) accepts explicit relative imports, but it recommends to use absolute imports if packages layout is not too complex.

Hermes/appscale/hermes/__init__.py
```diff
@@ -12,30 +9,31 @@
...
+from appscale.hermes.stats.handlers import Respond404Handler
+from appscale.hermes.stats.stats_app import StatsApp
+from handlers import MainHandler, TaskHandler
+from helper import JSONTags
+
+sys.path.append('../../../AppServer')
```
jeanleonov: Thanks!

Hermes/appscale/hermes/stats/producers/proxy_stats.py
```diff
@@ -0,0 +1,502 @@
...
+
+  def get_current(self):
+    """ Method which parses haproxy stats and returns detailed
+    proxy statistics for all proxies.
+
+    Returns:
+      ProxiesStatsSnapshot
+    """
+    # Get CSV table with haproxy stats
+    csv_text = subprocess.check_output(
```
jeanleonov: I like it!

Hermes/appscale/hermes/stats/producers/cluster_stats.py
jeanleonov: `include_lists` is a dictionary of lists. It can be huge.

Hermes/appscale/hermes/stats/pubsub_base.py
jeanleonov: When I add profile log, it will appear to be "collect and cache and write to log".
When portal support is added it can become "collect and cache and write to log and send to portal.

My idea was that there are 3 consumers of stats: local cache to deal with temporary network issues, profile log and Portal.

So I still believe it can be called as it is.

Hermes/appscale/hermes/stats/subscribers/cache.py
jeanleonov: I thought tornado creates multiple threads for serving web requests and running periodical tasks, isn't it?
I added locks because periodical task adds new snapshots, when handlers reads it (and possibly change if it was requested).
I can't see any chance that this operation can be interrupted by yield.

Hermes/appscale/hermes/__init__.py
```diff
@@ -195,67 +164,93 @@ def create_xmpp_user(password, uaserver):
...
+    # Note: Currently, any active handlers from the tornado app will block
+    # polling until they complete.
+    PeriodicCallback(poll, constants.POLLING_INTERVAL).start()
 
-  logging.info("Hermes is up and listening on port: {0}.".
-    format(options.port))
+    # Only master Hermes node handles /do_task route
+    task_route = ('/do_task', TaskHandler)
+  else:
+    task_route = ('/do_task', Respond404Handler,
```
jeanleonov: I think Hermes master is more like different application from Hermes slave. I don't think it is handler's responsibility to explain why this handler is not supported on slave node.
And `404 Hermes slaves do not manage tasks from Portal` seems representative enough to me.

Hermes/appscale/hermes/stats/handlers.py
```diff
@@ -0,0 +1,92 @@
...
+  """
+
+  def initialize(self, stats_cache):
+    self._stats_cache = stats_cache
+
+  def get(self):
+    if self.request.headers.get(SECRET_HEADER) != options.secret:
+      logging.warn("Received bad secret from {client}"
+                   .format(client=self.request.remote_ip))
+      self.set_status(401, "Bad secret")
```
jeanleonov: I don't claim to implement RESTfull API, but I thinks it's good practice to use existent tools (HTTP protocol) for solving standard problems.
Clients already have to handle HTTP errors, they don't have to parse JSON body to get short error message. It don't think Hermes will have big structured error messages.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3213/

Hermes/appscale/hermes/__init__.py
cdonati: Okay, that's fine with me.

Hermes/appscale/hermes/stats/handlers.py
cdonati: Makes sense. For me, it was more about consistency. However, I'm fine with keeping internal interfaces different than user-facing ones as long as each one is consistent.

Hermes/appscale/hermes/stats/pubsub_base.py
cdonati: During our last team discussion about it, the profile log was not going to be permanent. It was just for demonstration purposes. Conceptually, the portal is the same as any other client that is polling Hermes. The mechanism is different because of potential firewall issues, so it might make sense for the http client to be a different actor or process.

I've just never heard of a local cache "subscribing" to anything. With this terminology, clients that are asking for stats are not consumers, but the service's cache is. If you feel strongly about it, I will approve. It just seems counter-intuitive to me.

Hermes/appscale/hermes/stats/subscribers/cache.py
cdonati: No, everything uses the same IOLoop (unless you explicitly create multiple IOLoops in different threads, but that would be unusual). It's one thing that makes async programming nice because you don't have to worry about some types of race conditions. It also makes it important not to block the loop.

Hermes/appscale/hermes/__init__.py
```diff
@@ -64,8 +62,8 @@ def poll():
   logging.info("Polling for new task.")
 
   # Send request to AppScale Portal.
-  url = "{0}{1}".format(hermes_constants.PORTAL_URL,
-      hermes_constants.PORTAL_POLL_PATH)
+  url = "{0}{1}".format(constants.PORTAL_URL,
```
cdonati: fixed

Hermes/appscale/hermes/helper.py
```diff
@@ -2,21 +2,19 @@
...
 import logging
-import os
 import sys
 import threading
-import tornado.httpclient
 import urllib
 
-import hermes_constants
-
+import tornado.httpclient
```
tmarballi: Move this import to the line after the threading import.

Hermes/appscale/hermes/helper.py
```diff
@@ -76,7 +79,7 @@ def create_request(url=None, method=None, body=None):
   if not url or not method:
     raise MissingRequestArgs
   return tornado.httpclient.HTTPRequest(url=url, method=method, body=body,
-    validate_cert=False, request_timeout=hermes_constants.REQUEST_TIMEOUT)
+                                        validate_cert=False, request_timeout=constants.REQUEST_TIMEOUT)
```
tmarballi: Indent request_timeout to the next line so to be within the line limit of 80.

Hermes/appscale/hermes/stats/handlers.py
```diff
@@ -0,0 +1,84 @@
+import json
+import logging
+
+from tornado.options import options
+from tornado.web import RequestHandler
+
+from appscale.hermes.constants import SECRET_HEADER
```
tmarballi: Alphabetize imports here as well.

Hermes/appscale/hermes/stats/constants.py
```diff
@@ -0,0 +1,43 @@
...
+
+# The intervals for updating cluster stats
+CLUSTER_NODES_STATS_CACHE_SIZE = 1
+CLUSTER_PROCESSES_STATS_CACHE_SIZE = 1
+CLUSTER_PROXIES_STATS_CACHE_SIZE = 1
+
+# Path to haproxy stats socket
+HAPROXY_STATS_SOCKET_PATH = '/etc/haproxy/stats'
+
+# Quite logging intervals
```
tmarballi: Did you mean 'quiet' here instead of 'quite'?

Hermes/appscale/hermes/stats/constants.py
```diff
@@ -0,0 +1,43 @@
...
+LOCAL_STATS_DEBUG_INTERVAL = 5*60
+CLUSTER_STATS_DEBUG_INTERVAL = 15*60
+
+
+class _MissedValue(object):
+  """
+  Instance of this private class denotes missed value.
+  It's used to denote values of stats properties which are missed
+  in haproxy stats csv
```
tmarballi: You are missing a period at the end of this sentence.

Hermes/appscale/hermes/stats/producers/cluster_stats.py
```diff
@@ -0,0 +1,317 @@
...
+import logging
+import sys
+import time
+
+from appscale.common import appscale_info
+from tornado import gen, httpclient
+from tornado.options import options
+from tornado.web import HTTPError
+
+from appscale.hermes import constants
```
tmarballi: Alphabetize imports here as well.

Hermes/appscale/hermes/stats/producers/cluster_stats.py
```diff
@@ -0,0 +1,317 @@
...
+  using timestamp as a cursor.
+  """
+
+  def __init__(self, local_cache, include_lists=None, limit=None,
+               fetch_latest_only=False):
+    """ Initialises an instance of ClusterNodeStatsSource.
+
+    Args:
+      local_cache: an instance of LocalStatsCache where node stats of this node
+          is cached. It's used to avoid HTTP calls to local API
```
tmarballi: "is cached" => "are cached" and missing period at the end of the statement.

Hermes/appscale/hermes/stats/producers/node_stats.py
```diff
@@ -0,0 +1,296 @@
...
+      network=NodeNetwork(**{field: network.get(field, MISSED)
+                             for field in NodeNetwork.__slots__}),
+      loadavg=NodeLoadAvg(**{field: loadavg.get(field, MISSED)
+                             for field in NodeLoadAvg.__slots__})
+    )
+
+
+def node_stats_snapshot_to_dict(stats, include_lists=None):
+  """ Converts an instance of NodeStatsSnapshot to dictionary. Optionally
+  it can
```
tmarballi: Please complete the doc string.

Hermes/appscale/hermes/stats/producers/process_stats.py
```diff
@@ -0,0 +1,384 @@
...
+
+
+class ProcessesStatsSource(StatsSource):
+
+  last_debug = 0
+
+  def get_current(self):
+    """ Method for building a list of ProcessStats.
+    It parses output of `monit status` and generates ProcessStats object
+    for each monitored service
```
tmarballi: Missing period at the end of the sentence.

Hermes/appscale/hermes/stats/producers/process_stats.py
```diff
@@ -0,0 +1,384 @@
...
+    processes_stats=[
+      process_stats_from_dict(process_stats, strict)
+      for process_stats in dictionary.get('processes_stats', [])
+    ]
+  )
+
+
+def processes_stats_snapshot_to_dict(stats, include_lists=None):
+  """ Converts an instance of ProcessesStatsSnapshot to dictionary. Optionally
+  it can
```
tmarballi: Please complete the docstring.

Hermes/appscale/hermes/stats/producers/proxy_stats.py
```diff
@@ -0,0 +1,516 @@
...
+import csv
+import logging
+import socket
+import time
+from collections import defaultdict
+from datetime import datetime
+
+import attr
+
+from appscale.hermes.stats.constants import HAPROXY_STATS_SOCKET_PATH, \
```
tmarballi: Please alphabetize imports here.

Hermes/appscale/hermes/stats/producers/proxy_stats.py
```diff
@@ -0,0 +1,516 @@
...
+    proxies_stats=[
+      proxy_stats_from_dict(proxy_stats, strict)
+      for proxy_stats in dictionary.get('proxies_stats', [])
+    ]
+  )
+
+
+def proxies_stats_snapshot_to_dict(stats, include_lists=None):
+  """ Converts an instance of ProxiesStatsSnapshot to dictionary. Optionally
+  it can
```
tmarballi: Please complete the docstring.

Hermes/appscale/hermes/stats/producers/tests/test_cluster_stats.py
```diff
@@ -0,0 +1,425 @@
+import json
+import os
+
+from mock import patch, MagicMock
+from tornado import testing, gen
+
+from appscale.hermes.stats.producers import (
```
tmarballi: Please alphabetize imports here.

Hermes/appscale/hermes/stats/producers/tests/test_node.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+
+from mock import patch
+
+from appscale.hermes.stats.producers import node_stats
```
tmarballi: Please alphabetize imports here.

Hermes/appscale/hermes/stats/producers/tests/test_process.py
```diff
@@ -0,0 +1,125 @@
+import os
+import unittest
+
+from mock import patch, call
+
+from appscale.hermes.stats.unified_service_names import ServicesEnum
```
tmarballi: Please alphabetize imports here.

Hermes/appscale/hermes/stats/producers/tests/test_proxy.py
```diff
@@ -0,0 +1,142 @@
+import os
+from os import path
+import unittest
+
+from mock import patch, MagicMock
+
+from appscale.hermes.stats.constants import MISSED
```
tmarballi: Please alphabetize imports here.

common/appscale/common/appscale_info.py
```diff
@@ -49,6 +49,11 @@ def get_all_ips():
   nodes = nodes.split('\n')
   return filter(None, nodes)
 
+def get_load_balancer_ips():
+  raw_ips = file_io.read(constants.LOAD_BALANCER_IPS_LOC)
```
tmarballi: Add a doc string for this method.

Hermes/setup.py
```diff
@@ -0,0 +1,35 @@
+from setuptools import setup
+
+setup(
+  name='appscale-hermes',
+  version='0.0.2',
+  description='AppScale module which takes care about periodical backup and '
```
tmarballi: Please change "takes care about" => "takes care of".

Hermes/appscale/hermes/test/unit/test_hermes.py
```diff
@@ -1,32 +1,29 @@
...
+import tornado.httpclient
 from appscale.common import appscale_info
 from appscale.common.ua_client import UAClient
 from appscale.common.ua_client import UAException
-
-sys.path.append(os.path.join(os.path.dirname(__file__), "../../"))
-import helper
-import hermes
-
+from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
```
tmarballi: Please alphabetize imports here as well.

Hermes/appscale/hermes/test/unit/test_helper.py
```diff
@@ -1,21 +1,19 @@
...
 import Queue
 import sys
 import threading
-import tornado.httpclient
 import unittest
-from flexmock import flexmock
 
+import tornado.httpclient
 from appscale.common import appscale_info
+from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
```
tmarballi: Please alphabetize imports here.

Hermes/appscale/hermes/helper.py
```diff
@@ -222,11 +224,11 @@ def create_br_json_data(role, type, bucket_name, index, storage):
   if role == 'db_master':
     data[JSONTags.TYPE] = 'cassandra_{0}'.format(type)
     data[JSONTags.OBJECT_NAME] = "gs://{0}{1}".format(bucket_name,
-      hermes_constants.DB_MASTER_OBJECT_NAME)
+                                                      constants.DB_MASTER_OBJECT_NAME)
```
tmarballi: Could you indent the format line to the next line so that its within the 80 limit?

Hermes/appscale/hermes/helper.py
```diff
@@ -222,11 +224,11 @@ def create_br_json_data(role, type, bucket_name, index, storage):
...
   elif role == 'db_slave':
     data[JSONTags.TYPE] = 'cassandra_{0}'.format(type)
     data[JSONTags.OBJECT_NAME] = "gs://{0}{1}".format(bucket_name,
-      hermes_constants.DB_SLAVE_OBJECT_NAME.format(index))
+                                                      constants.DB_SLAVE_OBJECT_NAME.format(index))
```
tmarballi: Same here.

Hermes/appscale/hermes/stats/unified_service_names.py
```diff
@@ -0,0 +1,165 @@
...
+  )
+  APPLICATION = Service(
+    name='application',
+    monit_matcher=r'^app___(?P<app>[\w_-]+)-(?P<port>\d+)$',
+    haproxy_proxy_matcher=r'^gae_(?P<app>[\w_-]+)$',
+    haproxy_server_matcher=r'^gae_(?P<app>[\w_-]+)-(?P<ip>[\d.]+):(?P<port>\d+)$'
+  )
+
+  # Known only on Monit side
+  ZOOKEEPER = Service(name='zookeeper', monit_matcher='^zookeeper$')
```
tmarballi: Are these Monit service names already existing constants somewhere that we can make use of. If not, do you see it useful to make them constants so that we can have them in one single place?

Hermes/appscale/hermes/__init__.py
```diff
@@ -12,30 +10,30 @@
...
-
+from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
 from tornado.ioloop import IOLoop
 from tornado.ioloop import PeriodicCallback
-from tornado.options import define
 from tornado.options import options
-from tornado.options import parse_command_line
 
-sys.path.append(os.path.join(os.path.dirname(__file__), '../AppServer'))
+from appscale.hermes import constants
```
jeanleonov: @tmarballi , I guess you didn't refresh pull request page before reviewing. I did some changes to address Chris's comments and imports was changed too.
I could miss some module, but generally I followed practice of splitting imports into 3 sections, each section is alphabetically ordered. Build-in imports go first, next are third party (modules which don't belong to this project, so appscale.common is here), after that project imports are (appscale.hermes ...).
Import from google.appengine.api goes after other imports because it uses manual python path.

Hermes/appscale/hermes/handlers.py
```diff
@@ -1,41 +1,39 @@
...
 from tornado.ioloop import IOLoop
 from tornado.web import RequestHandler
 
-import hermes_constants
-import helper
-from helper import JSONTags
-from helper import NodeInfoTags
-from helper import TASK_STATUS
-from helper import TASK_STATUS_LOCK
+from appscale.hermes import constants
```
jeanleonov: The same: https://www.python.org/dev/peps/pep-0008/#imports

Hermes/appscale/hermes/stats/constants.py
jeanleonov: Yes! Thanks for catching

Hermes/appscale/hermes/stats/constants.py
jeanleonov: You're right, sentence seems interrupted.

Hermes/appscale/hermes/stats/unified_service_names.py
jeanleonov: I needed a list of all known services. It's used when processes stats and proxies stats producers tries to identify unified service name by monit or proxy name. Unified services name makes possible joining processes stats with proxies stats.
I put all known services into named enum because it's useful for testing and it doesn't add a lot of code (a couple of lines).

We can remove services which have the same monit and proxy name, but I don't think it worth to do.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3218/

Hermes/appscale/hermes/stats/pubsub_base.py
jeanleonov: I don't feel strongly about it, - your arguments make sense. But how do you suggest to change this?
I have object which can produce stats and I known that produced stats should be put into cache and sent to portal and profile log.
Do you think it's better to remove subscribers list from stats publisher and replace it with 3 fixed attributes: cache, profile_log, and portal? Then we may want to replace single StatsPublisher class with two: LocalStatsUpdater (updates only local cache) and ClusterStatsUpdater (updates local cache and sends collected stats to portal and writes it to profile log).

Yeah.. sounds fine. But I'm not sure if it worth to do. @cdonati , would you suggest to reimplement it this way?

Hermes/appscale/hermes/__init__.py
tmarballi: Yeah that makes sense @jeanleonov and I agree with having some standards in place for ordering of imports. My comment was basically towards what we have been following so far and we can change that to this new approach by making a conscious effort to fix the ordering in whichever class we work on.

Hermes/appscale/hermes/stats/unified_service_names.py
tmarballi: Yeah makes sense to me. We need to make sure to re-use this Enum whenever we need it in other places too.

Hermes/appscale/hermes/stats/pubsub_base.py
cdonati: You're right. It's probably not worth taking the time to rewrite with different terms at this time. If the terminology proves confusing in the future, we can change it then.

Hermes/appscale/hermes/stats/handlers.py
jeanleonov: Thanks, I didn't notice that. And also it makes sense to change 401 (Not authorized) to 403 (Denied).

Hermes/setup.py
```diff
@@ -0,0 +1,35 @@
+from setuptools import setup
+
+setup(
+  name='appscale-hermes',
+  version='0.0.2',
+  description='AppScale module which takes care of periodical backup and '
+              'restore tasks, collects statistics and can send it outside.',
```
jeanleonov: Agree, the description is to ambitious=)

Hermes/appscale/hermes/stats/stats_app.py
```diff
@@ -0,0 +1,400 @@
...
+      write_profile: a boolean indicating whether CSV log should be written
+      verbose_cluster_stats: a boolean indicating whether all available stats
+              should be collected on master node (otherwise master will
+              select only specific important fields)
+    """
+    my_ip = appscale_info.get_private_ip()
+    lb_ips = appscale_info.get_load_balancer_ips()
+
+    self._is_lb = my_ip in lb_ips
+    if is_master is not None:
```
jeanleonov: I didn't want to check if this node is master in two places. So this check is done only in main module.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3223/

PR #2427, opened 2017-05-02
Start AppScale services in background

cdonati:
Newer versions of Monit wait for the start_cmd to exit before checking if the process is running. This PR ensures that all of the services that the AppController starts get started in the background.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3171/

PR #2429, opened 2017-05-03
Move some AppController work to AdminServer

cdonati:
This moves app hoster identification and source extraction work to the AdminServer.

### DIFF
--- a/AdminServer/appscale/admin/__init__.py
+++ b/AdminServer/appscale/admin/__init__.py
@@ -7,11 +7,16 @@ import sys
 import time
 
 from appscale.common import appscale_info
-from appscale.common.constants import HTTPCodes
-from appscale.common.constants import LOG_FORMAT
+from appscale.common.constants import (
+  HTTPCodes,
+  LOG_FORMAT,
+  ZK_PERSISTENT_RECONNECTS
+)
 from appscale.common.ua_client import UAClient
 from appscale.common.ua_client import UAException
 from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
+from kazoo.client import KazooClient
+from kazoo.exceptions import NodeExistsError
 from tornado import gen
 from tornado.options import options
 from tornado import web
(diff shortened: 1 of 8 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3175/

AdminServer/appscale/admin/__init__.py
```diff
@@ -168,15 +173,17 @@ def write_error(self, status_code, **kwargs):
 
 class VersionsHandler(BaseHandler):
   """ Manages service versions. """
-  def initialize(self, acc, ua_client):
+  def initialize(self, acc, ua_client, zk_client):
     """ Defines an AppControllerClient and UAClient.
```
whoarethebritons: and KazooClient

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3207/

AdminServer/appscale/admin/__init__.py
```diff
@@ -310,6 +317,28 @@ def begin_deploy(self, project_id, source_path):
...
+      self.zk_client.create(hoster_node, str(source_location), ephemeral=True,
+                            makepath=True)
+    except NodeExistsError:
+      self.zk_client.set(hoster_node, str(source_location))
+
+    # Remove other hosters that have old code.
+    hosters = self.zk_client.get_children('/apps/{}'.format(project_id))
+    old_hosters = [hoster for hoster in hosters if hoster != private_ip]
+    for hoster in old_hosters:
+      self.zk_client.delete('/apps/{}/{}'.format(project_id, hoster))
```
jeanleonov: I'd prefer to avoid such multiple calls to zookeeper:
```
private_ip = appscale_info.get_private_ip()
hoster_node = '/apps/{}/{}'.format(project_id, private_ip)
self.zk_client.delete('/apps/{}'.format(project_id), recursive=True)
try:
   self.zk_client.create(hoster_node, str(source_location), ephemeral=True,
                          makepath=True)
 except NodeExistsError:
   self.zk_client.set(hoster_node, str(source_location))
```
I'm not sure if this should be done in transaction and if try/except is needed here. My point is that in large deployments sequential deletion of old hosters may take some time.

cdonati: @jeanleonov Unfortunately, Kazoo's recursive delete implementation is not a single call. It makes a delete call for every child.

PR #2430, opened 2017-05-03
Downscale instances not running any AppServers

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3179/

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -5743,6 +5743,7 @@ HOSTS
           @last_decision[app_name] = Time.now.to_i
           Djinn.log_info("Removing an AppServer for #{app_name} #{location}.")
           num_to_remove -= 1
+          examine_scaling_down_instances(app_name, node_ip)
           return true if num_to_remove == 0
         end
       }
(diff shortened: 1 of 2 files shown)


AppController/djinn.rb
```diff
@@ -5743,6 +5743,7 @@ def try_to_scale_down(app_name, delta_appservers)
           @last_decision[app_name] = Time.now.to_i
           Djinn.log_info("Removing an AppServer for #{app_name} #{location}.")
           num_to_remove -= 1
+          examine_scaling_down_instances(app_name, node_ip)
```
cdonati: Could this stop machines that are running services other than AppServers?

AppController/djinn.rb
```diff
@@ -5944,71 +5945,55 @@ def get_request_info(app_id, secret)
...
+  def examine_scaling_down_instances(app_name, node_ip)
+    # Only scale down in cloud environments.
     unless is_cloud?
       Djinn.log_debug("Not scaling down VMs, because we aren't in a cloud.")
       return 0
     end
 
+    # Only scale down, if we have more machines than the user specified 
+    # minimum number of machines.
     if @nodes.length <= Integer(@options['min_images']) or @nodes.length <= 1
```
cdonati: If the user specifies ips_layout instead of min/max, what gets defined for min_images?

AppController/djinn.rb
tmarballi: No it shouldn't, so above in the downscaling of AppServers (https://github.com/AppScale/appscale/blob/master/AppController/djinn.rb#L5738) it gets just a list of appengine nodes and goes through them in reverse order. The scaling down of instances, grabs the same node ip for which the last AppServer was removed and checks if there are any more running.

cdonati: Some of the appengine nodes from get_all_appengine_nodes might have other roles, though.

tmarballi: I see what you mean now, and I do need to put some more checks around it to make sure it does not terminate instances which weren't autoscaled.

tmarballi:
I need to move this downscaling instances logic to outside of the scaling of appservers. So I will close this PR and start a new one, once that is ready so that its cleaner to review.

PR #2433, opened 2017-05-05
Fix scaling down of instances

tmarballi:
This PR allows downscaling of instances to be evaluated every duty cycle, where we go through a list of added appengine nodes (in reverse), and terminate only autoscaled appengine nodes which are not currently running any AppServers.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -5482,7 +5482,7 @@ HOSTS
       @apps_loaded.each { |app_name|
         @app_info_map[app_name]['appengine'].each { |location|
           host, port = location.split(":")
-          if host == node_ip
+          if host == node.private_ip
             hosted_apps << "#{app_name}:#{port}"
           end
         }
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -5378,6 +5378,7 @@ HOSTS
       scale_up_instances(needed_appservers)
       return
     end
+    scale_down_instances
   end
 
   # Adds additional nodes to the deployment, depending on the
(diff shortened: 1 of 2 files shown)


tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3191/
The Autoscale test failed due to a timeout and I am looking into it.

AppController/djinn.rb
```diff
@@ -5427,6 +5440,111 @@ def scale_deployment
...
 
+  # Removes autoscaled nodes from the deployment as long as they are not running
+  # any AppServers and the minimum number of user specified machines are still
+  # running in the deployment.
+  def scale_down_instances
+    num_scaled_down = 0
+    # If we are already at the minimum number of machines that the user specified,
+    # then we do not have the capacity to scale down. 
+    max_scale_down_capacity = @nodes.length - Integer(@options['min_images'])
+    if max_scale_down_capacity == 0
```
obino: I would use <= to ensure we don't go negative.

AppController/djinn.rb
```diff
@@ -5427,6 +5440,111 @@ def scale_deployment
...
+        SCALE_TIME_MULTIPLIER * DUTY_CYCLE)
+      Djinn.log_info("Not scaling down right now, as we recently scaled " +
+        "up or down.")
+      return
+    end
+
+    get_all_appengine_nodes.reverse_each { |node_ip|
+      # If we have already scaled down to our maximum capacity
+      # then return.
+      if num_scaled_down == max_scale_down_capacity
```
obino: I would use the format return if num_scaled_down ...

AppController/djinn.rb
```diff
@@ -5427,6 +5440,111 @@ def scale_deployment
...
+        return
```
obino: I would use a break here instead of a return to ensure the timestamp is updated

AppController/djinn.rb
```diff
@@ -5427,6 +5440,111 @@ def scale_deployment
...
+        "running: #{hosted_apps}")
+        next
+      end
+
+      # Right now, only the autoscaled machines are started with just the 
+      # appengine role, so we check specifically for that during downscaling
+      # to make sure we only downscale the new machines added.
+      node_to_remove = nil
+      @state_change_lock.synchronize {
+        @nodes.each{ |node|
```
obino: Another check that we can add is the index number, since the min_images is set at the beginning (at least it should be set to be the number of nodes in the original layout in the AppScalefile) any nodes after @nodes[min_images + x] is been added afterward.

AppController/djinn.rb
```diff
@@ -5427,6 +5440,111 @@ def scale_deployment
...
+          if node.private_ip == node_ip && node.jobs == ['appengine']
+            Djinn.log_info("Removing node #{node}")
+            node_to_remove = node
+            break
+          end
+        }
+      }
+      terminate_node_from_deployment(node_to_remove)
```
obino: Should we check if node_to_remove is nil?

tmarballi: Yes, I do that check within the terminate_node_from_deployment method.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3192/

AppController/lib/infrastructure_manager_client.rb
```diff
@@ -167,6 +167,7 @@ def terminate_instances(options, instance_ids)
     end
     parameters['instance_ids'] = instance_ids
     parameters['region'] = options['region']
+    parameters['IS_VERBOSE'] = options['verbose']
```
jeanleonov: Why does this parameter have different case?

tmarballi: So, the tools agents somehow set the verbose parameter in uppercase. I realized that while testing this PR the terminate call is missing this parameter. But I intend to fix the upper case parameter in a separate tools PR if that sounds good?

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/3198/

AppController/djinn.rb
```diff
@@ -2834,6 +2834,17 @@ def get_all_appengine_nodes()
...
   end
 
+  # Gets a list of autoscaled nodes by going through the nodes array
+  # and splitting the array from index greater than the 
+  # minimum images specified.
+  def get_autoscaled_nodes()
+    autoscaled_nodes = []
+    min_images = Integer(@options['min_images'])
+    @state_change_lock.synchronize {
+      autoscaled_nodes = @nodes.drop(min_images)
```
whoarethebritons: Should this function have the check for 'appengine' only nodes instead of scale down? Just in case nodes could get reordered?

AppController/djinn.rb
```diff
@@ -5427,6 +5451,107 @@ def scale_deployment
...
+
+    imc = InfrastructureManagerClient.new(@@secret)
+    begin
+      imc.terminate_instances(@options, node_to_remove.instance_id)
+    rescue FailedNodeException
+      Djinn.log_warn("Failed to call terminate_instances")
+      return 0
+    end
+
+    @last_scaling_time = Time.now.to_i
```
whoarethebritons: Should this be locked under SCALE_LOCK?

AppController/djinn.rb
tmarballi: So we believe that the @nodes array does maintain the order of the added nodes because we do some additional operations which depends on the order. But yes, we should have a better indicator of whether or not the node is autoscaled and I can work on it in a separate PR.

AppController/djinn.rb
tmarballi: Yes, good point. I will put it in a lock and a thread because the terminate can take a long enough time.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3199/

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/3200/ and the Hawkeye failure is due to a timeout.

AppController/lib/infrastructure_manager_client.rb
jeanleonov: Yep, that fine

PR #2434, opened 2017-05-09
Fix images stub to work with newer pillow

cdonati:
In newer versions of pillow, rotating by 90 degrees does not change the dimensions of the image. However, the dimensions do change in App Engine.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3202/

PR #2436, opened 2017-05-10
Use control file to install ZooKeeper and Kazoo

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3206/

PR #2437, opened 2017-05-11
Add support for Xenial

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3215/

AppServer/google/appengine/api/urlfetch_stub.py
```diff
@@ -376,6 +379,9 @@ def _RetrieveURL(url, payload, method, headers, request, response,
                 ca_certs=CERT_PATH)
           else:
             connection_class = httplib.HTTPSConnection
+            # Disable native certificate verification.
+            if hasattr(ssl, '_create_unverified_context'):
+              use_unverified_context = True
```
obino: Is this a change that happened in xenial?

debian/appscale_install_functions.sh
```diff
@@ -15,12 +15,17 @@ if [ -z "${APPSCALE_PACKAGE_MIRROR-}" ]; then
     export APPSCALE_PACKAGE_MIRROR=http://s3.amazonaws.com/appscale-build
 fi
 
+JAVA_VERSION="java-8-openjdk"
```
obino: Does that mean that we support Java 1.8?

kerbymart:
@cdonati when will this be available for `wget -O bootstrap.sh http://bootstrap.appscale.com && bash ./bootstrap.sh`?

AppServer/google/appengine/api/urlfetch_stub.py
cdonati: Python 2.7.9 and above validates certificates by default. This disables that behavior in cases when validation is not requested. We do something similar with our soap clients, but it's a little bit cleaner here since it doesn't affect other connections.

debian/appscale_install_functions.sh
cdonati: OpenJDK 7 is not available in Xenial's repos, but the Java 8 runtime will run Java 7 apps. I haven't tested if apps that take advantage of Java 8 features work.

cdonati:
@kerbymart The bootstrap script picks the latest release, and the next release will likely be within the next couple of months.

If you want to test it out before then, you can clone the repo, check out my branch, and run `bash debian/appscale_build.sh`. You'll also want to clone appscale-tools and build that in the same way.

fastercheaper:
Tried this PR and tried to build on Ubuntu Xenial but build could not automatically install the libcapnp dependency (errors out). Had to manually install libcapnp-dev as well as pycapnp to get the build to work:


```
# apt-get install libcapnp-dev
# apt-get install pycapnp

# bash appscale_build.sh
....
AppScale installation completed successfully!
# 
```

FYI OS info:
```
# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.1 LTS
Release:        16.04
Codename:       xenial

```

cdonati:
@fastercheaper Interesting. I haven't been able to reproduce the failure to build libcapnp, yet. What's the output of `uname -a`? Do you still have the error output from the build failure?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3306/
https://ocd.appscale.com:8080/job/Hawkeye-Advanced-8/342/

fastercheaper:
Just seeing the comment.. sorry can't run uname as I wiped out the entire image after running into a new problem during startup https://groups.google.com/forum/#!topic/appscale_community/1BV0NC2-vH0

PR #2438, opened 2017-05-15
Node stats profile log

jeanleonov:
The first part of stats profile log is Node stats profile log.
Major part of this PR is an update of way how stats entities are rendered to dict/list and from dict.
IncludeLists class was introduced - it's validated and prepared for a fast use entity which is able to answer what fields should be rendered for the specific stats entity class.


jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3234/

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3239/
Failed hawkeye test: 
```
Python 2.7 tests failed: 
    tests.taskqueue_tests.RESTPullQueueTest.runTest ... ERROR (ok was expected)
Java       tests failed: 
    tests.taskqueue_tests.DeferredTaskTest.runTest ... ERROR (ok was expected)
    tests.taskqueue_tests.PullQueueTest.runTest ... FAIL (ok was expected)
    tests.taskqueue_tests.PushQueueTest.runTest ... ERROR (ok was expected)
    tests.taskqueue_tests.TaskRetryTest.runTest ... ERROR (ok was expected)
```
It doesn't seem to be relevant to my changes.

Hermes/appscale/hermes/stats/subscribers/profile.py
```diff
@@ -1,19 +1,70 @@
...
+    for node_ip, snapshots in nodes_stats_dict.iteritems():
+      with self._prepare_file(node_ip) as csv_file:
+        writer = csv.writer(csv_file)
+        for snapshot in snapshots:
+          row = converter.stats_to_list(snapshot, self._include_lists)
+          writer.writerow(row)
+
+  def _prepare_file(self, node_ip):
+    file_name = path.join(self._directory, '{}.csv'.format(node_ip))
+    if not path.exists(file_name):
```
cdonati: I think `is_file` might be better since it's possible (though super unlikely), that it could be a directory. Also, do you think this header writing could go in init so that the check doesn't need to be performed for each receive?

Hermes/appscale/hermes/stats/subscribers/profile.py
```diff
@@ -1,19 +1,70 @@
...
+
+    Args:
+      include_lists: an instance of IncludeLists describing which fields
+          of node stats should be written to CSV log.
+    """
+    self._header = converter.get_stats_header(
+      node_stats.NodeStatsSnapshot, include_lists)
+    self._include_lists = include_lists
+    self._directory = path.join(PROFILE_LOG_DIR, 'node')
+    if not path.exists(self._directory):
```
cdonati: EAFP. I think https://github.com/AppScale/appscale/blob/master/AdminServer/appscale/admin/utils.py#L98-L104 is the common convention.

Hermes/appscale/hermes/stats/subscribers/profile.py
jeanleonov: Good point.
But we have to perform this check for each receive, because new nodes can be added to the deployment.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3250/

Hermes/appscale/hermes/helper.py
```diff
@@ -309,6 +308,7 @@ def backup_apps(storage, bucket):
   Returns:
     True on success, False otherwise.
   """
+  from appscale.datastore.backup import backup_recovery_helper as BR
```
tmarballi: I don't know enough about this, but is importing within a function recommended?

tmarballi: I don't know enough about this, but is importing within a function recommended?

jeanleonov: [PEP8#imports](https://www.python.org/dev/peps/pep-0008/#imports) says:

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

But in this case we always save ~10MB of resident memory for all slave nodes and in most cases we save the same amount of memory on master node.
[Discussion on stackoverflow](http://stackoverflow.com/questions/128478/should-python-import-statements-always-be-at-the-top-of-a-module) with a high rating considers such solution as acceptable.

PR #2440, opened 2017-05-16
Handle version deletes with the AdminServer

cdonati:
This moves the undeploy work from the tools to the AdminServer. Accompanies https://github.com/AppScale/appscale-tools/pull/619

### DIFF
--- a/AdminServer/appscale/admin/__init__.py
+++ b/AdminServer/appscale/admin/__init__.py
@@ -451,15 +451,9 @@ class VersionHandler(BaseHandler):
                             message='Unable to fetch version info')
 
     try:
-      project_info = app_info_map[project_id]
+      http_port = app_info_map[project_id]['nginx']
     except KeyError:
       raise CustomHTTPError(HTTPCodes.NOT_FOUND,
-                            message='Specified version does not exist')
-
-    try:
-      http_port = project_info['nginx']
-    except KeyError:
-      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
                             message='Version serving port not found')
 
     try:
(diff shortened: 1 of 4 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3232/

AdminServer/appscale/admin/__init__.py
```diff
@@ -388,6 +417,66 @@ def post(self, project_id, service_id):
...
+      app_info_map = self.acc.get_app_info_map()
+    except AppControllerException:
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
+                            message='Unable to fetch version info')
+
+    try:
+      project_info = app_info_map[project_id]
+    except KeyError:
+      raise CustomHTTPError(HTTPCodes.NOT_FOUND,
+                            message='Specified version does not exist')
```
whoarethebritons: Are version_id and project_id the same?

AdminServer/appscale/admin/operation.py
```diff
@@ -8,8 +8,9 @@
 from .constants import Types
 
 
-class CreateVersionOperation(object):
+class Operation(object):
   """ A container that keeps track of deployment operations. """
```
obino: Should the docstring be changed to reflect that this is a parent class/interface?

AdminServer/appscale/admin/__init__.py
cdonati: No, app_info_map doesn't have version information, yet. I simplified it a bit since I don't think there is a need to differentiate between the two KeyErrors.

PR #2441, opened 2017-05-17
Rabbitmq

obino:
Don't use detach mode, and control the networking of the nodes outside monit.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3237/

PR #2450, opened 2017-05-24
Relocate versions without restarting AppServers

cdonati:
This allows relocates to complete without downtime. It also simplifies the administrative work required for a relocate.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -778,10 +778,6 @@ class Djinn
     # Notify nodes, and remove any running AppServer of the application.
     notify_restart_app_to_nodes([appid])
 
-    # Once we've relocated the app, we need to tell the XMPPReceiver about the
-    # app's new location.
-    MonitInterface.restart("xmpp-#{appid}")
-
     return "OK"
   end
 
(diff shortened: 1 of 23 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3277/

cdonati:
For a demo, see https://github.com/AppScale/appscale-testing/pull/231
Here's appscale-testing/seamless-relocate on appscale/master: https://ocd.appscale.com:8080/job/Tools-test/1917/
And on this branch: https://ocd.appscale.com:8080/job/Tools-test/1918/

cdonati:
@jeanleonov Yes, there is a performance cost for the filesystem reads.

For the push tasks, I think it would be appropriate to shift the burden of request routing to the Celery workers. They can maintain a ZK watch on the version node.

The other services can also use a ZK watch, but I'm not sure if we want each AppServer to have a ZK connection. I think it would make sense once we make the API Proxy (and dispatcher, for the login URL calls) a separate service, though.

This filesystem-based solution is temporary, and its main purpose is to make it easier to replace the app_info_map with version nodes.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3300/
Both hawkeyes pass after a relocate.

PR #2451, opened 2017-05-24
Change autoscale triggers to be within a desired capacity range

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3285/

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -133,6 +133,19 @@ PACKAGE_MIRROR_DOMAIN = 's3.amazonaws.com'
 PACKAGE_MIRROR_PATH = '/appscale-build'
 
 
+# The highest capacity of the deployment (in %) we handle before trying to scale up.
+MAX_CAPACITY_THRESHOLD = 90
+
+
+# The desired capacity of the deployment (in %) to achieve after scaling up or down.
+DESIRED_CAPACITY = 80.0
+
+
+# The lowest capacity of the deployment (in %) till which we wait for before 
+# trying to scale down.
+MIN_CAPACITY_THRESHOLD = 70
+
+
 # Djinn (interchangeably known as 'the AppController') automatically
 # configures and deploys all services for a single node. It relies on other
 # Djinns or the AppScale Tools to tell it what services (roles) it should
@@ -5516,7 +5529,7 @@ HOSTS
     return 0 if @options['autoscale'].downcase != "true"
 
     # We need the haproxy stats to decide upon what to do.
-    total_requests_seen, total_req_in_queue, current_sessions,
+    total_requests_seen, total_req_in_queue, current_sessions, 
       time_requests_were_seen = HAProxy.get_haproxy_stats(app_name)
 
     if time_requests_were_seen == :no_backend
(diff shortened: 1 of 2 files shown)


AppController/djinn.rb
```diff
@@ -5526,32 +5539,67 @@ def get_scaling_info_for_app(app_name)
...
-      "#{app_name} - advising that don't scale either way on this machine.")
-    return 0
+      appservers_to_scale = calculate_appservers_needed(num_appengines, current_sessions)
+      Djinn.log_debug("The deployment is below its minimum capacity threshold for " +
+        "app #{app_name} - Advising that we scale down #{appservers_to_scale.abs} AppServers " +
+        "within this machine.")
+      return appservers_to_scale
+    else
+      Djinn.log_debug("The deployment is within the desired range of capacity for " +
+        "app #{app_name} - Advising that we don't scale either way within this machine.")
```
whoarethebritons: "Advising that there is no need to scale currently" ?

tmarballi: Yeah that does sound better :)

AppController/lib/haproxy.rb
```diff
@@ -407,12 +410,15 @@ def self.get_haproxy_stats(app_name)
 
       if service_name == "BACKEND"
         total_req_in_queue = parsed_info[REQ_IN_QUEUE_INDEX].to_i
+        current_sessions = parsed_info[CURRENT_SESSIONS_INDEX].to_i
         Djinn.log_debug("#{full_app_name} #{service_name} Queued Currently " +
           "#{total_req_in_queue}")
+        Djinn.log_debug("Current sessions for #{full_app_name} per AppServer " +
```
jeanleonov: Are you sure that `scur` is a number of sessions per AppServer? I would expect that for backend items in haproxy stats this number is total number of sessions for the backend. When for server items it is number of sessions for the specific server (AppServer in this case).

AppController/djinn.rb
```diff
@@ -5526,32 +5539,67 @@ def get_scaling_info_for_app(app_name)
 
     update_request_info(app_name, total_requests_seen, time_requests_were_seen,
       total_req_in_queue)
-
-    if total_req_in_queue.zero?
+    
+    current_capacity = calculate_current_capacity(num_appengines, current_sessions)
+    if current_capacity >= MAX_CAPACITY_THRESHOLD
       if Time.now.to_i - @last_decision[app_name] < SCALEDOWN_THRESHOLD * DUTY_CYCLE
```
jeanleonov: `... < SCALEUP_THRESHOLD` I guess

tmarballi:
@jeanleonov Regarding the capacity usage bit, I can see how it tends to be a little confusing. But what we were trying to go for is the to check what capacity has the system achieved, and we have a range as to what would be the desired capacity we want the system to have. And if above the max threshold, then we assume the system needs more AppServers and autoscale it. And if capacity seems low and not doing much now then we downscale it. Is that helpful, what do you suggest? Maybe change the terminology?

AppController/lib/haproxy.rb
tmarballi: You are right, I will change the log statement. We don't really need per AppServer and make the decision based on the total current sessions. Thanks for pointing that out!

AppController/djinn.rb
tmarballi: Will change that! Thanks :)

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3294/

cdonati:
Recreating with different branch.

PR #2457, opened 2017-06-06
Xenial docker

obino:


obino:
OCD run with the fix for docker template (https://github.com/AppScale/appscale-testing/pull/236)

PR #2458, opened 2017-06-07
Install PyYAML with pip on Xenial

cdonati:
The package in Xenial's repositories uses over 30M more memory than the same version when it's installed with pip. This overhead quickly adds up among our many Python processes. On a machine with 2 CPUs, this branch saves about 500MB of memory.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3345/

PR #2460, opened 2017-06-09
Use eventlet pool for Celery workers

cdonati:
This makes the celery worker more efficient in memory usage, and it allows a higher concurrency (20 -> 1000).

This is a continuation from https://github.com/AppScale/appscale/pull/2387.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3390/

PR #2461, opened 2017-06-09
Allow fewer dashboards for small deployments

cdonati:
This saves some memory (about 120MB) on single-node deployments.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -4774,6 +4774,9 @@ HOSTS
       setup_app_dir(AppDashboard::APP_NAME, true)
     }
 
+    # Allow fewer dashboard instances for small deployments.
+    min_dashboards = [3, get_all_appengine_nodes.length].min
+
     # Assign the specific ports to it.
     APPS_LOCK.synchronize {
       if @app_info_map[AppDashboard::APP_NAME].nil?
@@ -4784,7 +4787,7 @@ HOSTS
           'appengine' => [],
           'language' => AppDashboard::APP_LANGUAGE,
           'max_memory' => DEFAULT_MEMORY,
-          'min_appengines' => 3
+          'min_appengines' => min_dashboards
         }
       end
       @app_names << AppDashboard::APP_NAME


AppController/djinn.rb
```diff
@@ -4774,6 +4774,9 @@ def prep_app_dashboard()
       setup_app_dir(AppDashboard::APP_NAME, true)
     }
 
+    # Allow fewer dashboard instances for small deployments.
+    min_dashboards = [3, get_all_appengine_nodes.length].min
```
whoarethebritons: Is it possible to change this to the calculation for how many appservers we're able to run currently minus an arbitrary number (3 would probably work) and then doing min, theoretically leaving room for at least another app to be deployed?

cdonati: If we have 100 appengine machines, I don't think we want 100 dashboards. Now that we no longer send logs there, I can't think of a use case that would require more than 1 to handle the load. The only reason we start 3 is for fault tolerance, which probably isn't a priority for someone running a single-node deployment.

whoarethebritons: right but I meant we do min(3, (available-3 or 1))

cdonati: We could, but the assignment here is permanent. Often, we have plenty of memory shortly after starting AppScale, but the available memory is reduced after running the first hawkeye test. The number of appengine machines can also change after this assignment (so neither assignment method is great), but that number is less volatile than the number of AppServers we can run at the moment.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3383/

AppController/djinn.rb
whoarethebritons: The number of appengine nodes being the determining factor doesn't make much sense to me since we wouldn't be specifically putting on dashboard on each node.

@obino any thoughts?

cdonati: In practice, we would be putting a dashboard AppServer on each node (unless we have more than 3 appengine machines): https://github.com/AppScale/appscale/blob/master/AppController/djinn.rb#L5797-L5809

PR #2462, opened 2017-06-09
Raise a better error when there are no key samples

cdonati:
When running the rebalance script, nodetool does not always return sample keys. This gives a better indication of what went wrong.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3458/

PR #2463, opened 2017-06-13
Keep process name intact

cdonati:
The last character was being removed.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3459/

PR #2464, opened 2017-06-20
Store version details in ZooKeeper

cdonati:
The new nodes, located at /appscale/`project_id`/services/`service_id`/versions/`version-id`, replace `@app_names` and part of `@app_info_map`.

In doing so, this also introduces the patch method for the version handler and the instanceClass key for version details.

### DIFF
--- a/AdminServer/appscale/admin/__init__.py
+++ b/AdminServer/appscale/admin/__init__.py
@@ -51,44 +51,6 @@ operations = OperationsCache()
 
 
 @gen.coroutine
-def wait_for_port_assignment(operation_id, deadline, acc):
-  """ Waits until port is assigned for version.
-
-  Args:
-    operation_id: A string specifying an operation ID.
-    deadline: A float containing a unix timestamp.
-    acc: An AppControllerClient.
-  Raises:
-    OperationTimeout if the deadline is exceeded.
-  """
-  try:
-    operation = operations[operation_id]
-  except KeyError:
-    raise OperationTimeout('Operation no longer in cache')
-
-  project_id = operation.project_id
-
-  while True:
-    if time.time() > deadline:
-      message = 'Deploy operation took too long.'
-      operation.set_error(message)
-      raise OperationTimeout(message)
-
-    try:
-      info_map = acc.get_app_info_map()
-    except AppControllerException as error:
-      logging.warning('Unable to fetch info map: {}'.format(error))
-      yield gen.sleep(1)
-      continue
-
-    try:
-      raise gen.Return(info_map[project_id]['nginx'])
-    except KeyError:
-      yield gen.sleep(1)
-      continue
-
-
-@gen.coroutine
 def wait_for_port_to_open(http_port, operation_id, deadline):
   """ Waits until port is open.
 
@@ -135,7 +97,7 @@ def wait_for_deploy(operation_id, acc):
   start_time = time.time()
   deadline = start_time + constants.MAX_OPERATION_TIME
 
-  http_port = yield wait_for_port_assignment(operation_id, deadline, acc)
+  http_port = operation.version['appscaleExtensions']['httpPort']
   yield wait_for_port_to_open(http_port, operation_id, deadline)
 
   url = 'http://{}:{}'.format(options.login_ip, http_port)
@@ -513,6 +475,26 @@ class VersionHandler(BaseHandler):
     self.version_update_lock = version_update_lock
     self.thread_pool = thread_pool
 
+  def get_version(self, project_id, service_id, version_id):
+    """ Fetches a version node.
+
+    Args:
+      project_id: A string specifying a project ID.
+      service_id: A string specifying a service ID.
+      version_id: A string specifying a version ID.
+    Returns:
+      A dictionary containing version details.
+    """
+    version_node = '/appscale/projects/{}/services/{}/versions/{}'.format(
+      project_id, service_id, version_id)
+
+    try:
+      version_json, _ = self.zk_client.get(version_node)
+    except NoNodeError:
+      raise CustomHTTPError(HTTPCodes.NOT_FOUND, message='Version not found')
+
+    return json.loads(version_json)
+
   def version_from_payload(self):
     """ Constructs version from payload.
 
@@ -616,14 +598,9 @@ class VersionHandler(BaseHandler):
     if version_id != constants.DEFAULT_VERSION:
       raise CustomHTTPError(HTTPCodes.BAD_REQUEST, message='Invalid version')
 
+    version = self.get_version(project_id, service_id, version_id)
     try:
-      app_info_map = self.acc.get_app_info_map()
-    except AppControllerException:
-      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
-                            message='Unable to fetch version info')
-
-    try:
-      http_port = app_info_map[project_id]['nginx']
+      http_port = version['appscaleExtensions']['httpPort']
     except KeyError:
       raise CustomHTTPError(HTTPCodes.NOT_FOUND,
                             message='Version serving port not found')
--- a/AdminServer/appscale/admin/__init__.py
+++ b/AdminServer/appscale/admin/__init__.py
@@ -487,14 +487,24 @@ class VersionsHandler(BaseHandler):
 class VersionHandler(BaseHandler):
   """ Manages particular service versions. """
 
-  def initialize(self, acc):
-    """ Defines an AppControllerClient.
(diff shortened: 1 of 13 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3417/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3419/

AdminServer/appscale/admin/__init__.py
```diff
@@ -266,6 +241,25 @@ def version_from_payload(self):
       raise CustomHTTPError(HTTPCodes.BAD_REQUEST,
                             message='Invalid version ID')
 
+    if 'basicScaling' in version or 'manualScaling' in version:
```
cdonati: Yes, manual scaling allows you to keep state in your AppServer, but it prevents you from using the autoscaler.

https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services.versions#ManualScaling

AdminServer/appscale/admin/__init__.py
```diff
@@ -266,6 +241,25 @@ def version_from_payload(self):
...
+      raise CustomHTTPError(HTTPCodes.BAD_REQUEST,
+                            message='Only automaticScaling is supported')
+
+    # Create a revision ID to differentiate between deployments of the same
+    # version.
+    version['revision'] = int(time.time() * 1000)
```
cdonati: I figured milliseconds would be unique enough to prevent collisions between versions, but I don't have anything against multiplying it by a million.

AdminServer/appscale/admin/constants.py
```diff
@@ -69,5 +70,19 @@ class Types(object):
 # The seconds to wait for redeploys.
 REDEPLOY_WAIT = 20
 
-# A list of projects that cannot be used or modified by users.
-RESERVED_PROJECTS = [DASHBOARD_APP_ID]
+# A list of projects that cannot be modified by users.
+IMMUTABLE_PROJECTS = [DASHBOARD_APP_ID]
+
+# The ZooKeeper node that prevents concurrent location assignments.
+VERSION_UPDATE_LOCK_NODE = '/appscale/version_update_lock'
```
jeanleonov: Isn't it a good idea to put version node pattern with named placeholders near this one?

AdminServer/appscale/admin/operation.py
```diff
@@ -123,3 +123,19 @@ def finish(self):
     """ Marks the operation as completed. """
     self.response = {'@type': Types.EMPTY}
     self.done = True
+
+
+class UpdateVersionOperation(Operation):
+  """ A container that keeps track of UpdateVersion operations. """
+
+  def __init__(self, project_id, service_id, version):
+    """ Creates a new CreateVersionOperation.
```
jeanleonov: new UpdateVersionOperation

AdminServer/appscale/admin/__init__.py
```diff
@@ -420,14 +458,144 @@ def post(self, project_id, service_id):
...
+    return json.loads(version_json)
+
+  def version_from_payload(self):
+    """ Constructs version from payload.
+
+    Returns:
+      A dictionary containing version details.
+    """
+    update_mask = self.get_argument('updateMask', None)
+    if update_mask is None:
```
jeanleonov: what if update_mask is an empty string? Maybe `if not update_mask` would work better?

AdminServer/appscale/admin/__init__.py
```diff
@@ -420,14 +458,144 @@ def post(self, project_id, service_id):
...
+                   'field(s): [{}]'.format(', '.join(supported_fields)))
+        raise CustomHTTPError(HTTPCodes.BAD_REQUEST, message=message)
+
+    try:
+      given_version = json_decode(self.request.body)
+    except ValueError:
+      raise CustomHTTPError(HTTPCodes.BAD_REQUEST,
+                            message='Payload must be valid JSON')
+
+    masked_version = utils.apply_mask_to_version(given_version, desired_fields)
```
jeanleonov: Shouldn't we log warning or respond with client error if given_version contains fields which will be ignored because of given mask?

AdminServer/appscale/admin/__init__.py
```diff
@@ -420,14 +458,144 @@ def post(self, project_id, service_id):
...
+
+    version = json.loads(version_json)
+    new_ports = utils.assign_ports(version, new_fields, self.zk_client)
+    version['appscaleExtensions'].update(new_ports)
+    self.zk_client.set(version_node, json.dumps(version))
+    return version
+
+  @gen.coroutine
+  def relocate_version(self, project_id, service_id, version_id, new_fields):
+    """ Assigns new ports to a version.
```
jeanleonov: It seems that this method does actually more than just relocation of version. Shouldn't it be reflected in method name and docstring?
Or maybe new_fields argument should be replaced with two arguments: http_port, https_port, to make it clear that only ports are updated.

AdminServer/appscale/admin/__init__.py
```diff
@@ -514,12 +718,21 @@ def main():
     hosts=','.join(appscale_info.get_zk_node_ips()),
     connection_retry=ZK_PERSISTENT_RECONNECTS)
   zk_client.start()
+  version_update_lock = zk_client.Lock(constants.VERSION_UPDATE_LOCK_NODE)
```
jeanleonov: Does this lock have some timeout which could help in case AdminServer acquired the lock but then crashed badly before releasing it?

common/appscale/common/ua_client.py
```diff
@@ -45,6 +45,20 @@ def add_admin_for_app(self, email, app_id):
...
+
+    Args:
+      app_id: A string specifying an application ID.
+      host: A string specifying a hostname.
+      port: An integer specifying a port.
+      https_port: An integer specifying a port.
+    """
+    response = self.server.add_instance(app_id, host, port, https_port,
+                                        self.secret)
+    if response.lower() != 'true':
```
jeanleonov: :cry:

AdminServer/setup.py
```diff
@@ -1,5 +1,17 @@
...
 
+install_requires = [
+  'appscale-common',
+  'kazoo',
+  'PyYaml',
+  'SOAPpy',
+  'tornado'
+]
+if sys.version_info < (3,):
+  install_requires.append('futures')
```
jeanleonov: Could you point me what feature of Python 3 is used?

AdminServer/appscale/admin/__init__.py
cdonati: Yes. I wasn't sure how the API behaves in this case. After testing it, it looks like GAE treats an empty string the same as not defining the argument.

AdminServer/appscale/admin/__init__.py
cdonati: Ignoring fields is the purpose of the mask, so I don't think we need a warning. The use case is that a client can keep reusing the same version object without masking it client-side for each request.

GAE does not warn or give an error when parts of the version are ignored, and strangely it does not even warn if the masked object turns out to be empty.

AdminServer/appscale/admin/__init__.py
cdonati: Yes, I think splitting the arguments up makes sense.

AdminServer/appscale/admin/__init__.py
cdonati: The ZooKeeper lock recipe creates ephemeral nodes. If the AdminServer crashes, the lock will be released automatically.

AdminServer/setup.py
cdonati: It's the ThreadPoolExecutor in concurrent.futures.

AdminServer/appscale/admin/constants.py
cdonati: Makes sense.

AppController/lib/app_dashboard.rb
```diff
@@ -65,7 +63,7 @@ def self.start(public_ip, private_ip, persistent_storage, secret)
 
     Djinn.log_debug("Done setting dashboard.")
 
-    return true
+    return app_location
```
tmarballi: Could you update the returns section in the doc string to be the app location instead of a boolean.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3429/
The single-node hawkeye failed due to a memory allocation issue.

AdminServer/appscale/admin/__init__.py
jeanleonov: Awesome!

PR #2465, opened 2017-06-20
Always include autoscale_agent parameter

cdonati:
This fixes disk attachment in GCE.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3421/

PR #2466, opened 2017-06-26
Hermes update for autoscale test

jeanleonov:
- Profiling interval become configurable in runtime.
 - Hermes writes fewer logs.
 - Fixed bug with getting UTC timestamp for profile log.


Hermes/appscale/hermes/__init__.py
```diff
@@ -225,13 +245,19 @@ def main():
     task_route = ('/do_task', TaskHandler)
 
     if args.write_nodes_log:
-      stats_app.configure_node_stats_profiling()
+      stats_app.configure_node_stats_profiling(
+        interval=int(args.nodes_log_interval)
```
cdonati: argparse should already be coercing this to an int.

Hermes/appscale/hermes/stats/producers/proxy_stats.py
```diff
@@ -229,6 +228,8 @@ class ProxyStats(object):
   backend = attr.ib(metadata={Meta.ENTITY: HAProxyBackendStats})
   servers = attr.ib(metadata={Meta.ENTITY_LIST: HAProxyServerStats})
   listeners = attr.ib(metadata={Meta.ENTITY_LIST: HAProxyListenerStats})
+  servers_count = attr.ib()  # number of servers serving this proxy
+  listeners_count = attr.ib()  # number of listeners serving this proxy
```
cdonati: What's the reason for adding these fields? It seems like redundant info.

jeanleonov: You're right, but alternative way to get these values is to count length of servers and listeners lists. So alternative way requires client to request list of servers or list of listeners. During my autoscale testing and Maxim's development of stats on tools side we realized that it's useful to know number of servers without digging into servers details.

jeanleonov:
I'm going to move Hermes configuration to zookeeper in separate PR.

Hermes/appscale/hermes/stats/producers/proxy_stats.py
cdonati: I see. Works for me.

If it's just a matter of responding differently depending on client requests, I think it would be simpler for the server to just call `len` on the list since it's a constant time operation. If it needs to be a class attribute, it could be defined in `__getattr__`. That way, whatever is using the object doesn't need to be responsible for keeping the attributes in sync.

jeanleonov: I see what you mean. It really could be modified to computed value. But I still don't think it worth to do. All stats models are `frozen`, so `attrs` doesn't allow to modify stats objects after it's created. I wouldn't like to maintain "computed" properties inside attrs class which already overrides some magic methods.

cdonati: Ah, I forgot that it was immutable. There's no issue about keeping the values in sync, then.

jeanleonov:
Codacy issue is fixed in https://github.com/AppScale/appscale/pull/2484 .

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/3503/

PR #2467, opened 2017-06-26
Don't scale versions that haven't been loaded

cdonati:
If a version node was added after `setup_appengine_application` and before `check_running_appservers`, the controller would crash.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3445/
https://ocd.appscale.com:8080/job/Migration/1609/

PR #2468, opened 2017-06-26
Upgrade Cassandra to 3.11

cdonati:
This fixes an issue that caused Cassandra to take more than 30 seconds to start in some cases. The start script contained a pre-check described in CASSANDRA-7254.

The slow start was an issue because monit would re-run the daemon script after the 30-second timeout, and two instances of Cassandra would start. In some cases, this resulted in a corrupt keyspace.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3451/

PR #2469, opened 2017-06-26
Claim ownership of source archive

cdonati:
This makes it easier to differentiate between revisions.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3454/

cdonati:
@jeanleonov My reasoning was that when the client SCPs in the archive, nothing really keeps track of it until the AdminServer renames it. In that sense, it's claiming responsibility or ownership of it from there.

However, I agree that the function name can be confusing, so I renamed it.

PR #2470, opened 2017-06-29
Prevent instance assignment from falling below min

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3474/

PR #2471, opened 2017-06-29
Ensure port files are correct during every cycle

cdonati:
This fixes a bug in relocate where appengine machines did not update their port files.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3465/

cdonati:
That would be an improvement. A watch would be even better. Do you think the delay on a relocate is okay for now, or should I close until we have a better fix?

obino:
@cdonati I think this is a better improvement for sure. It should go in, and we can have a better fix later.

PR #2472, opened 2017-06-29
Improve scaling policy

obino:
On downscaling we could have gone past the required mimimun number of
AppServers.  On upscaling we would have consider successful a partially
fulfill requests.


obino:
will refactor and resend.

PR #2474, opened 2017-06-29
Account for queued requests

obino:
The autoscaler was not accounting for queued requests.


obino:
current session has already the queued in it.

PR #2477, opened 2017-06-30
start/stop appserver till the DUTY_CYCLE is up

obino:
Before we were working on starting/stopping a single AppServer per
DUTY_CYCLE. This patch will work on multiple AppServer till our time is
up. We don't want to take it past the DUTY_CYCLE to ensure we can
re-evaluate priorities.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3478/

AppController/djinn.rb
```diff
@@ -5041,38 +5041,39 @@ def check_running_appservers
...
-            app, version_details['appscaleExtensions']['httpPort'],
-            version_details['runtime'])
-          Djinn.log_debug("add_appserver_process returned: #{ret}.")
-        elsif !to_end[0].nil?
-          Djinn.log_info("Terminate the following AppServer: #{to_end[0]}.")
-          app, port = to_end[0].split(":")
-          ret = remove_appserver_process(app, port)
-          @unaccounted.delete(to_end[0])
-          Djinn.log_debug("remove_appserver_process returned: #{ret}.")
+        # Work untill the next DUTY_CYCLE is up.
```
cdonati: until

Also, I think "starts" is clearer than "is up" since that could mean when the next duty cycle is over.

obino: /me shakes fists at english ... thanks, will fix

PR #2478, opened 2017-07-01
Only check disk usage for mountpoints in whitelist

cdonati:
psutil is not able to get disk usage information on some devices. For example, /var/lib/ureadahead/debugfs is problematic.

### DIFF
--- a/InfrastructureManager/system_manager.py
+++ b/InfrastructureManager/system_manager.py
@@ -84,9 +84,10 @@ class SystemManager():
 
     inner_disk_stats_dict = []
     for partition in psutil.disk_partitions(all=True):
-      disk_stats = psutil.disk_usage(partition.mountpoint)
       if partition.mountpoint not in MOUNTPOINT_WHITELIST:
         continue
+
+      disk_stats = psutil.disk_usage(partition.mountpoint)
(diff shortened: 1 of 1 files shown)


InfrastructureManager/system_manager.py
```diff
@@ -84,9 +84,10 @@ def get_disk_usage(self, secret):
 
     inner_disk_stats_dict = []
     for partition in psutil.disk_partitions(all=True):
-      disk_stats = psutil.disk_usage(partition.mountpoint)
       if partition.mountpoint not in MOUNTPOINT_WHITELIST:
         continue
+
+      disk_stats = psutil.disk_usage(partition.mountpoint)
```
jeanleonov: Though I absolutely agree that `if..continue` should go before `psutil.disk_usage`, I'd also suggest to remove `all=True` flag, as it tells psutil to extend list of physical devices with other patitions like memory partitions.
[psutil docs says](https://pythonhosted.org/psutil/#psutil.disk_partitions):
> If all parameter is False it tries to distinguish and return physical devices only (e.g. hard disks, cd-rom drives, USB keys) and ignore all others (e.g. memory partitions such as /dev/shm)

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3505/

PR #2479, opened 2017-07-02
Use database locations from arguments

cdonati:
Since 3.2.1, `appscale down` removes the location files.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3482/

PR #2480, opened 2017-07-02
Clean up old celery producer connections

cdonati:
Without this, each taskqueue server just increases the number of open rabbitmq connections each time it handles reloadworker calls.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3486/

PR #2481, opened 2017-07-02
Shorten the time to check for running instances

obino:
The default time, 10 seconds, was  too long for some infrastructure that are very fast to start VMs/containers.


obino:
OCD https://ocd.appscale.com:8080/job/Daily%20Build/3489/

PR #2483, opened 2017-07-03
Print node stats sporadically

obino:
This is to lessen the amount of logs on large deployment. We will print
node stats only sporadically as a debugging measure.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3498/

PR #2490, opened 2017-07-11
Stop empty ACKs during haproxy reload

obino:
This is a simple implementation so we can test the behavior.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3536

AppController/lib/haproxy.rb
```diff
@@ -112,7 +122,13 @@ def self.restart()
   end
 
   def self.reload()
+    (PORTS + (10000..10020).to_a).each { |port|
+      Djinn.log_run("iptables -I INPUT -p tcp --dport #{port} --tcp-flags SYN,PSH,ACK,FIN,RST ACK -m u32 \! --u32 '0>>22&0x3C@12>>26&0x3C@0&0x0=0:0' -j DROP")
```
jeanleonov: In a blog you referenced, dropping only SYN packages is suggested. Why do you think we need to block packages with other flags?

obino: @jeanleonov good question: I will add the reference also as a comment. The original blog post is https://www.haproxy.com/blog/truly-seamless-reloads-with-haproxy-no-more-hacks/ where they explain that  empty ACKs are good enough and with less side effects (at least for http protocol) then SYN packets.

obino:
Closing for now since we cannot reproduce the issue that this would fix.

PR #2493, opened 2017-07-13
Handle graceful instance termination with the AppManager

cdonati:
This simplifies the controller's work when terminating AppServer instances.

See individual commits for changes to Modules and JettyContainerService.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3553/

AppManager/app_manager_server.py
```diff
@@ -90,9 +93,16 @@
 PIDFILE_TEMPLATE = os.path.join('/', 'var', 'run', 'appscale',
                                 'app___{project}-{port}.pid')
 
+# The number of seconds an instance is allowed to finish serving requests after
+# it receives a shutdown signal.
+MAX_INSTANCE_RESPONSE_TIME = 600
```
cdonati: Are you referring to the constant I also added in the SDK? I want to avoid importing anything from the SDK (or importing from an appscale package to the SDK) since it makes it more difficult to separate from the rest of our codebase.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3589/

AppServer/google/appengine/tools/devappserver2/module.py
```diff
@@ -521,6 +525,7 @@ def _error_response(self, environ, start_response, status):
     start_response('%d %s' % (status, httplib.responses[status]), [])
     return []
 
+  # AppScale.
```
whoarethebritons: `AppScale: check if the instance should be shutting down before handling request`

PR #2495, opened 2017-07-14
Count the established HAProxy connections

cdonati:
This is more accurate than asking HAProxy since it loses its statistics after every reload.


obino:
👍

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3564/

PR #2497, opened 2017-07-14
Always round up when scaling

cdonati:
Previously, it was difficult to scale up when there was only one AppServer. This is because the desired appservers rounds down even for a full load (1.25 for 7 current sessions).

This change can result in scaling down 0 AppServers when below the minimum threshold (2 AppServers with 7 current sessions, for example).


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3569/

PR #2499, opened 2017-07-17
Taskqueue and Database slave nodes are no longer sent by tools

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3581/


AppController/lib/djinn_job_data.rb
```diff
@@ -117,6 +117,22 @@ def method_missing(id, *args, &block)
     super
   end
 
+  # In the process of removing roles, db_slave is no longer added to non
+  # db_master nodes. A node that is a database and is not the db_master is
+  # now considered a db_slave.
+  def is_db_slave?
+    return @jobs.include?('db_slave') ||
```
whoarethebritons: I kept the checks for the old roles to cover the upgrade case, but I'm not sure how necessary it is. Opinions?

jeanleonov:
Are these new methods used somewhere?

AppController/lib/djinn_job_data.rb
cdonati: Since older deployments still had the database role, I don't think it's necessary. There's never a case (as far as I know) where db_slave is defined and database is not.

cdonati:
@jeanleonov The rest of the codebase uses the convention `is_<rolename>` frequently. The object was handling those requests with `method_missing`.

AppController/lib/djinn_job_data.rb
whoarethebritons: Ok, I will remove it then.

PR #2501, opened 2017-07-19
Watch for deployment configuration section updates

cdonati:
Previously, updates to a node within /appscale/config would go unnoticed.

This change is needed for modules in order to make it easier to use a KazooClient with the AppManager. It's also in preparation for storing the max_memory parameter in a way that the AppManager can retrieve it.

### DIFF
--- a/common/appscale/common/deployment_config.py
+++ b/common/appscale/common/deployment_config.py
@@ -2,7 +2,6 @@ import json
 import logging
 import time
 
-from kazoo.client import KazooClient
 from kazoo.client import KazooException
 from kazoo.client import KazooState
 from kazoo.client import NoNodeError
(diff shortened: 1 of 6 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3590/

common/appscale/common/deployment_config.py
```diff
@@ -29,24 +27,56 @@ class ConfigStates(object):
...
+    self.logger = logging.getLogger(self.__class__.__name__)
+    self.section_name = section
+    self.data = {}
+
+    section_node = '/appscale/config/{}'.format(section)
+    self.watch = zk_client.DataWatch(section_node, self._update_section)
+
+  def stop(self):
+    """ Stops the DataWatch on the configuration section. """
+    self.watch._stopped = True
```
jeanleonov: I see that there is no obvious way to stop watcher using public API, but relying on non-public attribute still doesn't look good.
My alternative also relies on implementation of DataWatch, but it doesn't use non-public attributes, so I think you can take a look at this:
zk_client.remove_listener - https://github.com/python-zk/kazoo/blob/master/kazoo/client.py#L425
which is opposite to add_listener that is used in DataWatch - https://github.com/python-zk/kazoo/blob/master/kazoo/recipe/watchers.py#L150

cdonati: That's a great point about using private attributes, and it also made me realize that I was leaving around unnecessary "exists" watches. However, the vaguely-named `remove_listener` actually just removes the connection state listener and does not affect the watch.

You can't really stop a watch without closing the connection. You can only prevent it from being reset after it fires.

The reason I was using `_stopped` instead of returning `False` from the handler was because I was worried about the slim possibility of this happening:
```
[child 1 exists in ZK]
[child 2 is added]
ChildrenWatch fires: child1, child2
DataWatch (child 2) fires
DataWatch (child 2) reset
[child 2 is deleted from ZK]
DataWatch (child 2) fires (and is not reset)
[child 2 is re-added]
ChildrenWatch reset
ChildrenWatch fires: child 1, child 2
```
With the above timeline, the handler on the last ChildrenWatch fire does not see any changes, so it will not recreate the DataWatch for child 2.

With my latest commit, I addressed that by making the ChildrenWatch handlers ensure that all of the relevant DataWatches are active.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3593/
https://ocd.appscale.com:8080/job/Autoscale-test-manual-triggers/477/

PR #2505, opened 2017-07-20
Handle version deletions more gracefully

cdonati:
Previously, the AppController could crash when the version gets removed.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3601/

PR #2509, opened 2017-07-25
Add new features into AdminServer to handle requests to get information about project modules.

Maksym-Us:


AdminServer/appscale/admin/__init__.py
```diff
@@ -140,6 +140,67 @@ def wait_for_delete(operation_id, http_port):
...
+
+    url = '/appscale/projects'
+    try:
+      apps = self.zk_client.get_children(url)
+    except NoNodeError:
+      message = 'Applications not found'
+      logging.warn('{}. URL: {}'.format(message, url))
+      raise CustomHTTPError(HTTPCodes.NOT_FOUND, message=message)
+
+    self.finish(json_encode(apps))
```
cdonati: `finish` is a little unusual to see outside the context of asynchronous handlers, but I I guess it works. The standard conventions for normal handlers is just `self.write`.

AdminServer/appscale/admin/__init__.py
```diff
@@ -159,6 +220,46 @@ def initialize(self, acc, ua_client, zk_client, version_update_lock,
...
+      except NoNodeError:
+        message = 'Information has not obtained for version: {}'.format(version)
+        logging.warn(message)
+        continue
+
+      extensions = ports['appscaleExtensions']
+      ports_info['http_port'] = extensions['httpPort']
+      ports_info['https_port'] = extensions['httpsPort']
+
+    self.finish(json_encode(versions_info))
```
cdonati: The output should resemble what is described here: https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services.versions/list. For example:
```
{"versions": [{
  "name": "apps/guestbook/services/default/versions/1",
  "id": "1",
  "runtime": "python27",
  "threadsafe": true,
  "env": "standard",
  "servingStatus": "SERVING",
  "versionUrl": "http://192.168.33.10:8080/"
}]}
```
Most of what's in the node data can be used as-is, though non-relevant fields should be stripped (eg. appscaleExtensions and revision). The http port can be communicated through the versionUrl, which I think is enough for the tools. I'm open to suggestions, though, if anyone feels that it would be important to show the https port as well.

Maksym-Us: I think https port is not redundant information, e.g. if user wants to relocate version he should to specify both ports, or if he wants to update version's https port he should to have easy way to see which https port module's version is using now, doesn't it?

cdonati: Yes, I agree with your point about HTTPS port being important information.

PR #2510, opened 2017-07-31
Consider a missing process as already stopped

cdonati:
If the controller tries to stop a process that doesn't exist, keep calm and report it gone.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3662/

PR #2512, opened 2017-08-02
Define max_memory for the deployment

cdonati:
This is needed by the AppManager in order to start instances without direct input from the AppController.

### DIFF
--- a/AppController/lib/zkinterface.rb
+++ b/AppController/lib/zkinterface.rb
@@ -501,12 +501,11 @@ class ZKInterface
   end
 
 
-  # Defines a deployment-wide default for maximum instance memory.
-  def self.set_max_memory(max_memory)
-    max_memory_node = '/appscale/config/max_memory'
-    data = {'max_memory' => max_memory}
+  # Defines deployment-wide defaults for runtime parameters.
+  def self.set_runtime_params(parameters)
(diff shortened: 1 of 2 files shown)


AppController/lib/zkinterface.rb
```diff
@@ -501,6 +501,15 @@ def self.get_version_details(project_id, service_id, version_id)
   end
 
 
+  # Defines a deployment-wide default for maximum instance memory.
+  def self.set_max_memory(max_memory)
+    max_memory_node = '/appscale/config/max_memory'
+    data = {'max_memory' => max_memory}
```
jeanleonov: Isn't it better to save just integer number instead of object? `"max_memory"` is already presented in node path.

cdonati: I agree that it's redundant, but I would like to keep it consistent with the other nodes in that path. Our DeploymentConfig watch assumes that all configuration nodes are formatted as a JSON dictionary. Do you think I should rename the node to something like /appscale/config/runtime_parameters in case we add something other than max_memory?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3711/
https://ocd.appscale.com:8080/job/Autoscale-test-manual-triggers/565/
https://ocd.appscale.com:8080/job/Blobstore/2511/

PR #2513, opened 2017-08-02
Ensure lock path is a unicode string

cdonati:
Kazoo expects paths to be either ascii-limited strings or unicode strings. Entity keys can contain non-ascii characters.

### DIFF
--- a/AppDB/appscale/datastore/zkappscale/entity_lock.py
+++ b/AppDB/appscale/datastore/zkappscale/entity_lock.py
@@ -15,7 +15,7 @@ from kazoo.retry import (
 )
 
 # The ZooKeeper node that contains lock entries for an entity group.
-LOCK_PATH_TEMPLATE = '/appscale/apps/{project}/locks/{namespace}/{group}'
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3672/

AppDB/appscale/datastore/zkappscale/entity_lock.py
```diff
@@ -29,23 +29,23 @@ def zk_group_path(key):
...
+    namespace = u':default'
 
   first_element = key.path().element(0)
-  kind = first_element.type()
+  kind = first_element.type().decode('utf-8')
 
   # Differentiate between types of identifiers.
   if first_element.has_id():
-    suffix = ':{}'.format(first_element.id())
+    group = u'{}:{}'.format(kind, first_element.id())
```
jeanleonov: Are you sure that ID can contain only ascii symbols? Or why it isn't decoded?

cdonati: IDs are automatically allocated integers. Names are user-defined strings. An entity key can have either one or the other.

PR #2514, opened 2017-08-03
Don't allow DS clients to keep connections alive

cdonati:
The Java AppServer was keeping datastore connections alive. HAProxy treated these as requests in progress, which severely limited datastore capacity until the 50s client timeout passed.


cdonati:
@jeanleonov We do already have a 50s timeout defined by HAProxy. The problem is that the datastore server can't be used while the connection remains open. We could decrease the timeout, but we would still run into problems as soon as we had more AppServers than datastore servers. The AppServers could hog available slots as long as they were making at least 1 call per timeout interval, preventing other AppServers from making requests.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3786/

PR #2515, opened 2017-08-03
Reload haproxy before nginx when adding appservers

obino:
Avoid brief windows where the frontend is ready but the backend routing is not yet setup.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/3681/

PR #2517, opened 2017-08-05
TaskQueue Targets

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3696/


AppServer_Java/src/com/google/appengine/tools/development/DevAppServerImpl.java
```diff
@@ -190,6 +190,13 @@ public void run() {
     else
     {
         System.setProperty(APPLICATION_ID_PROPERTY, config.getAppId());
+        // AppScale: Set MODULE and VERSION env variables for taskqueue.
+        System.setProperty("MODULE", mainModule.getModuleName());
+        String versionId = config.getMajorVersionId();
+        if (versionId == null)
+          versionId = "v1";
```
whoarethebritons: v1 is what appscale defaults to use as the version id for now

whoarethebritons: although I suspect that with the changes that were made to the devappserver it might not be necessary to have this if statement because I think version is always set to v1... I'll test that

AppServer/google/appengine/api/taskqueue/taskqueue.py
```diff
@@ -784,11 +784,11 @@ def __resolve_hostname_and_target(self):
...
-
-
-
-
-
+    # AppScale: Set Version and Module headers to current version & module.
+    module_version_info = os.environ['CURRENT_VERSION_ID'].split(':')
+    self.__headers['Version'] = module_version_info.pop(-1).split('.')[0]
+    self.__headers['Module'] = module_version_info.pop(-1) if len(
+      module_version_info) > 1 else 'default'
```
whoarethebritons: its `module:major_version.minor_version`, I'm adding in a comment for it now.

AppTaskQueue/appscale/taskqueue/service_manager.py
```diff
@@ -0,0 +1,154 @@
...
+    """ Establishes watches for all existing versions and ports.
+
+    Args:
+      new_versions_list: A list of strings specifying existing version IDs.
+    """
+    to_remove = [service for service in self if service not in
+                 new_versions_list]
+    for version_id in to_remove:
+      logger.debug("updating: {}".format(to_remove))
+      self.watch.__stopped = True
```
cdonati: You don't want to stop the children watch when one of the children are removed. If there are multiple versions and one version is removed, this manager should still watch for new versions.

AppTaskQueue/appscale/taskqueue/distributed_tq.py
```diff
@@ -616,6 +641,22 @@ def get_task_args(self, request):
       args['max_backoff_sec'] = queue.max_backoff_seconds
       args['max_doublings'] = queue.max_doublings
 
+      # If we could not get the target from the host, try to get it from the
+      # queue config.
```
whoarethebritons: Shouldn't it be whatever target was specified as, then queue config, then current running?

AppTaskQueue/appscale/taskqueue/distributed_tq.py
```diff
@@ -632,6 +673,85 @@ def get_task_args(self, request):
...
+        the HTTP_HOST (which would contain AppScale's login ip).
+        
+    Returns:
+      A url as a string for the given target or None if target contains the
+        AppScale login ip because the Task did not specify a target. If this
+        method returns None the target will be determined by the queue or use
+        the current running version and module.
+    """
+    if not TARGET_REGEX.match(host):
+      return None
```
whoarethebritons: Unfortunately the taskqueue api side will either set Host to version.module or HTTP_HOST. if it is set to HTTP_HOST we should not raise an error because that means the task didn't specify a target.

AppTaskQueue/appscale/taskqueue/distributed_tq.py
cdonati: "All tasks added to a queue with a target will use that target, even if a different target was assigned to the task at construction time."

AppTaskQueue/appscale/taskqueue/service_manager.py
whoarethebritons: ok, so I should just remove this line? I was confused about this specific watch.

AppTaskQueue/appscale/taskqueue/service_manager.py
```diff
@@ -0,0 +1,154 @@
...
+    """ Establishes watches for all existing services.
+
+    Args:
+      new_services_list: A list of strings specifying existing service IDs.
+    """
+    to_remove = [service for service in self if service not in
+                 new_services_list]
+    for service_id in to_remove:
+      logger.debug("updating: {}".format(to_remove))
+      self[service_id].stop()
```
whoarethebritons: should I change it to self.watch.__stopped = True ?

AppTaskQueue/appscale/taskqueue/service_manager.py
```diff
@@ -0,0 +1,154 @@
...
+  def update_services(self, new_services_list):
+    """ Establishes watches for all existing services.
+
+    Args:
+      new_services_list: A list of strings specifying existing service IDs.
+    """
+    to_remove = [service for service in self if service not in
+                 new_services_list]
+    for service_id in to_remove:
+      logger.debug("updating: {}".format(to_remove))
```
whoarethebritons: I will be getting rid of them I'm just keeping them for testing.

AppTaskQueue/appscale/taskqueue/distributed_tq.py
cdonati: Okay, it sounds like we don't have a way to differentiate between a malformed target and a lack of a target. It could be an issue down the road, but it's fine for now.

AppTaskQueue/appscale/taskqueue/service_manager.py
cdonati: In order to prevent a watch from resetting after it fires, return `False` from the registered function. Here's an example: https://github.com/AppScale/appscale/pull/2520/files#diff-5cdbc79a945a409d5709a087a1270767R141

AppTaskQueue/appscale/taskqueue/service_manager.py
cdonati: Yes. Changes to the list of versions should not affect any existing watches since the version nodes themselves are not being watched.

AppTaskQueue/appscale/taskqueue/service_manager.py
```diff
@@ -0,0 +1,154 @@
+""" Keeps track of queue configuration details for producer connections. """
```
whoarethebritons: agh, missed one!

AppServer/google/appengine/api/taskqueue/taskqueue.py
```diff
@@ -864,17 +865,17 @@ def __host_from_target(target):
...
     default_hostname = app_identity.get_default_version_hostname()
     if default_hostname is None:
-
-
-
       return None
+    # AppScale: We do not support the instance parameter.
+    target_info = target.split('.')
+    if len(target_info) > 2:
+      raise InvalidTaskError(
```
whoarethebritons: I'm a little confused about this one, the target is a property of the task and is determined here by resolve_hostname_and_target. It's not technically its own header so the request object doesn't contain information about whether we have a target or not.

cdonati: Yes, you're right. When I wrote the comment, I was under the impression that the add request had a target field.

AppServer/google/appengine/api/taskqueue/taskqueue.py
jeanleonov: Thanks

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3746/

AppServer/google/net/proto/ProtocolBuffer.py
```diff
@@ -125,7 +126,15 @@ def sendCommand(self, server, url, response, follow_redirects=1,
     pb_type = str(self.__class__).split('.')[-1]
     conn.putheader("ProtocolBufferType" , pb_type)
     conn.putheader("AppData", url) # app id, user email, nick name, auth domain
-
+    try:
+      # AppScale: Set Version and Module headers to current version & module.
+      # CURRENT_VERSION_ID is formatted module:major_version.minor_version.
+      module_version_info = os.environ['CURRENT_VERSION_ID'].split(':')
+      conn.putheader("Version", module_version_info.pop(-1).split('.')[0])
```
whoarethebritons: for some reason I assumed that pop would return the first item...

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3781/

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3787/

PR #2522, opened 2017-08-09
Tunable haproxy timeout

obino:


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/3725/

AppController/djinn.rb
```diff
@@ -487,7 +487,7 @@ class Djinn
 
   # A String that is returned to callers of set_property that provide an invalid
   # instance variable name to set.
-  KEY_NOT_FOUND = "No property exists with the given name."
+  KEY_NOT_FOUND = "Invalid property name, or property value."
```
jeanleonov: This was really confusing.

obino: @jeanleonov just to make sure I understood: are you saying that the new message is more clear? or that the is more confusing :) And if you are suggesting the latter, what would you recommend?

PR #2525, opened 2017-08-11
Don't require an email address when deploying apps

cdonati:
This will make it easier to remove app metadata from the UAServer since it's stored in ZooKeeper now.

### DIFF
--- a/AppDB/appscale/datastore/scripts/ua_server.py
+++ b/AppDB/appscale/datastore/scripts/ua_server.py
@@ -405,7 +405,6 @@ def add_admin_for_app(user, app, secret):
 
 def commit_new_app(appname, language, secret):
   global db
-  global user_schema
   global app_schema
   global super_secret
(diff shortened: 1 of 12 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3743/

AppDB/appscale/datastore/scripts/ua_server.py
```diff
@@ -403,7 +403,7 @@ def add_admin_for_app(user, app, secret):
     return "Error: Unable to update the user."
 
 
-def commit_new_app(appname, user, language, secret):
+def commit_new_app(appname, language, secret):
   global db
   global user_schema
```
whoarethebritons: is this still needed?

cdonati: Good catch. No, it is not.

PR #2529, opened 2017-08-15
Remove routing for AppServer before stopping it

cdonati:
Typically, the routing has already been removed. However, there are rare cases when the instance in in the app_info_map and the AppServer is not running.

During the euca reboot test, the app_info_map ended up with two AppServers (for two different apps) assigned to the same port on the same machine. The autoscaler did not know anything was amiss because the AppServer for the first app was running. The controller on the appengine machine knew that the AppServer was not running for the second app, but it was not able to remove the routing. This should fix that issue.

### DIFF
--- a/AppServer/google/appengine/api/appcontroller_client.py
+++ b/AppServer/google/appengine/api/appcontroller_client.py
@@ -329,6 +329,19 @@ class AppControllerClient():
       app_id, appserver_ip, port, self.secret)
 
 
+  def remove_routing_for_appserver(self, version_key, ip, port):
+    """ Instructs the AppController to stop routing traffic to an AppServer.
+
+    Args:
+      version_key: A string specifying the version key.
+      ip: A string specifying the location of the AppServer instance.
+      port: An integer specifying the location of the AppServer instance.
+    """
+    return self.call(
+      self.MAX_RETRIES, self.server.remove_routing_for_appserver, version_key,
+      ip, port, self.secret)
+
+
   def add_routing_for_blob_server(self):
     """ Tells the AppController to begin routing traffic to the
         BlobServer(s).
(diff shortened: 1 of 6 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3767/
https://ocd.appscale.com:8080/job/Tools-test/2326/
https://ocd.appscale.com:8080/job/Euca-Faststart/1502/

AppServer/google/appengine/api/appcontroller_client.py
```diff
@@ -329,6 +329,19 @@ def add_routing_for_appserver(self, app_id, appserver_ip, port):
       app_id, appserver_ip, port, self.secret)
 
 
+  def remove_routing_for_appserver(self, version_key, ip, port):
+    """ Instructs the AppController to stop routing traffic to an AppServer.
+
+    Args:
+      version_key: A string specifying the version key.
+      ip: A string specifying the location of the AppServer instance.
+      port: An integer specifying the location of the AppServer instance.
```
whoarethebritons: `An integer specifying the port the AppServer instance is using` otherwise its very similar to the other docstring

AppServer/google/appengine/api/appcontroller_client.py
```diff
@@ -329,6 +329,19 @@ def add_routing_for_appserver(self, app_id, appserver_ip, port):
       app_id, appserver_ip, port, self.secret)
 
 
+  def remove_routing_for_appserver(self, version_key, ip, port):
```
obino: Do we need this to be exposed to the dashboard? From the look of it, I think it should only be an internal call between services instead of a callable function?

cdonati: This is so that the AppManager can make the call. We don't have a separate AC client for services and the dashboard.

cdonati:
https://github.com/AppScale/appscale/pull/2530 fixes the underlying issue.

PR #2534, opened 2017-08-22
Define variable needed for block

cdonati:
This fixes an AppController crash when there are instances assigned to a machine that is going to be terminated.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3803/

PR #2535, opened 2017-08-23
Only take port from pending_appservers when checking availability

cdonati:
The app engine controllers only keep track of their own pending instances. The first part of the instance key does not contain an IP address.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -5029,8 +5029,7 @@ HOSTS
   # Returns:
   def is_port_assigned(port_to_check)
     @pending_appservers.each { |appserver, _|
-      host, port = appserver.split(":")
-      next if @my_private_ip != host
+      port = appserver.split(':')[1]
       return true if port_to_check == Integer(port)
     }
     return false


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3807/

AppController/djinn.rb
```diff
@@ -5029,8 +5029,7 @@ def is_port_already_in_use(port_to_check)
   # Returns:
   def is_port_assigned(port_to_check)
     @pending_appservers.each { |appserver, _|
-      host, port = appserver.split(":")
-      next if @my_private_ip != host
```
whoarethebritons: correct me if I'm wrong but, it seems like this check is necessary because the ip isn't checked at any other point in this specific part of the process and I think we could end up trying to use a port that is in use on the current machine but not on the host.

cdonati: Each appengine controller only keeps track of their own pending instances. The keys in `@pending_appservers` are "project_id:port".

PR #2543, opened 2017-09-06
Try to resume package downloads after failures

cdonati:
This helps when there are connectivity issues when downloading from S3.


cdonati:
Exhibits retry: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/4857/

PR #2544, opened 2017-09-06
Use fake RPC when import lock is held

cdonati:
When the application is importing modules, an API call can cause a deadlock if it performs any imports in the RPC thread. See https://docs.python.org/2/library/threading.html#importing-in-threaded-code.

This is a temporary fix. It does not allow applications to make async API calls during the import process.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3878/

PR #2546, opened 2017-09-08
Use custom logger for AdminServer

cdonati:
This filters out debug-level logs from modules like Kazoo that use the root logger.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3901/

PR #2547, opened 2017-09-08
Use APP_ID as defined in properties

cdonati:
The "APPLICATION_ID" property is not always defined when setupIndexes is called. This fixes a NullPointerException when that happens.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3902/

PR #2554, opened 2017-09-18
Undeploy projects and delete services

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3921/
https://ocd.appscale.com:8080/job/Migration/1964/

AppScale/appscale-tools#648


whoarethebritons:
I have not yet run a build on eb2c1ed

AdminServer/appscale/admin/__init__.py
```diff
@@ -130,20 +132,212 @@ def wait_for_delete(operation_id, http_port):
...
+
+  @gen.coroutine
+  def delete(self, project_id):
+    """ Deletes a project.
+
+    Args:
+      project_id: A string specifying a project ID.
+    """
+    project_path = '/appscale/projects/{0}'.format(project_id)
+    self.zk_client.exists(project_path)
```
cdonati: What's the reason for this call?

AdminServer/appscale/admin/__init__.py
```diff
@@ -710,6 +880,10 @@ def main():
   app = web.Application([
     ('/v1/apps/([a-z0-9-]+)/services/([a-z0-9-]+)/versions', VersionsHandler,
      all_resources),
+    ('/v1/apps/([a-z0-9-]+)', ProjectHandler,
```
cdonati: We may want to make this a Cloud Resource Manager API call for GCP compatibility (https://cloud.google.com/resource-manager/reference/rest/v1/projects/delete) since the Admin API doesn't support deleting projects.

AppManager/app_manager_server.py
```diff
@@ -475,9 +475,12 @@ def stop_app(version_key):
   for entry in version_entries:
     yield unmonitor_and_terminate(entry)
 
-  if not remove_logrotate(project_id):
-    logging.error("Error while setting up log rotation for application: {}".
-      format(project_id))
+  try:
+    projects_manager[project_id]
```
cdonati: Minor point, but `if project_id not in projects_manager` might be easier to read since we're not really trying to perform an operation here.

AdminServer/appscale/admin/__init__.py
```diff
@@ -130,20 +132,212 @@ def wait_for_delete(operation_id, http_port):
...
+  def delete(self, project_id, service_id):
+    """ Deletes a service.
+
+    Args:
+      project_id: A string specifying a project ID.
+      service_id: A string specifying a service ID.
+    """
+    service_path = '/appscale/projects/{0}/services/{1}'.format(project_id,
+                                                                service_id)
+    self.zk_client.exists(service_path)
```
cdonati: and this one

cdonati:
@whoarethebritons Make sure the new methods have calls to the authorize function. I'm starting to think it was a poor choice on my part to design it that way since it's so easy to forget to add that.

AdminServer/appscale/admin/__init__.py
whoarethebritons: So we'd change '/v1/apps/' to '/v1/projects'? Or are you saying we'd move it completely

cdonati: It probably doesn't need to be moved at this time.

The other major difference is that the method does not return a response body. Instead of polling an operation ID, my guess is that it would be easiest to check if the project ID is still listed in a call to https://cloud.google.com/resource-manager/reference/rest/v1/projects/list.

AdminServer/appscale/admin/__init__.py
```diff
@@ -130,20 +132,212 @@ def wait_for_delete(operation_id, http_port):
...
+    try:
+      self.acc.stop_version(version_key)
+    except AppControllerException as error:
+      message = 'Error while stopping version: {}'.format(error)
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR, message=message)
+
+    raise gen.Return(http_port)
+
+
+class ProjectHandler(BaseVersionHandler):
```
jeanleonov: ProjectHandler doesn't look like a particular case of version handler. But instead it just uses some common functionality.
I'd suggest to avoid using subclassing here and move start_delete_version outside.
Your implementation looks good and seems to correspond to popular practice of reusing code. But I want to leave this remark here and hear what you think about it=)

whoarethebritons: My reason for having them extend the BaseVersionHandler is because at the root of it the ProjectHandler and ServiceHandler undeploy versions, they just loop through and undeploy all the versions of the project or the service.

AdminServer/appscale/admin/__init__.py
```diff
@@ -130,20 +132,255 @@ def wait_for_delete(operation_id, http_port):
...
+    try:
+      try:
+        self.zk_client.delete(project_path, recursive=True)
+      except NoNodeError:
+        pass
+    finally:
+      self.version_update_lock.release()
+
+  @gen.coroutine
+  def post(self):
```
cdonati: This should be a GET, and the output should be a list of Project objects. https://cloud.google.com/resource-manager/reference/rest/v1/projects#Project

AdminServer/appscale/admin/__init__.py
```diff
@@ -130,20 +132,255 @@ def wait_for_delete(operation_id, http_port):
...
+    self.authenticate()
+    try:
+      projects = self.zk_client.get_children('/appscale/projects')
+    except NoNodeError:
+      projects = []
+    logging.info(projects)
+    self.write(json.dumps(projects))
+
+  @gen.coroutine
+  def delete(self):
```
cdonati: This should be `/v1/projects/{projectId}`.

whoarethebritons: wow. I swear I was looking at that yesterday and it wasn't part of the url and I thought it was weird... this is what I get for not wearing my glasses I guess.

PR #2557, opened 2017-09-20
Monitor also the cron daemon

obino:
This is to ensure cron won't be accidentally stop, and to avoid extra
logs from monit on xenial machines.

### DIFF
--- a/debian/appscale_install_functions.sh
+++ b/debian/appscale_install_functions.sh
@@ -542,7 +542,9 @@ EOF
     sed -i 's/set daemon.*/set daemon 5/' /etc/monit/monitrc
 
     # Monitor cron.
-    if [ -e /etc/monit/conf-available/cron -a ! -e /etc/monit/conf-enabled/cron ]; then
+    if [ -e /etc/monit/conf-available/cron ] &&
+            [ -e /etc/monit/conf-enabled ] &&
+            [ ! -e /etc/monit/conf-enabled/cron ]; then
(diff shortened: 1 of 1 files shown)


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/3935/

debian/appscale_install_functions.sh
```diff
@@ -541,6 +541,11 @@ EOF
     # Check services every 5 seconds
     sed -i 's/set daemon.*/set daemon 5/' /etc/monit/monitrc
 
+    # Monitor cron.
+    if [ -e /etc/monit/conf-available/cron -a ! -e /etc/monit/conf-enabled/cron ]; then
+        ln -s /etc/monit/conf-available/cron /etc/monit/conf-enabled
```
cdonati: It looks like if the conf-enabled directory does not exist, the link will be created there (instead of in a directory).

debian/appscale_install_functions.sh
```diff
@@ -541,6 +541,11 @@ EOF
     # Check services every 5 seconds
     sed -i 's/set daemon.*/set daemon 5/' /etc/monit/monitrc
 
+    # Monitor cron.
+    if [ -e /etc/monit/conf-available/cron -a ! -e /etc/monit/conf-enabled/cron ]; then
```
cdonati: https://github.com/koalaman/shellcheck/wiki/SC2166

PR #2558, opened 2017-09-21
Stop group after terminating process

cdonati:
There are times when killing the main AppServer process does not stop all of the processes in the group.

### DIFF
--- a/AppManager/app_manager_server.py
+++ b/AppManager/app_manager_server.py
@@ -318,23 +318,6 @@ def setup_logrotate(app_name, log_size):
 
   return True
 
-@gen.coroutine
-def kill_instance(watch, instance_pid):
-  """ Stops an AppServer process.
-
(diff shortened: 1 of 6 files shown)


AppManager/app_manager_server.py
```diff
@@ -414,7 +400,8 @@ def unmonitor_and_terminate(watch):
   except OSError:
     logging.error("Error deleting {0}".format(monit_config_file))
 
-  IOLoop.current().spawn_callback(kill_instance, watch, instance_pid)
```
whoarethebritons: so then the function kill_instance isn't needed anymore right?

cdonati: Good point. No, it is not.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3948/

PR #2559, opened 2017-09-21
Ensure the correct epmd service is started

cdonati:
After a down, reboot, up on Xenial, rabbitmq starts an epmd daemon that ejabberd does not recognize. If epmd.service is started first, both rabbitmq and ejabberd can start.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3953/

cdonati:
https://ocd.appscale.com:8080/job/Hawkeye/11756/

PR #2560, opened 2017-09-22
Transaction groomer

cdonati:
This new service replaces the garbage collector thread on the datastore servers. It allows expired transactions to be cleaned up much more quickly.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3955/
https://ocd.appscale.com:8080/job/Blobstore/2673/
https://ocd.appscale.com:8080/job/Tools-test/2435/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3958/

AppDB/appscale/datastore/scripts/transaction_groomer.py
```diff
@@ -0,0 +1,362 @@
...
+  def _groom_project(self):
+    """ Runs the grooming process. """
+    index = self._coordinator.index
+    worker_count = self._coordinator.total_workers
+
+    oldest_valid_tx = yield self._fetch_and_clean(index, worker_count)
+
+    # Wait until there's a reasonable chance that some transactions have
+    # timed out.
+    next_timeout_eta = oldest_valid_tx + MAX_TX_DURATION
```
whoarethebritons: so oldest_valid_tx is a time? If so, could you make the variable name clearer, I assumed it held the transaction.

AppDB/appscale/datastore/scripts/transaction_groomer.py
```diff
@@ -0,0 +1,362 @@
...
+
+  @gen.coroutine
+  def _resolve_txid(self, tx_node, composite_indexes):
+    """ Cleans up a transaction if it has expired.
+
+    Args:
+      tx_node: A string specifying the name of the ZooKeeper node.
+      composite_indexes: A list of CompositeIndex objects.
+    Returns:
+      The transaction start time if still valid, None if invalid.
```
whoarethebritons: "None if invalid because this method will also delete it"?

AppDB/appscale/datastore/scripts/transaction_groomer.py
```diff
@@ -0,0 +1,362 @@
...
+  def _fetch_and_clean(self, worker_index, worker_count):
+    """ Cleans up expired transactions.
+
+    Args:
+      worker_index: An integer specifying this worker's index.
+      worker_count: An integer specifying the number of total workers.
+    """
+    oldest_valid_tx = time.time()
+    future = self._tornado_zk.get_children(self._txids_path)
+    children = yield future
```
whoarethebritons: just curious, why split it into two lines?

AppDB/appscale/datastore/cassandra_env/large_batch.py
```diff
@@ -214,3 +224,215 @@ def claim(self):
...
+    """
+    statement = self._get_prepared("""
+      INSERT INTO batch_status (txid_hash, applied, op_id)
+      VALUES (?, ?, ?)
+      IF NOT EXISTS
+    """)
+    bound_statement = statement.bind((txid_hash, False, op_id))
+    bound_statement.retry_policy = NO_RETRIES
+    results = yield self._tornado_cassandra.execute(bound_statement)
+    if not results[0].applied:
```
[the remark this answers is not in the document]
cdonati: A lightweight transaction insert should always have exactly one row in the result set that indicates whether or not the statement was applied. Typically, it's clearer to use the `.was_applied` method instead of `[0].applied`, but I made the mistake of naming one of the columns "applied", and that prevents the method from working as expected.

The execute method is generic, and can be used with any type of statement. If a select statement returned no rows, we would not need to raise an exception.

AppDB/appscale/datastore/cassandra_env/large_batch.py
```diff
@@ -214,3 +224,215 @@ def claim(self):
...
+
+      new_entity = result.new_value
+      if new_entity is not None:
+        new_entity = entity_pb.EntityProto(new_entity)
+
+      if new_entity is None:
+        mutations = deletions_for_entity(old_entity, composite_indexes)
+      else:
+        mutations = mutations_for_entity(new_entity, txid, old_entity,
+                                         composite_indexes)
```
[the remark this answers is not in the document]
cdonati: Yes, that's better.

AppDB/appscale/datastore/scripts/transaction_groomer.py
cdonati: Good point. That wasn't very clear.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4429/

AppDB/appscale/datastore/cassandra_env/large_batch.py
```diff
@@ -214,3 +224,214 @@ def claim(self):
...
+      WHERE txid_hash = ?
+      IF op_id = ?
+      AND applied = ?
+    """)
+    params = (new_op_id, txid_hash, old_op_id, applied_status)
+    bound_statement = statement.bind(params)
+    bound_statement.retry_policy = NO_RETRIES
+    results = yield self._tornado_cassandra.execute(bound_statement)
+    if not results[0].applied:
+      raise BatchNotOwned('Batch status changed after checking')
```
jeanleonov: Is there any other possible reason for the result to be not applied? Or BatchNowOwned is one possible cause?

AppDB/appscale/datastore/cassandra_env/large_batch.py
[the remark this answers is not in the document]
jeanleonov: Is there any other possible reason for the result to be not applied? Or BatchNowOwned is one possible cause?

AppDB/appscale/datastore/cassandra_env/large_batch.py
```diff
@@ -214,3 +224,214 @@ def claim(self):
...
+        mutations = mutations_for_entity(new_entity, txid, old_entity,
+                                         composite_indexes)
+
+      statements_and_params = self._db_access.statements_for_mutations(
+        mutations, txid)
+      for statement, params in statements_and_params:
+        futures.append(self._tornado_cassandra.execute(statement, params))
+
+    for future in futures:
+      yield future
```
jeanleonov: Isn't it better to yield all futures together?
```
yield futures
```

AppDB/appscale/datastore/cassandra_env/large_batch.py
```diff
@@ -1,13 +1,23 @@
+import sys
 import uuid
 
 from cassandra.query import ConsistencyLevel
 from cassandra.query import SimpleStatement
+from tornado import gen
 from .retry_policies import (BASIC_RETRIES,
                              NO_RETRIES)
+from .utils import deletions_for_entity
```
jeanleonov: Imho, it's harder to navigate with relative imports. I often open files without noticing its location (from search window or by 'show definition' function of pycharm) so it's harder to understand, what exactly is imported.

According to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) relative imports are __acceptable__ when absolute are __recommended__:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved 

> However, explicit relative imports are an acceptable alternative to absolute imports

This comment shouldn't block PR, and I suspect that new relative imports are added just to keep style in the file, but anyway I'd like to hear your opinion on this.

cdonati: I have no problem with absolute imports. Previously, I thought that it was rather verbose for namespace-style packages, but I see now that it can make things clearer.

AppDB/appscale/datastore/cassandra_env/large_batch.py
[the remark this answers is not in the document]
cdonati: The insert can fail due to a connection issue (in which case, the above yield will raise an exception), but if we get a response and the statement was not applied, it can only be because the row already exists (the if clause of the statement). If the row already exists, a different process must have inserted it.

AppDB/appscale/datastore/cassandra_env/large_batch.py
cdonati: Cool. I did not know you could yield a list of futures. That is a convenient syntax.

AppDB/appscale/datastore/cassandra_env/large_batch.py
cdonati: If the statement was not applied, it could only be because the op_id changed (indicating another process claimed ownership of the batch) or the applied status changed since reading it (shouldn't be possible if the op_id is the same, but would indicate that another process modified the batch).

cdonati:
@jeanleonov For some reason, github does not show my reply to your first comment, but it's visible in the comment thread that @whoarethebritons started.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4450/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4458/
https://ocd.appscale.com:8080/job/Tools-test/2858/

AppDB/appscale/datastore/scripts/transaction_groomer.py
```diff
@@ -162,6 +166,14 @@ def __init__(self, project_id, coordinator, zk_client, db_access,
     self._stop_event = AsyncEvent()
     self._stopped_event = AsyncEvent()
 
+    # Keeps track of cleanp results for each round of grooming.
```
whoarethebritons: cleanup

PR #2561, opened 2017-09-22
Close KazooClient when stopping servers

cdonati:
When trying to reconnect, the client can block the servers from shutting down.


cdonati:
@jeanleonov I share your negative feelings towards those variables. Do you know of a way we could avoid using them?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3962/

PR #2564, opened 2017-09-27
#OCD-STABILITY Wait 60s (instead of 30s) for RabbitMQ to come up

jeanleonov:
Admin server gives a minute for app to start. Starting RabbitMQ is part of this process.
I have at least one case when 30 seconds were not enough for RabbitMQ to start: https://ocd.appscale.com:8080/job/Blobstore/2681 . When 30s were exceeded, controller restarted RabbitMQ, so start process was repeated. But meanwhile AdminServer's timeout was exceeded.


cdonati:
We may also want to configure the monit start time since by default it only waits 30 seconds for processes to start.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4198

PR #2566, opened 2017-09-28
Use new format for AppScalefile

obino:


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/3988

PR #2567, opened 2017-09-29
Enable root login on GCE

cdonati:
A recent update to GCE images caused cloud-init to behave differently, adding entries to the authorized_keys file that forbid root access.


cdonati:
This build still failed, but it got past the initialization phase: https://ocd.appscale.com:8080/view/All/job/Hawkeye-GCE-Advanced-4/6/

PR #2569, opened 2017-09-29
Remove flag that is incompatible with trusty

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4016/

PR #2572, opened 2017-09-29
Wait for lock when installing packages

cdonati:
Xenial sometimes automatically upgrades packages. This waits for that process to complete before using the package manager.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -131,7 +131,33 @@ echo "Exit now (ctrl-c) if this is incorrect"
 echo
 
 sleep 5
+
+# Wait up to 30 seconds for the package lists lock to become available.
+lock_wait_start=$(date +%s)
+printed_status=false
+while fuser /var/lib/apt/lists/lock; do
+    if [ "${printed_status}" = false ]; then
+        echo "Waiting for another process to update package lists"
+        printed_status=true
+    fi
+    elapsed_time=$(($(date +%s) - $lock_wait_start))
+    if [ "${elapsed_time}" -gt 30 ]; then break; fi
+    sleep 1
+done
 apt-get update
+
+# Wait up to 2 min for the dpkg lock to become available.
+lock_wait_start=$(date +%s)
+printed_status=false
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -131,7 +131,35 @@ echo "Exit now (ctrl-c) if this is incorrect"
...
+lock_wait_start=$(date +%s)
+printed_status=false
+while fuser /var/lib/apt/lists/lock; do
+    if [ "${printed_status}" = false ]; then
+        echo "Waiting for another process to update package lists"
+        printed_status=true
+    fi
+    current_time=$(date +%s)
+    elapsed_time=$((current_time - lock_wait_start))
+    if [ "${elapsed_time}" -gt 30 ]; then break; fi
```
jeanleonov: So we leave it to `apt-get update` to fail if lock is closed?

bootstrap.sh
```diff
@@ -131,7 +131,35 @@ echo "Exit now (ctrl-c) if this is incorrect"
...
+lock_wait_start=$(date +%s)
+printed_status=false
+while fuser /var/lib/dpkg/lock; do
+    if [ "${printed_status}" = false ]; then
+        echo "Waiting for another process to update packages"
+        printed_status=true
+    fi
+    current_time=$(date +%s)
+    elapsed_time=$((current_time - lock_wait_start))
+    if [ "${elapsed_time}" -gt 120 ]; then break; fi
```
jeanleonov: So we leave it to `apt-get install ...` to fail if lock is closed?

bootstrap.sh
cdonati: Yes, we could also print a warning here, but I think the error from the package manager is clear enough. The wait is mostly for us when we are building the image. A human operator can look into what is holding the lock and rerun the build.

cdonati:
Demo:
master: https://ocd.appscale.com:8080/job/Create-GCE-Image/435/
branch: https://ocd.appscale.com:8080/job/Create-GCE-Image/437/

PR #2573, opened 2017-10-01
Stop old epmd daemon if it's running

cdonati:
If the old daemon is running, the systemd version cannot bind to the port.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3999/

PR #2574, opened 2017-10-03
Be persistent when handling zk watch update

jeanleonov:
Before this PR data watch and children watch never retried to react on zk update.
Particularly, if you run multiple deploy and undeploy actions in short period of time, there is a chance that handling undeploy operation fail and deploy operation won't be handled at all.

### DIFF
--- a/AdminServer/appscale/admin/utils.py
+++ b/AdminServer/appscale/admin/utils.py
@@ -10,10 +10,6 @@ import tarfile
 
 import random
 
-import traceback
-
-import weakref
-
 import collections
 from appscale.common.constants import HTTPCodes
 from appscale.common.constants import VERSION_PATH_SEPARATOR
@@ -21,7 +17,6 @@ from appscale.taskqueue import constants as tq_constants
 from appscale.taskqueue.constants import InvalidQueueConfiguration
 from kazoo.exceptions import NoNodeError
 from tornado import gen, locks
-from tornado.gen import BadYieldError
 
 from . import constants
 from .constants import (
--- a/AdminServer/appscale/admin/utils.py
+++ b/AdminServer/appscale/admin/utils.py
@@ -510,7 +510,7 @@ def retry_coroutine(func, async=True, backoff_base=2, backoff_multiplier=0.2,
                     retry_on_exception=None, refresh_args_kwargs_func=None):
   """ Wraps func with retry mechanism which runs up to max_retries attempts
   with exponential backoff (sleep = backoff_multiplier * backoff_base**X).
-  
+
   Args:
     func: function or coroutine to wrap.
     async: a boolean indicating if result of function should be yielded.
--- a/AdminServer/appscale/admin/utils.py
+++ b/AdminServer/appscale/admin/utils.py
@@ -523,7 +523,7 @@ def retry_coroutine(func, async=True, backoff_base=2, backoff_multiplier=0.2,
     refresh_args_kwargs_func: a function without arguments which should return
       tuple of two elements (arguments list, keyword arguments dict).
   Returns:
-    Wrapped persistent function
+    A tornado coroutine wrapping function with retry mechanism.
   """
 
   @gen.coroutine
@@ -533,6 +533,10 @@ def retry_coroutine(func, async=True, backoff_base=2, backoff_multiplier=0.2,
 
     while True:
       try:
+        if retries and refresh_args_kwargs_func is not None:
+          # Refresh arguments if applicable after the first try
+          args, kwargs = refresh_args_kwargs_func()
+
         result = func(*args, **kwargs)
 
         if async:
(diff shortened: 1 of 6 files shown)


AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +503,119 @@ def queues_from_dict(payload):
...
+    refresh_args_kwargs_func: a function without arguments which should return
+      tuple of two elements (arguments list, keyword arguments dict).
+  Returns:
+    A tornado coroutine wrapping function with retry mechanism.
+  """
+
+  @gen.coroutine
+  def persistent_execute(*args, **kwargs):
+    retries = 0
+    backoff = backoff_multiplier * (backoff_base ** 1)
```
cdonati: What's the purpose of `** 1` here?

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +503,119 @@ def queues_from_dict(payload):
...
+
+    while True:
+      try:
+        if retries and refresh_args_kwargs_func is not None:
```
cdonati: I don't think this refresh should be needed. If the watched node gets updated during the first execution of the function, the function will be triggered again when the watch is re-established. Since this function yields, though, you may want to use a tornado async lock to prevent the second trigger from being run before the first finishes.

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +503,119 @@ def queues_from_dict(payload):
...
+      except Exception as e:
+        # Decide if retries is needed
+        if max_retries is not None and retries >= max_retries:
+          raise
+        if retry_on_exception is not None and not retry_on_exception(e):
+          raise
+        retries += 1
+
+        # Proceed with exponential backoff
+        sleep_time = backoff * (random.random()*0.3 + 0.85)
```
cdonati: I'm curious why this is randomized.

AdminServer/appscale/admin/utils.py
jeanleonov: I believe it makes more clear that backoff_base is base for exponent.

AdminServer/appscale/admin/utils.py
jeanleonov: Random part (up to 15% change) helps to avoid collisions in case some service is crashed and many services start retrying requests to it. If all services use the same backoff then there is a chance that too recovered service will receive too many requests in the same time.

AdminServer/appscale/admin/utils.py
jeanleonov: I was trying to resolve following problem:
 0. AdminServer known_projects=[projA, projB, projC]
 1. children are updated, new list is [projA, projB]
 2. watch finds out that projC should be removed, but fails to do this
 3. retrying after 0.2s, 0.4s, 0.8s, 1.6s, 3.2s, next retry in 6.4s
 4. children are updated, new list is [projA, projB, projD]
 5. update is handled properly, projC is removed, projD is created. Now known_projects=[projA, projB, projD]
 6. retry num# 6 with [projA, projB] removes projD

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4007/
https://ocd.appscale.com:8080/job/Hawkeye-Simple-4/874/
https://ocd.appscale.com:8080/job/Hawkeye-Simple-8/939/
https://ocd.appscale.com:8080/job/Migration/2033/

AdminServer/appscale/admin/utils.py
cdonati: That is definitely the right situation to consider. An async lock will prevent this by serializing the triggers. One problem with this implementation is that it adds another call that can fail or block the loop. While it makes race conditions less likely, it doesn't completely resolve the situation you describe. For example,

1. Trigger 1 (T1) gets scheduled to run on IOLoop (projects A, B)
2. T1 gets the current list (A, B)
3. The children are updated (A, B, D)
4. The ZK thread schedules trigger 2 to run (A, B, D)
5. T1 tries to yield result from function 
6. Function from T1 gets scheduled to run on IOLoop with args (A, B) (F1)
7. T2 gets the current list (A, B, D)
8. T2 tries to yield result from function
9. Function from T2 gets scheduled to run with args (A, B, D) (F2)

Now it's a race between F1 and F2. If any async operation in F1 (an http call to monit, for example) takes a while, F2 could finish before F1.

AdminServer/appscale/admin/utils.py
cdonati: Cool, that makes sense.

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +503,119 @@ def queues_from_dict(payload):
...
+        backoff *= backoff_base
+        if backoff > backoff_threshold:
+          backoff = backoff_threshold
+        # Report error to logs
+        error = traceback.format_exc()
+        logger.error(u"{error}Retry #{retry} in {sleep:0.1f}s"
```
cdonati: logger.exception is a handy shortcut for this.

AdminServer/appscale/admin/utils.py
jeanleonov: I see. As you suggest i need to wrap watch functions with async lock to avoid concurrent update. But async sleep between retries must be outside lock (to be reactive to updates from zk), so i need to make sure i don't overwrite latest update. I'll try to find the best solution) thanks for raising this issue!

cdonati: Okay, it's up to you how to implement it, but I personally would be fine with the sleep inside the lock. The reactivity comes from the watch, which will ensure that the function gets called with the latest state. Retrying the function with slightly stale state is not a problem because it will get called again with fresh state as soon as ZK is available again.

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +502,105 @@ def queues_from_dict(payload):
...
+      # Wake older update calls (*)
+      node_lock.condition.notify_all()
+      node_lock.waiters += 1
+      with (yield node_lock.lock.acquire()):
+        node_lock.waiters -= 1
+
+        while True:
+          try:
+            result = func(*args, **kwargs)
+            if isinstance(result, gen.Future):
```
cdonati: nice. I like how this no longer requires an async flag.

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +502,105 @@ def queues_from_dict(payload):
...
   return {'queue': queues}
+
+
+class _PersistentWatch(object):
+
+  def __init__(self):
+
+    class CompliantLock(object):
+      """
+      A container for three pieces which makes possible to
```
tmarballi: Wording is a bit confusing. Could we change it to "A container which allows to organize compliant locking of zk_node by making sure:"

AdminServer/appscale/admin/utils.py
```diff
@@ -497,3 +502,105 @@ def queues_from_dict(payload):
...
+            # Proceed with exponential backoff
+            backoff = backoff * backoff_base
+            if backoff > backoff_threshold:
+              backoff = backoff_threshold
+            sleep_time = backoff * (random.random() * 0.3 + 0.85)
+            # Report error to logs
+            logger.exception(u"Retry #{retry} in {sleep:0.1f}s"
+                             .format(retry=retries, sleep=sleep_time))
+
+            # (*) Sleep with one eye open, give up if newer update wake you
```
tmarballi: "wake" ==> "wakes"

jeanleonov:
https://ocd.appscale.com:8080/job/Hawkeye-Simple-8/955/
https://ocd.appscale.com:8080/job/Hawkeye-Advanced-8/889/
https://ocd.appscale.com:8080/job/Blobstore/2732/
https://ocd.appscale.com:8080/job/Migration/2047/

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4020/

jeanleonov:
https://ocd.appscale.com:8080/job/Hawkeye-Simple-8/956/
https://ocd.appscale.com:8080/job/Hawkeye-Simple-8/957/
https://ocd.appscale.com:8080/job/Migration/2048/
https://ocd.appscale.com:8080/job/Autoscale-test-manual-triggers/799/

PR #2581, opened 2017-10-06
Handle missing WEB-INF directory

cdonati:
The tools should prevent the user from uploading a tarball with this issue, but it makes sense to handle it anyway.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4182/

PR #2584, opened 2017-10-13
Define property name based on rsyslog version

cdonati:
This fixes the combined app log in Xenial.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4050/

PR #2585, opened 2017-10-14
Access Tokens for Admin API

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/4052/
https://ocd.appscale.com:8080/job/Migration/2082/
https://ocd.appscale.com:8080/job/Autoscale-test-manual-triggers/827/


AdminServer/appscale/admin/__init__.py
```diff
@@ -952,6 +976,109 @@ def get(self, project_id, operation_id):
...
+    # so we create it.
+    metadata = {
+      'user': username,
+      'exp': time.time() + 3600,
+      'scope': self.AUTH_SCOPE
+    }
+
+    metadata_base64 = base64.urlsafe_b64encode(json.dumps(metadata))
+
+    method = {'type': 'JWT', 'alg': 'RS256'}
```
cdonati: I think 'SHA-1' would be more appropriate for the alg field.

AdminServer/appscale/admin/__init__.py
```diff
@@ -952,6 +976,109 @@ def get(self, project_id, operation_id):
...
+
+    username = self.get_argument('username', default=None, strip=True)
+    if not username:
+      missing_arguments.append('username')
+
+    password = self.get_argument('password', default=None, strip=True)
+    if not password:
+      missing_arguments.append('password')
+
+    if missing_arguments:
```
cdonati: The client should specify the scope when requesting the token. That way, users with different privileges can request different scopes.

whoarethebritons: scope is technically an optional argument though isn't it? and do we want to accept whichever scope they specify?

cdonati: We should only allow the single scope for now. It would be optional if we didn't anticipate requiring different scopes for different actions in the future. Without requiring a scope, we won't know what permissions the client is asking for.

Clients that depend on this functionality would need to change their behavior if we wait until later to make that a required field.

AdminServer/appscale/admin/__init__.py
```diff
@@ -270,11 +273,25 @@ def initialize(self, acc, ua_client, zk_client, version_update_lock,
...
-    self.authenticate()
-    try:
-      projects = self.zk_client.get_children('/appscale/projects')
-    except NoNodeError:
-      projects = []
+    if 'AppScale-Secret' not in self.request.headers:
+      # Run all the authenticate functions in authenticate_access_token but
+      # don't check if user is authenticated for a specific project,
+      # get projects user can access.
+      token = self.request.headers['Authorization'].split()[1]
```
cdonati: This should probably handle the case when there is no "Authorization" header. A bad request status code would be more informative to the client than a 500.

AdminServer/appscale/admin/__init__.py
```diff
@@ -270,11 +273,25 @@ def initialize(self, acc, ua_client, zk_client, version_update_lock,
...
+      method_base64, metadata_base64, signature = token.split('.')
+      self.check_token_hash(method_base64, metadata_base64, signature)
+
+      metadata = json.loads(base64.urlsafe_b64decode(metadata_base64))
+      self.check_token_expiration(metadata)
+      self.check_token_scope(metadata)
+      user = metadata['user']
+      if self.ua_client.is_user_cloud_admin(user):
```
cdonati: This should handle UAException.

AdminServer/appscale/admin/__init__.py
```diff
@@ -952,6 +984,119 @@ def get(self, project_id, operation_id):
...
+      error_msg['error'] = constants.AccessTokenErrors.INVALID_GRANT
+      error_msg['error_description'] = "Incorrect password for {}"\
+        .format(username)
+      raise CustomHTTPError(HTTPCodes.UNAUTHORIZED, message=error_msg)
+
+    # If we have gotten here, the user is granted an Access Token,
+    # so we create it.
+    metadata = {
+      'user': username,
+      'exp': time.time() + 3600,
```
cdonati: This isn't super important, but you could round to the nearest int to make the token a bit smaller. We don't need microsecond precision.

AdminServer/appscale/admin/base_handler.py
```diff
@@ -1,27 +1,165 @@
...
-    if self.request.headers['AppScale-Secret'] != options.secret:
+    if secret_header and self.request.headers['AppScale-Secret'] != \
+        options.secret:
       raise CustomHTTPError(HTTPCodes.UNAUTHORIZED, message='Invalid secret')
+    elif authorization_header:
+      try:
+        self.authenticate_access_token(
+          self.request.headers['Authorization'].split()[1],
+          project_id, ua_client)
+      except IndexError:
```
cdonati: I may have misled you on this one. I can't remember. From an API design perspective, it's a little unusual to have to handle an IndexError at this level. I would recommend raising the CustomHTTPError in the authenticate_access_token method.

AdminServer/appscale/admin/base_handler.py
```diff
@@ -1,27 +1,165 @@
...
+
+    # If user is cloud admin, return since they can access all projects.
+    if ua_client.is_user_cloud_admin(user):
+      return
+
+    projects_list = self.get_users_projects(user, ua_client)
+
+    # Unauthorized if user cannot access project.
+    if project_id not in projects_list:
+      user = metadata['user']
```
cdonati: Duplicate declaration.

AdminServer/appscale/admin/__init__.py
whoarethebritons: Bad Request or Unauthorized? In the base_handler if its missing that header or AppScale-Secret we say Unauthorized

AdminServer/appscale/admin/__init__.py
whoarethebritons: If we can't check if the user is cloud admin, should we assume they are not or should we raise an error?

AdminServer/appscale/admin/__init__.py
cdonati: You're right. Unauthorized is more specific.

AdminServer/appscale/admin/__init__.py
cdonati: Since it is an operation error and does not have to do with the request, it should be an error in the 500 range.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/4236/
https://ocd.appscale.com:8080/job/Hawkeye-Simple-8/1133/
https://ocd.appscale.com:8080/job/Hawkeye-Simple-4/1057/
https://ocd.appscale.com:8080/job/Migration/2236/

PR #2591, opened 2017-10-19
Use custom template for combined app log messages

cdonati:
This fixes the issue with outdated timestamps being used in the combined app log. The version of logger in Xenial does not generate new timestamps when sending each log line, so this template uses a generated timestamp instead of the one that logger provides.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4179/
https://ocd.appscale.com:8080/job/Migration/2185/

PR #2592, opened 2017-10-20
Do not raise HttpError outside context of HTTP request

jeanleonov:
wait_for_delete coroutine is run as a background job, it shouldn't raise HttpError - writing error to logs and raising `gen.Return() ` is enough.


jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4134/

PR #2602, opened 2017-11-02
Specify full path to appscale-stop-instance

cdonati:
In Trusty, /usr/local/bin is not on the path under monit even when starting under bash.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4464/

PR #2605, opened 2017-11-06
Simplify datastore error handling

cdonati:
If an error code is defined, the response body does not need to be specified. This prevents encoding issues in cases when a response message has required fields.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4281/

cdonati:
Previously, it was common convention at this layer to do something like:
```
response = datastore_pb.Transaction()
response.set_app(project_id)
try:
  response.set_handle(datastore.create_txid())
  return response, 0, ''
except Exception:
  return response, error_code, 'error message'
```
and the parent function always tries to set the message with `response.Encode()`, which in the above example would raise `ProtocolBufferEncodeError: Required field: handle not set.` if there was an error creating a transaction ID.

The result is that the client would receive a 500 with a stacktrace as the body instead of a properly-formatted protobuf response with an error code set.

jeanleonov:
Got it, thanks for the explanation.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4837/

PR #2607, opened 2017-11-07
Set up blobstore routing on each load balancer

cdonati:
Previously, only one load balancer would set up routing for the blob server.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4300/

PR #2609, opened 2017-11-08
Keep track of ZooKeeper state with TransactionManager

cdonati:
This manager allows projects to create transaction IDs larger than 2^31 - 1. It can also account for a txid offset, which is useful after a binary Cassandra migration when future txids need to be larger than the largest txid from the backup.

Note: Setting the offset should only be done when the datastore is in read-only mode after 60 seconds have passed. Otherwise, it can affect open transactions.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4298/

AppDB/appscale/datastore/zkappscale/transaction_manager.py
```diff
@@ -0,0 +1,329 @@
...
+
+    # Containers that do not need to be checked for open transactions.
+    self._inactive_containers = set()
+
+  def create_transaction_id(self, xg):
+    """ Generates a new transaction ID.
+
+    Args:
+      xg: A boolean indicating a cross-group transaction.
+    """
```
tmarballi: Possibly needs a "Returns" field in here.

AppDB/appscale/datastore/backup/datastore_backup.py
```diff
@@ -190,31 +190,6 @@ def get_entity_batch(self, first_key, batch_size, start_inclusive):
 
     return batch
 
-  def verify_entity(self, key, txn_id):
```
tmarballi: I don't think I understand why this is no longer needed. Do you mind explaining that?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4363/

AppDB/appscale/datastore/backup/datastore_backup.py
cdonati: It was left over from before we started writing entity data on commit. We used to write entity data before a transaction was committed and keep track of its status with the combination of a blacklist and a list of running transactions in ZooKeeper. Since 3.0, we no longer use this function.

tmarballi: Got it! Thanks for clarifying that!

AppDB/appscale/datastore/zkappscale/transaction_manager.py
```diff
@@ -0,0 +1,329 @@
...
+  def create_transaction_id(self, project_id, xg=False):
+    """ Generates a new transaction ID.
+
+    Args:
+      project_id: A string specifying a project ID.
+      xg: A boolean indicating a cross-group transaction.
+    Raises:
+      BadRequest if the project does not exist.
+      InternalError if unable to create a new transaction ID.
+    """
```
whoarethebritons: missing Returns

AppDB/appscale/datastore/zkappscale/transaction_manager.py
```diff
@@ -160,10 +161,14 @@ def _delete_counter(self, path):
       path: A string specifying a ZooKeeper path.
     """
     try:
-      self.zk_client.delete(path)
+      try:
+        self.zk_client.delete(path)
+      except NotEmptyError:
+        # Cross-group transaction nodes have a child node.
+        self.zk_client.delete(path, recursive=True)
```
whoarethebritons: why not always make recursive True?

cdonati: If we don't expect the node to have children, the delete call is more efficient. A recursive delete first makes a get_children call: https://github.com/python-zk/kazoo/blob/4d268adf9837836f05dde5ec81be0d7bbd759e78/kazoo/client.py#L1329

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4373/

PR #2615, opened 2017-11-11
Kill instances when they exceed memory limit

cdonati:
Previously, Monit would just try to teminate it gently, allowing the instance to finish serving open requests. In some cases, this limit was not long enough (30 seconds). In other cases, it was too long. It triggered the controller to try to stop the instance through the AppManger, resulting in a sequence that caused monit to restart the instance after the AppManger unmonitored it.

### DIFF
--- a/common/appscale/common/monit_app_configuration.py
+++ b/common/appscale/common/monit_app_configuration.py
@@ -18,7 +18,8 @@ MONIT_CONFIG_DIR = '/etc/monit/conf.d'
 
 
 def create_config_file(watch, start_cmd, pidfile, port=None, env_vars=None,
-                       max_memory=None, syslog_server=None, check_port=False):
+                       max_memory=None, syslog_server=None, check_port=False,
+                       kill_exceeded_memory=False):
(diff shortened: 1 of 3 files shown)


common/appscale/common/monit_app_configuration.py
```diff
@@ -31,6 +32,8 @@ def create_config_file(watch, start_cmd, pidfile, port=None, env_vars=None,
       megabytes that the process should use.
     syslog_server: The IP address of the remote syslog server to use.
     check_port: A boolean specifying that monit should check host and port.
+    kill_exceeded_memory: A boolean indicating that a process should be killed
+      (instead of terminated) if its memory exceeds the limit.
```
obino: I think the comment should be something like: ... instead of terminated). This is mainly used when the process exceed its memory limits. Reason is the memory limit is not checked in this script, but somewhere else.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4330/

PR #2616, opened 2017-11-11
Treat HAProxy 'DOWN' states the same

cdonati:
In HAProxy 1.6, a server can perpetually have the value "DOWN 1/2" despite the process being completely gone.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4339/

PR #2617, opened 2017-11-11
Set log after restart

obino:
With this, the AC will have the same log level used before it was restarted.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/4322/

PR #2618, opened 2017-11-13
Move AppControllerClient to separate package

cdonati:
This allows packages to import it directly without including AppServer code. It also makes it unavailable to applications unless they vendor it in.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4331/

obino:
Should we also have a whitelist or a way to specify which application can vendor in the appcontroller client?

cdonati:
There's not really a way we can restrict what the user decides to include with their application, and I don't think we need to. The AppControllerClient is just making SOAP calls through the urlfetch api. The restriction is the secret key, which only the dashboard should have access to in order to actually use the controller client effectively.

PR #2626, opened 2017-12-05
Use datastore proxy on load balancer

cdonati:
Previously, the groomer would try to use a datastore location that may not exist.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4438/

PR #2630, opened 2017-12-14
Add a datastore viewer to the dashboard

cdonati:
This bypasses the API server to make calls directly to the datastore server on behalf of a given project. Most of the interface was taken from the Python SDK's admin server, but calls to the datastore API were replaced with urlfetch calls to the datastore server.

### DIFF
--- a/AppDashboard/lib/datastore_viewer.py
+++ b/AppDashboard/lib/datastore_viewer.py
@@ -1,0 +1,637 @@
+""" Implements the datastore viewer interface. """
+import math
+import urllib
+
+from app_dashboard import AppDashboard
+from datastore_location import DATASTORE_LOCATION
+from google.appengine.api import datastore
+from google.appengine.api import memcache
+from google.appengine.api.datastore_distributed import DatastoreDistributed
+from google.appengine.api import api_base_pb
+from google.appengine.datastore import datastore_pb
+
+from google.appengine.tools.devappserver2.admin.datastore_viewer import (
+  _property_name_to_value,
+  DataType)
+
+
+def _format_datastore_key(key):
+  """Return a nicely formatted decomposition of a datastore key.
+
+  Args:
+    key: The datastore_types.Key object to format.
+
+  Returns:
+    A string or Unicode object containing nicely formatted information about
+    the given key e.g. "ParentKind: name=Animal > ChildKind: id=123".
+  """
+  path = key.to_path()  # [kind, id/name, kind, id/name, ...]
+  parts = []
+  for i in range(0, len(path)//2):
+    kind = path[i*2]
+    value = path[i*2 + 1]
+    if isinstance(value, (int, long)):
+      parts.append('%s: id=%d' % (kind, value))
+    else:
+      parts.append('%s: name=%s' % (kind, value))
+  return ' > '.join(parts)
+
+
+def _property_name_to_values(entities):
+  """Returns a a mapping of entity property names to a list of their values.
+
+  For example:
+    _property_name_to_values([{'cat': 5, 'dog': 10},
+                              {'dog': 15, 'mouse': 'happy'}])
+    => {'cat': [5], 'dog': [10, 15], 'mouse': ['happy']}
+
+  Args:
+    entities: A sequence of mappings (i.e. datastore.Entity) that represent
+        datastore properties and their values.
+
+  Returns:
+    A dict whose keys are the union of the keys of the given entities and
+    whose values are the list of values for those keys.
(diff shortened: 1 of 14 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4449/

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
+""" Implements the datastore viewer interface. """
+import math
+import urllib
+
+from app_dashboard import AppDashboard
+from datastore_location import DATASTORE_LOCATION
+from google.appengine.api import datastore
+from google.appengine.api import memcache
+from google.appengine.api.datastore_distributed import DatastoreDistributed
+from google.appengine.api import api_base_pb
```
whoarethebritons: alphabetize

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
...
+from google.appengine.datastore import datastore_pb
+
+from google.appengine.tools.devappserver2.admin.datastore_viewer import (
+  _property_name_to_value,
+  DataType)
```
whoarethebritons: should this parentheses be on the next line?

not hugely important, mostly curious

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
...
+    value = path[i*2 + 1]
+    if isinstance(value, (int, long)):
+      parts.append('%s: id=%d' % (kind, value))
+    else:
+      parts.append('%s: name=%s' % (kind, value))
+  return ' > '.join(parts)
+
+
+def _property_name_to_values(entities):
+  """Returns a a mapping of entity property names to a list of their values.
```
whoarethebritons: a a

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
...
+  property_name_to_values = {}
+  for entity in entities:
+    for property_name, value in entity.iteritems():
+      property_name_to_values.setdefault(property_name, []).append(value)
+
+  return property_name_to_values
+
+
+def _delete_entities(ds_access, entity_keys):
+  """ Deletes a list of datastre entities.
```
whoarethebritons: datastore

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
...
+
+    property_name_to_value = _property_name_to_value(entities)
+
+    headers = [{'name': property_name}
+               for property_name in sorted(property_name_to_value)]
+
+    template_entities = []
+    for entity in entities:
+      attributes = []
+      for property_name in sorted(property_name_to_value):
```
whoarethebritons: since this is the second sort should we store the sorted one in a variable or use the headers list and set a temp variable property_name to header['name'] ? Or since we never use the list without sorting it at the assignment it could be `property_name_to_value = sorted(_property_name_to_value(entities))` or the method could return it sorted

I just would rather we don't sort the same thing twice in this method.

AppDashboard/lib/datastore_viewer.py
```diff
@@ -0,0 +1,637 @@
...
+    request.set_value(ds_access.project_id)
+    response = datastore_pb.CompositeIndices()
+    ds_access._Dynamic_GetIndices(request, response)
+    indexes = []
+    for index in response.index_list():
+      props = [(prop.name(), prop.direction())
+               for prop in index.definition().property_list()]
+      new_index = datastore.Index(index.id(), index.definition().entity_type(),
+                                  index.definition().ancestor(), props)
+      indexes.append((new_index, None))
```
whoarethebritons: why is this a tuple if the second value is always going to be `None`?
  
Also technically doesn't this mean the docstring should say returns a list of tuples?

AppDashboard/lib/datastore_viewer.py
cdonati: Not sure. I didn't see anything in PEP 8 to indicate one way or the other.

AppDashboard/lib/datastore_viewer.py
cdonati: The second item in the tuple was indicating index state sometimes in the SDK. I just used the same signature to avoid modifying the method that used it.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4499/

PR #2631, opened 2017-12-15
Add option to update all composite indexes

cdonati:
If there are many composite indexes that need updating, selecting them one by one is not practical.

### DIFF
--- a/AppDB/appscale/datastore/scripts/update_index.py
+++ b/AppDB/appscale/datastore/scripts/update_index.py
@@ -1,4 +1,4 @@
-import getopt
+import argparse
 import sys
 
 from appscale.common import appscale_info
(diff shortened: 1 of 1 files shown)


AppDB/appscale/datastore/scripts/update_index.py
```diff
@@ -43,39 +34,34 @@ def prettify_index(index, initial_indent=3):
...
-    elif opt in ('-a', '--app_id'):
-      app_id = arg
-
-  if not db_type or not app_id:
-    usage()
-    sys.exit(1)
+  parser = argparse.ArgumentParser(description='Updates composite indexes')
+  parser.add_argument('--type', '-t', default='cassandra',
+                      help='The datastore backend type')
+  parser.add_argument('--app_id', '-a', required=True, help='The project ID')
```
whoarethebritons: if we're moving to use project id in most cases, should the arg be changed to project_id?

cdonati: It would make more sense. I just generally try to change as little as possible to fix a problem in each pull request.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4841/

PR #2633, opened 2017-12-22
Don't wait for taskqueue master to start rabbitmq

cdonati:
If RabbitMQ is stopped on the taskqueue master before any of the other machines, it won't start without other machines being up.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4471/

PR #2634, opened 2017-12-26
Allow Java AppServer to use an external API Server

cdonati:
- Brings ApiProxyLocalImpl, DevAppServerMain, and DevAppServerImpl up to date with the 1.8.4 SDK
- Adds changes from the 1.9.54 SDK which allow starting a Python server for making API calls
- Adds a flag to use an external API server instead of starting one (eg. `--external_api_port=19998 --api_using_python_stub=app_identity_service`)

In my opinion, it's easiest to review this PR by commit because there are a lot of whitespace changes introduced by commits whose only purpose is to reduce the diffs against the SDK.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4857/

PR #2640, opened 2018-01-10
Register running instances with ZooKeeper

cdonati:
This replaces the HTTP callback that the AppManager server used to perform to register an AppServer instance.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -2531,78 +2531,6 @@ class Djinn
     return uuid
   end
 
-  # Instructs Nginx and HAProxy to begin routing traffic for the named
-  # version to a new AppServer.
-  #
-  # This method should be called at the AppController running the login role,
-  # as it is the node that receives application traffic from the outside.
-  #
-  # Args:
-  #   version_key: A String that identifies the version that runs the new
-  #     AppServer.
-  #   ip: A String that identifies the private IP address where the new
-  #     AppServer runs.
-  #   port: A Fixnum that identifies the port where the new AppServer runs at
-  #     ip.
-  #   secret: A String that is used to authenticate the caller.
-  #
-  # Returns:
-  #   "OK" if the addition was successful. In case of failures, the following
-  #   Strings may be returned:
-  #   - BAD_SECRET_MSG: If the caller cannot be authenticated.
-  #   - NO_HAPROXY_PRESENT: If this node does not run HAProxy (and thus cannot
-  #     add AppServers to HAProxy config files).
-  #   - NOT_READY: If this node runs HAProxy, but hasn't allocated ports for
-  #     it and nginx yet. Callers should retry at a later time.
-  def add_routing_for_appserver(version_key, ip, port, secret)
-    return BAD_SECRET_MSG unless valid_secret?(secret)
-
-    unless my_node.is_shadow?
-       # We need to send the call to the shadow.
-       Djinn.log_debug("Sending routing call for #{version_key} to shadow.")
-       acc = AppControllerClient.new(get_shadow.private_ip, @@secret)
-       begin
-         return acc.add_routing_for_appserver(version_key, ip, port)
-       rescue FailedNodeException
-         Djinn.log_warn("Failed to forward routing call to shadow (#{get_shadow}).")
-         return NOT_READY
-       end
-    end
-
-    project_id, service_id, version_id = version_key.split(
-      VERSION_PATH_SEPARATOR)
-    begin
-      version_details = ZKInterface.get_version_details(
-        project_id, service_id, version_id)
-    rescue VersionNotFound => error
-      return "false: #{error.message}"
-    end
-
-    APPS_LOCK.synchronize {
-      if @app_info_map[version_key].nil? ||
-          @app_info_map[version_key]['appservers'].nil?
-        return NOT_READY
-      elsif @app_info_map[version_key]['appservers'].include?("#{ip}:#{port}")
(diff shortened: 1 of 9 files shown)


AppController/djinn.rb
```diff
@@ -4927,6 +4855,40 @@ def find_lowest_free_port(starting_port)
   end
 
 
+  # Updates @app_info_map with registered instances.
```
whoarethebritons: can you add a note about how this must be called under APPS_LOCK

it currently is from scale_deployment since that is only ever run under APPS_LOCK, so there's no problem.

AppManager/app_manager_server.py
```diff
@@ -658,6 +675,168 @@ def create_java_start_cmd(app_name, port, load_balancer_host, max_heap,
...
+  Returns:
+    A tuple of strings specifying the instances nodes.
+  """
+  # Instances are indexed both by version and by machine.
+  instance_node_by_machine = '/'.join([machine_instances_node, instance_key])
+
+  revision_key, port = instance_key.split(':')
+  revision_parts = revision_key.split(VERSION_PATH_SEPARATOR)
+  version_key = VERSION_PATH_SEPARATOR.join(revision_parts[:3])
+  instance_node_by_version = '/appscale/instances_by_version/{}/{}'.format(
```
whoarethebritons: should this path be a constant?

AppManager/app_manager_server.py
```diff
@@ -724,6 +903,11 @@ def delete(version_key, port):
   options.define('db_proxy', appscale_info.get_db_proxy())
   options.define('tq_proxy', appscale_info.get_tq_proxy())
 
+  machine_instances_node = '/appscale/instances_by_machine/{}'.format(
```
whoarethebritons: where is instances_by_machine used?

AppController/djinn.rb
```diff
@@ -4927,6 +4855,40 @@ def find_lowest_free_port(starting_port)
...
+      ip = instance_key.split(':')[0]
+      match = known_instances.index("#{ip}:-1")
+      if match
+        known_instances.delete_at(match)
+      else
+        Djinn.log_warn("Adding an unassigned instance: #{instance_key}.")
+      end
+      known_instances << instance_key
+    }
+  end
```
obino: I think here we should also do the _other side of the coin_ check: anything in app_info_map but not registered, shouldn't it be removed? There is something like that for the compute node in check_running_appservers, also that part I think is redundant since now the AMS is doing the check of what is running locally (ie less interaction with Monit from different processes).

AppManager/app_manager_server.py
cdonati: @obino In the design I'm aiming at, we'll need two nodes to keep track of instance assignments (by machine and by version) and two nodes to keep track of instance registrations (by machine and by version).

The AppManager will need a watch on assignments_by_machine. The router will need a watch on instances_by_version. The autoscaler will need a watch on assignments_by_version (to determine if configured constraints are being met) and on instances_by_machine (to determine if assignments have been fulfilled).

AppManager/app_manager_server.py
cdonati: In a later branch, the autoscaler will have a watch on that node to help make assignment decisions. I can take it out for now though since nothing is using it at the moment.

AppController/djinn.rb
cdonati: The AppManager is not yet checking what is running locally (at least not on a regular basis). The router is still responsible for detecting failed instances and mutating app_info_map state.

I was considering this block to just be a direct replacement for the add_routing call to minimize changes. I was hoping to avoid a huge diff. However, I see now that instances could be erroneously re-added to app_info_map if the AppManager does not remove the instance registration entry before the next duty cycle. I'll probably need a separate pull request before this one to separate assignment and registration structures.

AppManager/app_manager_server.py
```diff
@@ -658,6 +675,168 @@ def create_java_start_cmd(app_name, port, load_balancer_host, max_heap,
...
+    monit_entries: A list of dictionaries containing instance details.
+  """
+  monitored = {(entry['revision'], entry['port']) for entry in monit_entries}
+  to_stop = []
+  for process in psutil.process_iter():
+    cmd = process.cmdline()
+    if len(cmd) < 2:
+      continue
+
+    if JAVA_APPSERVER_CLASS in cmd:
```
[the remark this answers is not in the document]
cdonati: You mean check if the process line has 'java' in it? I think `com.google.appengine.tools.development.DevAppServerMain` is unambiguous enough.

cdonati:
Closing this for now. I'll re-create it after the necessary changes to `@app_info_map` have been made.

AppManager/app_manager_server.py
obino: @cdonati got it. Would you mind if we get a quick presentation of this design after one of our daily? I think it's interesting to share :)

AppController/djinn.rb
obino: I see. But couldn't we give the shadow the biz of checking what's running? I guess my question is: do you believe to be monit or haproxy the best check for the failure or well-being of AppServers?

cdonati: I think the AppManager should be the one checking instance health. I don't think the autoscaler should be able to start and stop individual instances. It should just assign a count per version to each machine. Each compute machine should be responsible for the details about which individual instances should be running.

The AppManager could still use HAProxy as one condition of its health check.

PR #2641, opened 2018-01-10
Add GQL query to datastore viewer

cdonati:
This makes it easier to filter entities when viewing them in the dashboard.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4547/

PR #2642, opened 2018-01-12
Allow Hermes to check RabbitMQ status

cdonati:
This adds a `/stats/cluster/rabbitmq` endpoint that returns a structure like
```
{"nodes": [{"mem_alarm": false,
            "name": "rabbit@appscale-image0",
            "disk_free_alarm": false}],
 "queues": [{"messages": 0,
             "name": "appscaledashboard___default"}]}
```
If either mem_alarm or disk_free_alarm is `true` for any of the nodes, the TQ servers will be unable to enqueue messages.

I'm not attached to the endpoint. It just happens that `/stats/cluster...` is accessible from the AdminServer port already.


Hermes/appscale/hermes/stats/producers/rabbitmq.py
```diff
@@ -0,0 +1,78 @@
...
+    Returns:
+      A list of API resources.
+    """
+    ip = random.choice(self.taskqueue_machines)
+    url = 'http://{}:{}{}'.format(ip, self.API_PORT, endpoint)
+    creds = base64.b64encode(':'.join([self.USER, self.PASS]))
+    headers = {'Authorization': 'Basic {}'.format(creds)}
+    try:
+      response = yield self.client.fetch(url, headers=headers)
+    except Exception as error:
```
jeanleonov: Does tornado http client raise error if it receives 400+ code?

cdonati:
@jeanleonov I moved things around to more closely match the other producers. Let me know if that works. I removed the queue stats for now since they aren't divisible by machine. I'll proabably submit a separate pull request with a partial implementation of https://cloud.google.com/appengine/docs/standard/python/taskqueue/rest/taskqueues/get for push queues.

Hermes/appscale/hermes/stats/producers/rabbitmq_stats.py
```diff
@@ -0,0 +1,87 @@
...
+  disk_free_alarm = attr.ib()
+  mem_alarm = attr.ib()
+  name = attr.ib()
+
+
+@attr.s(cmp=False, hash=False, slots=True, frozen=True)
+class RabbitMQStatsSnapshot(object):
+  """ A container for holding RabbitMQ node stats. """
+  utc_timestamp = attr.ib()
+  rabbitmq_node_stats = attr.ib(metadata={Meta.ENTITY_LIST: RabbitMQNodeStats})
```
jeanleonov: @cdonati , I don't think that nested list is required here. Do we expected that there can be multiple node stats objects? Anyway, `RabbitMQStatsSource.get_current` method puts single object into this attribute.
So I would say that `RabbitMQNodeStats` and `RabbitMQStatsSnapshot` can be merged into single class with node stats and timestamp.

Hermes/appscale/hermes/stats/producers/rabbitmq_stats.py
```diff
@@ -0,0 +1,77 @@
...
+      raise APICallFailed('Call to {} failed: {}'.format(url, error))
+
+    try:
+      node_info = json.loads(response.body)
+    except ValueError:
+      raise APICallFailed('Invalid response from '
+                          '{}: {}'.format(url, response.body))
+
+    snapshot = RabbitMQStatsSnapshot(
+      utc_timestamp=int(time.time()),
```
jeanleonov: oh.. this was so simple.. and I was using `utc_timestamp = time.mktime(datetime.now().timetuple())`. (facepalm).

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4558/

PR #2643, opened 2018-01-12
Fixed overzealous ruby code checker

obino:
We need to have 'return' if the statement is not the last one of the
functions. Make sure we don't continue execution past a logic return.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/4541/

PR #2644, opened 2018-01-12
Fix AppControllerClient update_cron method

cdonati:
This bug crashed AppScale during an up when the shadow was not the first load balancer IP address.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4545/

PR #2645, opened 2018-01-12
Reduce Cassandra memory if running other services

cdonati:
When a database machine is running AppServer instances and other services, it's helpful to have a little extra room. This is most important for single-node deployments.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4585/

PR #2648, opened 2018-01-17
Respect defective 3rd-party tools and allow post requests

jeanleonov:
Why?????
https://github.com/christiangalsterer/httpbeat/blob/master/beater/poller.go#L132

HttpBeat rejects body for get requests, but I need to specify custom include_lists when I'm fetching stats from Hermes.

Any better ideas?


whoarethebritons:
I'll look into the metricbeat http module right now, I had put it on low priority.

alternatively, exclude on logstash side?

jeanleonov:
Excluding won't add needed fields which are missing in default include lists

jeanleonov:
Fortunately, @whoarethebritons found a way to use metricbeat.http instead of httpbeat, so there is no need to maintain POST requests instead of GET.

PR #2651, opened 2018-01-24
Pin eventlet version

cdonati:
This resolves https://github.com/eventlet/eventlet/issues/401.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4582/

obino:
Is this visible on docker? or jessie? If so, can we get a build on jessie, or a docker runtest?

cdonati:
@obino I was not able to reproduce the issue outside of our vagrant box. I'm not sure which package upgrade is causing the problem. In this case pinning the package is just forcing an upgrade, which fixes the issue.

PR #2653, opened 2018-01-28
Improve comment in fast-start script

obino:
Close #563

This PR is to mostly test the connection to redmine to ensure we can operate on tickets upon commits.


obino:
OCD runhttps://ocd.appscale.com:8080/job/Daily%20Build/4594/

cdonati:
Would a full link to the issue work (https://redmine.appscale.com/issues/563)? I don't think we should use just the bare number since it conventionally refers to an item in this repo.

whoarethebritons:
I think we'll have to change both the close keyword and the way to reference the issue.
http://www.redmine.org/projects/redmine/wiki/RedmineSettings#Referencing-issues-in-commit-messages

obino:
@cdonati yeah that was what the syntax should be to close ticket in redmine. I guess I can add the link independently, say on the next line?

cdonati:
@obino The problem is that Github is claiming that it will close "issue #563", which doesn't exist because that ID was assigned to a pull request. However, if that issue existed on Github, it seem like this syntax would close it.

I don't think it's a problem for this particular pull request, but I don't think this should be the norm for closing a redmine ticket. It would be confusing for someone viewing the pull request (seeing a link to a github resource), and it has the potential for unwanted side effects (closing an unrelated Github issue).

obino:
@cdonati ok I don't think redmine will accept full URLs for this, so I think we are stuck for now ... We can revisit the strategy for our integration.

whoarethebritons:
@cdonati @obino actually it will close the pull request because the github syntax recognizes that format and will close either the issue or the pull request. We'd run into bad things as soon as the redmine issue numbers were the same as pull requests

obino:
Ok. let me close this one, we'll have to test with a different repo then.

PR #2655, opened 2018-01-31
Register instances with ZooKeeper

cdonati:
This replaces the HTTP callback that the AppManager server used to perform to register an AppServer instance.

While similar to https://github.com/AppScale/appscale/pull/2640, it addresses the code review comments and moves the instance health check from the router to the AppManager.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -4909,6 +4909,9 @@ HOSTS
 
   # Updates @app_info_map with registered instances. This must be called under
   # APPS_LOCK.
+  #
+  # Args:
+  #   version_key: A string specifying the version key to update.
   def update_registered_instances(version_key)
     begin
       zk_instances = ZKInterface.get_children(
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -4921,11 +4921,9 @@ HOSTS
       return
     end
 
-    unless zk_instances.empty?
-      @app_info_map[version_key] = {} unless @app_info_map.key?(version_key)
-      unless @app_info_map[version_key].key?('appservers')
-        @app_info_map[version_key]['appservers'] = []
-      end
+    @app_info_map[version_key] = {} unless @app_info_map.key?(version_key)
+    unless @app_info_map[version_key].key?('appservers')
+      @app_info_map[version_key]['appservers'] = []
     end
     known_instances = @app_info_map[version_key]['appservers']
 
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -1883,7 +1883,7 @@ class Djinn
       # Load balancers and shadow need to check/update nginx/haproxy.
       if my_node.is_load_balancer?
         APPS_LOCK.synchronize {
-          check_haproxy
+          regenerate_routing_config
         }
       end
       @state = "Done starting up AppScale, now in heartbeat mode"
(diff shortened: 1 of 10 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4606/

AppManager/app_manager_server.py
```diff
@@ -767,6 +818,204 @@ def create_java_start_cmd(app_name, port, load_balancer_host, max_heap,
...
+  Args:
+    running_instances: An iterable of Instances.
+    zk_client: A KazooClient.
+  """
+  registered_instances = set()
+  for version_key in zk_client.get_children(VERSION_REGISTRATION_NODE):
+    version_node = '/'.join([VERSION_REGISTRATION_NODE, version_key])
+    for instance_entry in zk_client.get_children(version_node):
+      machine_ip, port = instance_entry.split(':')
+      port = int(port)
```
obino: sorry couldn't resist: perhaps move this after the check? No need to initialize it if we are not the right compute node.

AppController/djinn.rb
```diff
@@ -4996,6 +4907,49 @@ def find_lowest_free_port(starting_port)
...
+      else
+        Djinn.log_warn("Ignoring unassigned instance: #{instance_key}.")
+      end
+    }
+
+    # Account for instances that have been stopped.
+    known_instances.delete_if { |instance_key|
+      pending = instance_key.split(':')[-1] == '-1'
+      registered = zk_instances.include?(instance_key)
+      !pending && !registered
```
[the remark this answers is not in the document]
cdonati: This block is doing what you described as the "other side of the coin" check. If there's a non-pending entry in `@app_info_map` and the entry is not in the list of registered instances that we fetched from ZooKeeper, then we remove the entry from `@app_info_map` because the AppManager has stopped that previously-registered instance.

For example,
1. The shadow assigns an instance to machine1 (machine1:-1)
2. The AppManager on machine one starts the instance and adds a registration node.
3. The AppManager sees that HAProxy believes the instance is down. It stops the instance and removes the registration node.
4. The shadow (in this block) sees that it has an entry in `@app_info_map` that does not exist in the list of registered instances, so it removes the entry from `@app_info_map`.

AppController/djinn.rb
```diff
@@ -4996,6 +4907,49 @@ def find_lowest_free_port(starting_port)
...
 
 
+  # Updates @app_info_map with registered instances. This must be called under
+  # APPS_LOCK.
+  def update_registered_instances(version_key)
+    begin
+      zk_instances = ZKInterface.get_children(
+        "/appscale/instances_by_version/#{version_key}")
+    rescue FailedZooKeeperOperationException
+      Djinn.log_warn('Unable to update registered instances.')
```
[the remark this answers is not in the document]
cdonati: Unlike the library we use in Python, this method does not raise an exeption when the parent node does not exist. So if the version node does not exist yet, zk_instances will be an empty array.

The rescue happens when we have trouble making a ZooKeeper request. I've updated the log line to be more specific (unable to fetch list of registered instances).

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4851/

AppController/djinn.rb
```diff
@@ -4996,6 +4907,52 @@ def find_lowest_free_port(starting_port)
...
+      return
+    end
+
+    unless zk_instances.empty?
+      @app_info_map[version_key] = {} unless @app_info_map.key?(version_key)
+      unless @app_info_map[version_key].key?('appservers')
+        @app_info_map[version_key]['appservers'] = []
+      end
+    end
+    known_instances = @app_info_map[version_key]['appservers']
```
obino: Is there a chance that @app_info_map[version_key]['appservers'] is non-existent? Could zk_instances be empty and also app_info_map not been registered?

cdonati: Good point. There's no reason to avoid populating that if it doesn't exit.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4896/
https://ocd.appscale.com:8080/job/Migration/2765/

PR #2660, opened 2018-02-03
Execute tar command with environment variable

cdonati:
Older versions of Ruby do not support starting a shell line by defining an environment variable.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4617/

PR #2661, opened 2018-02-03
Install php-bcmath

cdonati:
The php7 distribution on xenial does not include bcmath, which the SDK uses.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4656/

PR #2664, opened 2018-02-06
Replace reference to deprecated role

cdonati:
This fixes the AppScalefile that the fast-start script generates.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4639/

PR #2671, opened 2018-02-09
Don't clear cron jobs during an up

cdonati:
We don't yet have a way to automatically recreate cron entries when they are removed, so we need to keep them across down,up cycles for now.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4650/

PR #2673, opened 2018-02-13
Wait for all load balancers to route instances

cdonati:
There are times when it takes a significant amount of time for a load balancer machine to reflect the latest instance registration list. This delay can cause push workers to fail if they try to make a request through a load balancer before it is updated.

This change makes that situation less likely at the cost of longer deploy operations.

### DIFF
--- a/AdminServer/appscale/admin/__init__.py
+++ b/AdminServer/appscale/admin/__init__.py
@@ -93,6 +93,20 @@ def wait_for_port_to_open(http_port, operation_id, deadline):
 
     yield gen.sleep(1)
 
+  for load_balancer in appscale_info.get_load_balancer_ips():
+    while True:
+      if time.time() > deadline:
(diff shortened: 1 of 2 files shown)


AdminServer/appscale/admin/__init__.py
```diff
@@ -93,6 +93,20 @@ def wait_for_port_to_open(http_port, operation_id, deadline):
 
     yield gen.sleep(1)
 
+  for load_balancer in appscale_info.get_load_balancer_ips():
```
obino: should you just skip the login_ip since you tested it just before getting into this loop?

cdonati: That would be more efficient, but I'm not sure how to find out which load balancer private IP matches the public login IP.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4816/

PR #2676, opened 2018-02-16
Stick to the duty cycle length

obino:


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/4700/

PR #2678, opened 2018-02-21
Prevent ProjectGroomer worker from stopping

cdonati:
Previously, an exception would stop the worker. If all the workers stop, the ProjectGroomer can't do any work.

### DIFF
--- a/AppDB/appscale/datastore/scripts/transaction_groomer.py
+++ b/AppDB/appscale/datastore/scripts/transaction_groomer.py
@@ -293,7 +293,11 @@ class ProjectGroomer(object):
       The transaction start time if still valid, None if invalid because this
       method will also delete it.
     """
-    tx_data = yield self._tornado_zk.get(tx_path)
+    try:
+      tx_data = yield self._tornado_zk.get(tx_path)
+    except NoNodeError:
+      raise gen.Return()
+
     tx_time = float(tx_data[0])
 
     _, container, tx_node = tx_path.rsplit('/', 2)
--- a/AppDB/appscale/datastore/scripts/transaction_groomer.py
+++ b/AppDB/appscale/datastore/scripts/transaction_groomer.py
@@ -213,6 +213,8 @@ class ProjectGroomer(object):
 
         if tx_time is not None and tx_time < self._oldest_valid_tx_time:
           self._oldest_valid_tx_time = tx_time
+      except Exception:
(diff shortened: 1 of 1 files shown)


AppDB/appscale/datastore/scripts/transaction_groomer.py
```diff
@@ -213,6 +213,8 @@ def _worker(self):
 
         if tx_time is not None and tx_time < self._oldest_valid_tx_time:
           self._oldest_valid_tx_time = tx_time
+      except Exception:
```
scragraham: Should the except block(s) be more specific here instead of a catch all Exception? or do we not know what kinds of specific exceptions self._resolve_txid() will throw?

cdonati: Usually, I think it's better to catch only expected exceptions. However, in cases where the loop should never stop, I think it's better to catch Exception. In this situation, we would not be treating different errors in a different way.

I'm pretty confident that most of the exceptions will be a subset of KazooException or a cassandra-driver related exception, but I can't be 100% certain that the library or our code does not have a bug that would cause a different exception to be raised.

scragraham: /me grumbles. 

I get the need for the Exception, but that level of exception shouldn't really happen. It should be more alarming if we got an unexpected exception. Even if we aren't treating the exceptions differently, it is better to go ahead and catch the expected exceptions. 

My reasoning is twofold: 

1. If needed, it becomes easier to debug the exceptions as the expected exceptions are already "filled out"
2. I've seen blanket Exceptions come back to haunt you. When someone, **cough** not me **cough**, does something as simple as access a method/attribute that doesn't exist and it flies under the radar because nobody ever really looks at the stack trace...because "that always happens".

cdonati: Would it help if I moved the worker portion out to a separate AsyncExecutorPool class? I'd like to avoid a situation where changes to the resolution logic requires changes to the executor.

scragraham: @cdonati you are way more qualified to answer that question that I am :) I can only suggest that we do it right the first time.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4723/
The hawkeyes failed due to cassandra connection timeouts.
https://ocd.appscale.com:8080/job/Tools-test/3082/

PR #2681, opened 2018-02-22
Run apps without application element in app.yaml

cdonati:
This uses the application flag introduced in the 1.9.8 SDK to override the application element in the app.yaml.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4739/

PR #2684, opened 2018-02-27
Allow group locks to persist when commits fail

cdonati:
This prevents other clients from writing to a group before a large batch is fully applied. During each commit, there is now an additional ZooKeeper write to the transaction ID node that associates a list of entity groups with the transaction.

This pull request also fixes an issue where reads during puts (to determine index deletions) were not performed inside of a group lock. The fix may slightly decrease performance.

Finally, this pull request fixes an issue where transaction IDs were not cleaned up after a ConcurrentModificationError (when one of the groups that were read for a transaction was updated after the transaction started).

### DIFF
--- a/AppDB/appscale/datastore/datastore_distributed.py
+++ b/AppDB/appscale/datastore/datastore_distributed.py
@@ -542,13 +542,6 @@ class DatastoreDistributed():
       composite_indexes: A list or tuple of CompositeIndex objects.
     """
     self.logger.debug('Inserting {} entities'.format(len(entities)))
-    entity_keys = []
-    for entity in entities:
-      prefix = self.get_table_prefix(entity)
-      entity_keys.append(get_entity_key(prefix, entity.key().path()))
-
-    current_values = self.datastore_batch.batch_get_entity(
-      dbconstants.APP_ENTITY_TABLE, entity_keys, APP_ENTITY_SCHEMA)
 
     by_group = {}
     for entity in entities:
@@ -569,6 +562,12 @@ class DatastoreDistributed():
         lock.acquire()
       except entity_lock.LockTimeout:
         raise TimeoutError('Unable to acquire entity group lock')
+
+      entity_keys = [
+        get_entity_key(self.get_table_prefix(entity), entity.key().path())
+        for entity in entity_list]
+      current_values = self.datastore_batch.batch_get_entity(
+        dbconstants.APP_ENTITY_TABLE, entity_keys, APP_ENTITY_SCHEMA)
 
       batch = []
       entity_changes = []
--- a/AppDB/appscale/datastore/datastore_distributed.py
+++ b/AppDB/appscale/datastore/datastore_distributed.py
@@ -560,6 +560,7 @@ class DatastoreDistributed():
       group_key = entity_pb.Reference(encoded_group_key)
 
       txid = self.transaction_manager.create_transaction_id(app, xg=False)
+      self.transaction_manager.set_groups(app, txid, [group_key])
       lock = entity_lock.EntityLock(self.zookeeper.handle, [group_key], txid)
       try:
         with lock:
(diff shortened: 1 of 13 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4753/

AppDB/appscale/datastore/zkappscale/transaction_manager.py
```diff
@@ -152,18 +142,31 @@ def get_open_transactions(self):
...
     Args:
       path: A string specifying a ZooKeeper path.
     """
     try:
-      try:
-        self.zk_client.delete(path)
-      except NotEmptyError:
-        # Cross-group transaction nodes have a child node.
-        self.zk_client.delete(path, recursive=True)
+      self.zk_client.delete(path, recursive=True)
```
jeanleonov: `NotEmptyError` can be removed from imports.

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -558,34 +552,47 @@ def put_entities(self, app, entities, composite_indexes=()):
...
+
+        batch.append({'table': 'group_updates',
+                      'key': bytearray(encoded_group_key),
+                      'last_update': txid})
+
+        entity_changes.append(
+          {'key': entity.key(), 'old': current_value, 'new': entity})
+      self.datastore_batch.batch_mutate(app, batch, entity_changes, txid)
+
+      lock.release()
```
jeanleonov: Does timeout suppose to help to resolve issue when lock is acquired but couldn't be released because of failure in between?
Doesn't it make sense to wrap it with try/finally?

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -894,23 +898,28 @@ def dynamic_delete(self, app_id, delete_request):
...
+          raise TimeoutError('Unable to acquire entity group lock')
+
+        self.delete_entities(
+          group_key,
+          txid,
+          key_list,
+          composite_indexes=filtered_indexes
+        )
+
+        lock.release()
```
jeanleonov: The same question here

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -3217,66 +3226,72 @@ def apply_txn_changes(self, app, txn):
...
+    lock = entity_lock.EntityLock(self.zookeeper.handle, decoded_groups, txn)
+    try:
+      lock.acquire()
+    except entity_lock.LockTimeout:
+      raise TimeoutError('Unable to acquire entity group locks')
+
+    group_txids = self.datastore_batch.group_updates(metadata['reads'])
+    for group_txid in group_txids:
+      if group_txid in metadata['in_progress'] or group_txid > txn:
+        lock.release()
```
jeanleonov: And here

AppDB/appscale/datastore/datastore_distributed.py
```diff
@@ -3217,66 +3226,72 @@ def apply_txn_changes(self, app, txn):
...
+         'last_update': txn})
 
-      for group in groups_mutated:
-        batch.append(
-          {'table': 'group_updates', 'key': bytearray(group),
-           'last_update': txn})
+    self.datastore_batch.batch_mutate(app, batch, entity_changes, txn)
 
-      self.datastore_batch.batch_mutate(app, batch, entity_changes, txn)
+    lock.release()
```
jeanleonov: And here

AppDB/appscale/datastore/datastore_distributed.py
cdonati: Are you referring to the LockTimeout when acquiring the lock? If that is thrown, the lock object cleans up after itself.

If the lock is acquired, we don't want the locks to be cleaned up unless everything succeeds. This branch removes the try/finally behavior on the lines you commented because we want the locks to stick around when there is an issue applying the commit. If we wrap it in try/finally, the datastore will always clean up the locks even in cases when a large batch might not be fully committed.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4771/

PR #2691, opened 2018-03-12
Handle docker as a provider

cdonati:
This treats the "Docker" provider the same way as it does a cluster.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4798/

PR #2692, opened 2018-03-14
Use datastore server registry to toggle read-only mode

cdonati:
This anticipates moving datastore servers away from set locations.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -1468,10 +1468,8 @@ class Djinn
   def set_node_read_only(read_only, secret)
     return BAD_SECRET_MSG unless valid_secret?(secret)
     return INVALID_REQUEST unless %w(true false).include?(read_only)
-    read_only = read_only == 'true'
 
-    DatastoreServer.set_read_only_mode(read_only)
-    if read_only
+    if read_only == 'true'
       GroomerService.stop
     else
       GroomerService.start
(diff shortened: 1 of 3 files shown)


AppController/lib/zkinterface.rb
```diff
@@ -443,6 +443,14 @@ def self.get_cron_config(project_id)
     return JSON.load(cron_config_json)
   end
 
+  def self.get_datastore_servers()
```
obino: No need for ().

AppController/lib/zkinterface.rb
```diff
@@ -443,6 +443,14 @@ def self.get_cron_config(project_id)
...
+    return get_children('/appscale/datastore/servers').map { |server|
```
obino: Should there be some sort of contant here for the path? like /appscale should be our ZK_ROOT or something similar?

AppController/djinn.rb
```diff
@@ -1488,6 +1486,13 @@ def set_read_only(read_only, secret)
     return BAD_SECRET_MSG unless valid_secret?(secret)
     return INVALID_REQUEST unless %w(true false).include?(read_only)
 
+    ZKInterface.get_datastore_servers.each { |machine_ip, port|
+      http = Net::HTTP.new(machine_ip, port)
+      request = Net::HTTP::Post.new('/read-only')
```
obino: Is there a chance of exception here or timeouts? Should we consider to retry in case something goes bad? Wondering if we can get in a situation were some datastore are not in read only mode.

cdonati: Yes, there's always a chance that the call will timeout or fail. However, that would cause the SOAP request to fail. The user can retry the call since it's idempotent.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4817/

PR #2694, opened 2018-03-16
Adjustable max appservers

obino:


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/4808/

AppController/djinn.rb
```diff
@@ -5320,6 +5333,8 @@ def get_scaling_info_for_version(version_key)
     scaling_params = version_details.fetch('automaticScaling', {})
     min = scaling_params.fetch('minTotalInstances',
                                Integer(@options['default_min_appservers']))
+    max = scaling_params.fetch('maxTotalInstances',
```
jeanleonov: Do we already support autoscaling parameters in `app.yaml` Or it's reserve for the future?
Both ways looks good to me.

obino: I don't think so: it's more for future development.

cdonati: Just FYI, we support the params server-side. Setting those params for the Admin API call from the app.yaml would just require tools changes.

PR #2695, opened 2018-03-19
Replace usages of execute_concurrent

cdonati:
This is to address https://datastax-oss.atlassian.net/browse/PYTHON-939, which prevents us from returning a useful error in some cases.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4816/

obino:
@cdonati is there any performance implication with this?

cdonati:
@obino The main difference is that this change removes the concurrency limit. By default, `execute_concurrent` limits it to 100 requests at a time. Though it's possible a commit would exceed this, it would be unlikely (eg. cross-group transaction of 10 entites with each having 10 index writes). In that situation, we would use a bit more memory (keeping track of the future objects), but I don't think there would be any significant performance impact.

PR #2696, opened 2018-03-25
Use the ServiceManager to start datastore servers

cdonati:
This adds a new service to the AdminServer called the ServiceManager.

The ServiceManager starts and monitors processes according to ZooKeeper assignments. It watches `/appscale/assignments/<private_ip>`.

For now, there is no change in policy about how many datastore servers to start and where they should be started. Moving the datastore servers to the compute machines will come in a later pull request.

Since datastore servers are no longer managed by Monit, this PR also adds an `appscale-admin summary` command prints a table of all running AppScale-related processes on the machine.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4829/

AdminServer/appscale/admin/service_manager.py
```diff
@@ -0,0 +1,410 @@
...
+
+    log_file = os.path.join(LOG_DIR,
+                            'datastore_server-{}.log'.format(self.port))
+    self._stdout = open(log_file, 'a')
+    self.process = psutil.Popen(
+      ['cgexec', '-g', ':'.join(DATASTORE_CGROUP)] + start_cmd,
+      stdout=self._stdout, stderr=subprocess.STDOUT)
+
+    # Wait for server to bind to port before making HTTP requests.
+    yield gen.sleep(1)
```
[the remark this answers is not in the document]
cdonati: The problem I ran into during testing is that if the http client was in the process of trying a request while the server started, the server could not bind to the port. The ServiceManager would eventually get a server started, but it was kind of annoying to have failed starts. Do you think it would be better to be more persistent with the server to try to bind to port on startup?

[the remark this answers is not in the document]
jeanleonov: I didn't know that requesting port can affect server ability to bind port.
Adding retries on binding port process with reasonable timeout sounds good to me.
Though it can go as a separate issue out of this PR.

cdonati:
@jeanleonov Thank you for the detailed review. Let me know if the added commits are adequate for this pull request.

AdminServer/tests/test_service_manager.py
```diff
@@ -0,0 +1,85 @@
...
+from appscale.admin.service_manager import (
+  DatastoreServer, gen, options, psutil, ServerStates, ServiceManager,
+  ServiceTypes)
+
+FakeHTTPResponse = namedtuple('Response', ['code'])
+
+# Skip sleep calls.
+sleep_response = Future()
+sleep_response.set_result(None)
+gen.sleep = MagicMock(return_value=sleep_response)
```
jeanleonov: Though I don't think this particular mock can affect any other tests, it is always safer to us `setUp*` and `tearDown*` function/methods for making sure that some real functions are mocked just for time test is running.
To patch something once for all test module you can use [setUpModule and tearDownModule functions](https://docs.python.org/2/library/unittest.html#setupmodule-and-teardownmodule).

So it may look like:
```
patchers = []

def setUpModule():
  patcher = patch.object(gen, 'sleep')
  patchers.append(patcher)
  sleep_response = Future()
  sleep_response.set_result(None)
  sleep_mock = patcher.start()
  sleep_mock.return_value = sleep_response

def tearDownModule():
  for patcher in patchers:
    patcher.stop()
```

AdminServer/appscale/admin/service_manager.py
```diff
@@ -0,0 +1,409 @@
...
+    """
+    persistent_update_services = retry_data_watch_coroutine(
+      self._assignments_path, self._update_services)
+    try:
+      assignments = json.loads(encoded_assignments)
+    except (TypeError, ValueError):
+      if encoded_assignments:
+        raise
+
+      assignments = {}
```
jeanleonov: Isn't it better to check if `encoded_assignments` is empty before loading json instead of `try-except`?
```
assignments = json.loads(encoded_assignments) if encoded_assignments else {}
```

AdminServer/appscale/admin/service_manager.py
```diff
@@ -0,0 +1,409 @@
...
+        logger.info('Stopping {}'.format(server))
+        IOLoop.current().spawn_callback(server.stop)
+        stopped += 1
+
+      return
+
+    for _ in range(to_start):
+      port = self._get_open_port()
+      server_class = self.SERVICE_MAP[service_type]['server']
+      server = server_class(port, self._http_client, options['verbose'])
```
jeanleonov: Shouldn't service general options be stored outside assignments nodes which are node-specific, when `verbose` is service specific option (independent from nodes).

cdonati: I agree that the verbose flag should be stored separately from the assignment. It requires a bit of preparatory work on the logger object and setting up an additional watch in the datastore server. I'm inclined to leave this pull request with the instance options just so we can unblock datastore autoscaling, but I plan to follow up with an additional pull request to remove that option from the assignment.

AdminServer/appscale/admin/service_manager.py
cdonati: Good point. That is more concise and probably more readable too.

PR #2697, opened 2018-03-26
Give more informative messages for GQL errors

cdonati:
This gives more insight into the most common issues when using the GQL query feature.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4961/

PR #2699, opened 2018-03-28
Raise exception when unable to fetch children

cdonati:
Previously, an error (a connection loss, for example) would return an empty array instead of raising an exception.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4839/

PR #2700, opened 2018-03-29
Fix nginx config for handlers with secure field

tmarballi:
Currently, we don't handle well the handlers specified in the app.yaml without the 'secure:' and with 'secure: never' field. This PR does the following:
(1) When 'secure: never' is specified, it has a locations section in both http and https (with a redirect back to http port)
(2) When 'secure:' field is omitted, it has a locations section in http and https does not have a redirect.


tmarballi:
@cdonati That makes complete sense. I will make the changes to remove the grouping of all the handlers and update this PR.

cdonati:
@tmarballi Removing the combined blocks was a step in the right direction, but by avoiding the grouping, I meant that there shouldn't be separate sections for different handler types (eg. non_secure_http_locations, always_secure_locations, etc). The order of the nginx blocks should match the order of the app yaml handlers.

For example, let's say an app.yaml has
1. handler1: secure not specified
2. handler2: secure: always
3. handler3: secure not specified

The HTTP section of the nginx config should have
1. handler1 block: forward to app
2. handler2 block: respond with 302
3. handler3 block: forward to app

Since master (and this PR) groups the handlers, it generates
1. handler1 block: forward to app
2. handler3 block: forward to app
3. handler2 block: respond with 302

AppController/lib/helperfunctions.rb
```diff
@@ -1155,12 +1152,11 @@ def self.get_secure_handlers(version_key)
...
     handlers.map! do |handler|
-      next unless handler.key?('secure')
-
-      if handler['secure'] == 'always'
-        secure_handlers[:always] << handler
-      elsif handler['secure'] == 'never'
-        secure_handlers[:never] << handler
+      if !handler.key?("static_dir") && !handler.key?("static_files")
+        if !handler.key?('secure')
+            handler['secure'] = 'non_secure'
```
obino: Ruby prefers handler['secure'] = 'non_secure' if !handler.key?('secure')

AppController/lib/nginx.rb
```diff
@@ -107,12 +107,41 @@ def self.write_fullproxy_version_config(version_key, http_port, https_port,
...
+        "\n\tclient_body_timeout   600;\n\tclient_max_body_size  2G;" \
+        "\n    }\n"
+
+    combined_http_locations = ""
+    combined_https_locations = ""
+    secure_handlers.each do |handler|
+      if handler["secure"] == "always"
+        handler_location = HelperFunctions.generate_secure_location_config(handler, https_port)
+        combined_http_locations = combined_http_locations + handler_location
+        always_secure_locations = always_secure_locations + handler_location
```
obino: You can use the += form here.

AppController/lib/nginx.rb
```diff
@@ -107,12 +107,41 @@ def self.write_fullproxy_version_config(version_key, http_port, https_port,
...
+      elsif handler["secure"] == "never"
+        handler_https_location = HelperFunctions.generate_secure_location_config(handler, http_port)
+        combined_https_locations = combined_https_locations + handler_https_location
```
obino: Same += ...

AppController/lib/nginx.rb
```diff
@@ -107,12 +107,41 @@ def self.write_fullproxy_version_config(version_key, http_port, https_port,
...
+        handler_http_location = "\n    location ~ #{handler['url']} {"
+        handler_http_location << http_location_params
+        combined_http_locations = combined_http_locations + handler_http_location
+        never_secure_locations = never_secure_locations + handler_http_location
+      elsif handler["secure"] == "non_secure"
+        handler_https_location = "\n    location ~ #{handler['url']} {"
+        handler_https_location << http_location_params
+        combined_http_locations = combined_http_locations + handler_https_location
```
obino: same +=

AppController/lib/nginx.rb
```diff
@@ -107,12 +107,41 @@ def self.write_fullproxy_version_config(version_key, http_port, https_port,
...
+    http_location_params = \
+        "\n\tproxy_set_header      X-Real-IP $remote_addr;" \
+        "\n\tproxy_set_header      X-Forwarded-For $proxy_add_x_forwarded_for;" \
+        "\n\tproxy_set_header      X-Forwarded-Proto $scheme;" \
+        "\n\tproxy_set_header      X-Forwarded-Ssl $ssl;" \
+        "\n\tproxy_set_header      Host $http_host;\n\tproxy_redirect        off;" \
+        "\n\tproxy_pass            http://gae_ssl_#{version_key};" \
+        "\n\tproxy_connect_timeout 600;\n\tproxy_read_timeout    600;" \
+        "\n\tclient_body_timeout   600;\n\tclient_max_body_size  2G;" \
+        "\n    }\n"
```
obino: Do we need to set all of these on each URL?

tmarballi: @obino Quite frankly I am not entirely sure, I tried to keep the same configuration without grouping the handlers. I could look more into it and see if we do indeed have to set all of these but right now I just kept the same config.

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/5213/
Tools test (rerun): https://ocd.appscale.com:8080/job/Tools-test/3486/

cdonati:
Demo:
https://github.com/AppScale/hawkeye/pull/143 with master: https://ocd.appscale.com:8080/job/Hawkeye/15313/
https://github.com/AppScale/hawkeye/pull/143 with this branch: https://ocd.appscale.com:8080/job/Hawkeye/15314/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5269/
https://ocd.appscale.com:8080/job/Tools-test/3533/

PR #2702, opened 2018-04-03
Consider the version's previous ports as available

cdonati:
Without this, it's not possible to relocate just one port (eg. 8080/4380 -> 8080/443).


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4944/

PR #2705, opened 2018-04-04
Use zookeeper for stopped versions

whoarethebritons:
stop_versions used to use the file tree to determine what was running, which was inaccurate.

ocd.appscale.com:8080/job/Daily%20Build/4864/


AppController/djinn.rb
```diff
@@ -1901,13 +1853,11 @@ def job_start(secret)
...
-      # Detect applications that have been undeployed and terminate all
-      # running AppServers.
-      check_stopped_apps
-
-      # Load balancers and shadow need to check/update nginx/haproxy.
+      # Load balancers and shadow need to check/update applications that have
+      # been undeployed and nginx/haproxy.
       if my_node.is_load_balancer?
         APPS_LOCK.synchronize {
+          check_stopped_apps
```
cdonati: I think the shadow also needs to run this.

whoarethebritons: The shadow is a load_balancer

cdonati: It might be right now by coincidence, but it's a separate role conceptually. There shouldn't be anything that prevents us from assigning instance locations from a machine that's not a load balancer.

AppController/djinn.rb
```diff
@@ -4597,49 +4566,44 @@ def place_error_app(version_key, err_msg)
...
-      configured_versions = ZKInterface.get_versions
-    rescue FailedZooKeeperOperationException
-      Djinn.log_warn(
-        'Failed to fetch configured versions when checking stopped apps')
-      return
-    end
-
-    version_list.each { |version_key|
+    removed_versions = []
+    @versions_loaded.each { |version_key|
```
cdonati: Rather than a single loop on `@versions_loaded`, I think the actual resources that are being affected should be checked. For example, something like
```
if my_node.is_shadow?
  syslog_files.each do |version_key, file_location|
    next if version_key in configured_versions
    FileUtils.rm(file_location)
    HelperFunctions.shell("service rsyslog restart")
  end
end
```
I don't think we have any guarantees that the configuration files on disk match `@versions_loaded`. For example, the controller could finish setting up an application and then crash before writing its state. When it comes back up, there could be artifacts on disk that it should clean up even though the restored `@versions_loaded` does not account for them.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/5760/
https://ocd.appscale.com:8080/job/Tools-test/3943/

PR #2706, opened 2018-04-04
Replace gen.Return() with bare return statements

cdonati:
According to http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.Return, this is preferred when returning early from a coroutine.

A bare return has the benefit of not being caught by an `except Exception` block in the same coroutine.


scragraham:
> "In Python 3.3, this exception is no longer necessary: "

Dumb question: do we require python 3.3+?

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4892/

PR #2707, opened 2018-04-05
Do not reuse SourceManager future if it failed

cdonati:
Previously, a failure to fetch the source would prevent the archive from ever being fetched since the the manager would just yield the failed future on subsequent requests.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4890/

PR #2708, opened 2018-04-06
Add logservice API to external API server

cdonati:
The request log writer runs on a separate thread so that it can keep a socket open to the log server and continue to write logs as long as there are items in the queue.

Using the external API server for the logservice API adds a bit of latency to each request since the dispatcher needs to send an HTTP request every time it starts and stops an application request. This is a temporary design decision because the dispatcher and API server will eventually be re-merged.


APIServer/appscale/api_server/log/writer.py
```diff
@@ -0,0 +1,107 @@
...
+            except Empty:
+                if self._worker_socket is not None:
+                    self._worker_socket.close()
+                    self._worker_socket = None
+                continue
+
+            message = request_log.to_capnp().to_bytes()
+            self._send_message(message)
+
+    @retry(retry_on_exception=socket.error, backoff_threshold=60, max_retries=None)
```
jeanleonov: By default, `@retry` decorator is configured to run up to 10 retries in at most 60 seconds. To remove timeout limit you need to set `retrying_timeout=None`.

Do you think it's bad idea to have two retrying constraints by default?

APIServer/tests/unit/test_log_messages.py
```diff
@@ -0,0 +1,232 @@
...
+                                              RequestLog)
+from appscale.api_server.log import log_service_pb2
+import unittest
+import time
+import base64
+
+LOG_LEVEL_INFO = 1
+LOG_LEVEL_WARNING = 2
+
+MESSAGE_1 = 'show me what you got'
```
jeanleonov: =)

AppServer/google/appengine/tools/devappserver2/module.py
```diff
@@ -32,6 +33,7 @@
 import threading
 import time
 import urllib
+import urllib2
```
jeanleonov: As far as I know urllib2 doesn't manage connections pool, so it opens creates new connection for every request. Shouldn't we try using `requests` or other lib which manager connection pool?

APIServer/appscale/api_server/log/messages.py
```diff
@@ -0,0 +1,292 @@
+""" LogService message containers. """
+
+from __future__ import division
+
+import base64
+import time
+from collections import deque
+
+import capnp
```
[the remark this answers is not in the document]
cdonati: This is the same thing we were doing before. I just tried to make our custom implementation more explicit. Here's our existing taskqueue stub: https://github.com/AppScale/appscale/blob/master/AppServer/google/appengine/api/logservice/logservice_stub.py#L21. I would like to remove capnp eventually.

APIServer/appscale/api_server/app_identity/service.py
```diff
@@ -19,26 +28,26 @@
 logger = logging.getLogger('appscale-api-server')
 
 
-class UnknownError(Exception):
+class UnknownError(ApplicationError):
     """ Indicates that the request cannot be completed at this time. """
-    pass
+    def __init__(self, message):
```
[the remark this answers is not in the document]
cdonati: Yes. Good idea.

APIServer/appscale/api_server/log/messages.py
```diff
@@ -0,0 +1,292 @@
...
+                '{status} {size} "{ua}"').format(
+            ip=self.ip, nickname=self.nickname, date=date, method=self.method,
+            resource=self.resource, ver=self.http_version, status=self.status,
+            size=self.response_size, ua=self.user_agent)
+
+    @staticmethod
+    def from_capnp(log):
+        """ Creates a RequestLog from a capnp RequestLog. """
+        service_id = 'default'
+        if ':' in log.versionId:
```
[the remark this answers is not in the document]
cdonati: When starting the request, the SDK uses the version string, which does not have the service ID defined if "default": https://github.com/AppScale/appscale/blob/dbf42d8b3208d995d67d0f8529a6f34a169661cb/AppServer/google/appengine/tools/devappserver2/application_configuration.py#L136

APIServer/appscale/api_server/app_identity/service.py
```diff
@@ -147,15 +156,17 @@ def sign(self, blob):
...
         """ Makes the appropriate API call for a given request.
 
         Args:
             method: A string specifying the API method.
             encoded_request: A binary type containing the request details.
+            request_id: A string specifying the request ID.
         Returns:
             A binary type containing the response details.
         """
+        del request_id
```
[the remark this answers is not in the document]
cdonati: It's not strictly required, but it deals with unused variable warnings. I took the advice from https://stackoverflow.com/a/14836005.

AppServer/google/appengine/tools/devappserver2/module.py
cdonati: That's a good point. We are making a lot of requests to the same socket.

APIServer/appscale/api_server/log/writer.py
cdonati: I don't think it's bad to have two constraints. However, it might make sense to have the default be infinite. https://pypi.org/project/backoff seems to be the most popular library that's similar to this, and it seems like it is infinite by default.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5028/

APIServer/appscale/api_server/log/messages.py
```diff
@@ -0,0 +1,292 @@
...
+        for capnp_field in ['requestId', 'appId', 'ip', 'nickname', 'userAgent',
+                            'host', 'method', 'resource', 'httpVersion', 'status',
+                            'responseSize', 'offset', 'startTime', 'endTime']:
+            field = self.CAPNP_FIELDS.get(capnp_field, capnp_field)
+            value = getattr(self, field)
+            if value is not None:
+                setattr(request_log, capnp_field, value)
+
+        request_log.versionId = ':'.join([self.service_id, self.version_id])
+        app_logs = request_log.init('appLogs', len(self.app_logs))
```
whoarethebritons: Is this app_logs variable ever appended to again? Because capnp lists initialized this way cannot be resized.

APIServer/appscale/api_server/log/writer.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _process_logs(self):
+        """ Fetches logs from the queue and sends them to the log server. """
+        while True:
+            timeout = None
+            if self._worker_socket is not None:
+                timeout = self.IDLE_SOCKET_TIMEOUT
+
+            try:
+                request_log = self._queue.get(timeout=timeout)
```
whoarethebritons: Does this end up with the EOF error when the appserver is shutting down?

APIServer/tests/unit/test_log_messages.py
whoarethebritons: 🎉

APIServer/appscale/api_server/log/messages.py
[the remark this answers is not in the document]
whoarethebritons: I think that capnp might be useful and that the main annoying thing about the logservice is not capnp but actually twistd.

APIServer/appscale/api_server/log/messages.py
cdonati: No. All of the capnp serialization happens only when the message is sent.

APIServer/appscale/api_server/log/writer.py
cdonati: Since the server running this is external to AppServer instances, it won't be affected by them shutting down.

For this thread worker, there will only be a message in the queue if the AppServer sends the final "end request" http call.

APIServer/appscale/api_server/log/messages.py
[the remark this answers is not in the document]
cdonati: It might be useful if it were saving us any CPU when dealing with serializing and deserializing messages. At this time, the runtime serializes to protobuf, the API server deserializes protobuf and reserializes to capnp. Coming back out, it needs to deserialize from capnp and reserialize to protobuf. Since the runtime sends serialized protobuf, it would actually save us quite a bit of CPU to just keep that way.

On the Python side, it's an additional dependency that takes quite a long time to install. It also leaves us with extra message definitions and conversion code.

cdonati:
I will wait to demo this one. Performance can be an issue. My plan is to allow the appmanager to start multiple API servers if it's running more than one runtime instance. Once that is merged, I will demo this branch.

jeanleonov:
Just as a side comment, if it appears that tornado consumes noticeable amount of CPU or introduces unnecessary delays (when it manages coroutines concurrency) isn't API Server a good candidate to try Async IO?
I was already mentioning in comments that it has better functionality to work with sockets and performance benchmark shows that even 2 years ago aiohttp was faster than tornado (https://klen.github.io/py-frameworks-bench/). I couldn't find newer benchmarks=(

cdonati:
I'm closing this for now. I think the external API server performance should scale with the number of CPUs available on the machine it's running on.

PR #2711, opened 2018-04-09
Use HTTP API to stop Monit services

cdonati:
Monit 5.18 changed its default output to a table, so it's no longer reasonable to loop through lines in the output of `monit summary`.

As a bonus, this branch also cuts the stop time in half by stopping ZooKeeper last.

### DIFF
--- a/AdminServer/appscale/admin/stop_services.py
+++ b/AdminServer/appscale/admin/stop_services.py
@@ -12,24 +12,27 @@ def main():
   hostname = socket.gethostname()
 
   print('Waiting for monit to stop services')
+  stopped_count = 0
   while True:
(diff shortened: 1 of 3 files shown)


jeanleonov:
On the other hand, it is always nice to see how djinn is getting smaller and smaller=)

cdonati:
@obino Probably not. It was just easier to implement in Python since we already have the library for dealing with the Monit API.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5112/
The functional test failures were due to a msrest package update. Here are the same build builds with the appropriate tools branch:
https://ocd.appscale.com:8080/job/Blobstore/3547/
https://ocd.appscale.com:8080/job/Tools-test/3385/

PR #2720, opened 2018-04-13
Ensure Celery worker gets restarted

cdonati:
Occasionally, Monit will get interrupted during a restart. This pidfile check retries the update_worker call if the PID did not change.

### DIFF
--- a/AdminServer/appscale/admin/push_worker_manager.py
+++ b/AdminServer/appscale/admin/push_worker_manager.py
@@ -125,7 +125,8 @@ class ProjectPushWorkerManager(object):
     except IOError:
       new_pid = None
 
-    assert new_pid != old_pid
+    if new_pid == old_pid:
+      raise AssertionError
 
   @gen.coroutine
   def stop_worker(self):
--- a/AdminServer/appscale/admin/push_worker_manager.py
+++ b/AdminServer/appscale/admin/push_worker_manager.py
@@ -11,7 +11,7 @@ from tornado.ioloop import IOLoop
 from tornado.options import options
 
 from appscale.common.async_retrying import (
-  retry_children_watch_coroutine, retry_data_watch_coroutine
+  retry_children_watch_coroutine, retry_coroutine, retry_data_watch_coroutine
 )
 from appscale.common.constants import (
   LOG_DIR,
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4921/

AdminServer/appscale/admin/push_worker_manager.py
```diff
@@ -101,6 +107,22 @@ def update_worker(self, queue_config):
...
+    except IOError:
+      new_pid = None
+
+    # Occasionally, Monit will get interrupted during a restart. Retry the
+    # restart if the Celery worker PID is the same.
+    if new_pid == old_pid:
+      logger.warning(
+        '{} worker PID did not change. Restarting it.'.format(self.project_id))
+      yield gen.sleep(1)
+      yield self.update_worker(queue_config)
```
jeanleonov: I think update in this method will badly work in following scenario:

1. method is called and get old PID 12345
2. it sends command to monit to starts/restarts celery worker
5. sleep, check new PID and get the same 12345, sleep
6. do recursive call
7. get old PID 56789 (it could be changed after initial call checked new PID)
8. `-.-.-` doing the same until process ID is changed strictly during method execution.

Do I missing something? It seems that start/restart operations should be done once and then we should wait until PID is changed.

cdonati: That's a good point. I think the sleep after the log line is not needed.

However, we can't just wait until the PID changes because if Monit is interrupted, it will never change.

cdonati:
@jeanleonov I'm wary about waiting less than a second. We could probably go down to half a second, but anything smaller than that has the potential to confuse Monit. This can definitely be improved when we move away from using Monit here.

AdminServer/appscale/admin/push_worker_manager.py
cdonati: Oh, are you saying we should wait with an exponential backoff for the pid to change with a longer time limit before retrying the function? That's a good idea.

AdminServer/appscale/admin/push_worker_manager.py
```diff
@@ -101,6 +107,26 @@ def update_worker(self, queue_config):
...
+  @staticmethod
+  @retry_coroutine(retrying_timeout=10, retry_on_exception=[AssertionError])
+  def ensure_pid_changed(old_pid, pid_location):
+    try:
+      with open(pid_location) as pidfile:
+        new_pid = int(pidfile.read().strip())
+    except IOError:
+      new_pid = None
+
+    assert new_pid != old_pid
```
jeanleonov: Though I like implementation with assert statement, Codacy raised a good point, that assert statements can be ignored in some interpreters for optimization purposes.

AdminServer/appscale/admin/push_worker_manager.py
jeanleonov: I'm afraid I don't fully understand a possible issue with monit interruption. But I like you solution and it resolves my initial concern.

PR #2722, opened 2018-04-13
RM-626 Fetch application archive as soon as it's available in deployment

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4923/
I'm going to measure how faster deploy process became (if did).
I don't really like the fact that presence of archive is verified on different levels.. Will try to improve this logic.


AdminServer/appscale/admin/instance_manager/source_manager.py
```diff
@@ -25,6 +27,10 @@
 logger = logging.getLogger('appscale-admin')
 
 
+class AlredyHoster(Exception):
```
cdonati: Already

AdminServer/appscale/admin/instance_manager/source_manager.py
```diff
@@ -37,6 +43,61 @@ def __init__(self, zk_client, thread_pool):
...
+      revision_key_parts = revision_key.split(VERSION_PATH_SEPARATOR)
+      project_id = revision_key_parts[0]
+      service_id = revision_key_parts[1]
+      version_id = revision_key_parts[2]
+
+      service_manager = self.projects_manager[project_id][service_id]
+      version_details = service_manager[version_id].version_details
+      runtime = version_details['runtime']
+      source_archive = version_details['deployment']['zip']['sourceUrl']
+      yield self.ensure_source(revision_key, source_archive, runtime)
```
cdonati: I would recommend reducing the new_apps_list to a set of versions. The sourceUrl from the version node will not match an older revision from /apps, and it would not make sense to fetch an older revision anyway (if, for example, the AppManager gets restarted and loses track of its fetched versions).

AdminServer/appscale/admin/instance_manager/source_manager.py
jeanleonov: ooops, thanks!

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4929/

PR #2725, opened 2018-04-18
Fix the version delete API method

### LINKED ISSUE #2724: Admin API version delete calls always fail
Version delete calls always return a 500 due to an incorrect treatment of the version ID.

cdonati:
This gives clients back a proper operation object after a successful delete instead of a 500.

Resolves #2724


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4967/

PR #2726, opened 2018-04-18
Fix dashboard app list

obino:
The dashboard will report no application after an attempt to delete/undeploy a currently running application. This PR fixes this behavior, making sure it shows the current list of running applications after any delete attempt. Also sometime an error wouldn't be displayed if there is not running application: this fixes that too, ensuring the error message takes priority.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/4932/

PR #2729, opened 2018-04-21
Reduce diffs with the 1.8.0 SDK

cdonati:
This updates outdated components that lagged behind the 1.8.0 SDK and reduces non-semantic changes to make updating easier.

I got about halfway through the diffs before deciding that the pull request was getting large and that further changes should go in a separate pull request.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4953/

cdonati:
@jeanleonov When reducing diffs, I compare the AppServer directory with the target version directory (in this case, 1.8.0) with a diff tool. For every diff, I check:

1. Is this an AppScale-related change?
2. Does this come from an earlier SDK?
3. Does this come from a later SDK?

Sometimes it's tricky to determine which category the diff falls under, but I find it's helpful to use a git annotation tool to see the commit message and date associated with the lines.

If it's an AppScale-related change, I try to add a comment that indicates as much.

If it's from an earlier SDK, I bisect previous versions until I find the SDK where that portion is updated. If one of the lines is distinctive enough, it's often faster to search a directory (containing a bunch of unpacked SDKs) for the line and see the earliest ref. When I commit, I include the SDK version where that change was introduced in the commit message.

If it's from a later SDK, I ignore it. When we finally upgrade to that SDK, it will no longer be a diff.

AppServer/google/appengine/api/memcache/memcache_stub.py
```diff
@@ -322,6 +335,7 @@ def _Dynamic_Increment(self, request, response):
           memcache_service_pb.MemcacheServiceError.UNSPECIFIED_ERROR)
     response.set_new_value(new_value)
 
+  @apiproxy_stub.Synchronized
```
sjones4: Should _Dynamic_Increment also have @apiproxy_stub.Synchronized?

cdonati: It looks like the inner function (_internal_increment) has it, and that's the only portion that has a side effect. The rest of the function is just reading the request and preparing the response.

I should mention for context that this particular part of the code is not used in AppScale. It's useful as a reference for implementing the API calls, but it does not share state between AppServer processes. Our implementation is in the memcache_distributed module.

PR #2733, opened 2018-04-26
Use ranges to keep track of merge join queries

### LINKED ISSUE #2730: Allow merge join queries to skip ahead
At the moment, the skip ahead feature for merge joins is not working. If there are no suitable entries in a given chunk for one of the ranges, the scan will stop. For example:

Range 1:
```
value1/key00001
...[10,000 keys later]
value1/key10002
```

Range 2:
```
value2/key10002
```

After fetching the first chunk from range 1, the query does not skip ahead like it should (to the first key in range 2) since there are no relevant entries from the range.

cdonati:
This allows the matcher to skip ahead with only the ranges that need to be adjusted. It also reduces the index entry fetch size from 10,000 to 1000.

Resolves #2730


cdonati:
I'm going to re-create this PR based on post-async master to avoid the huge merge commit.

PR #2735, opened 2018-05-01
Treat commit operations on expired transactions as bad requests

cdonati:
This matches the behavior in GAE and allows the client to differentiate between write timeouts and situations when the commit wasn't even tried.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5010/

scragraham:
demo looks good.

PR #2736, opened 2018-05-03
Use ranges to keep track of merge join queries

### LINKED ISSUE #2730: Allow merge join queries to skip ahead

cdonati:
This allows the matcher to skip ahead with only the ranges that need to be adjusted. It also reduces the index entry fetch size from 10,000 to 1000.

Resolves #2730

### DIFF
--- a/AppDB/appscale/datastore/range_iterator.py
+++ b/AppDB/appscale/datastore/range_iterator.py
@@ -1,0 +1,231 @@
+""" Iterates through a range of index entries. """
+
+import sys
+from collections import namedtuple
+
+from tornado import gen
+
+from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
+from appscale.datastore.dbconstants import (
+  ASC_PROPERTY_TABLE, BadRequest, KEY_DELIMITER, PROPERTY_SCHEMA,
+  TERMINATING_STRING)
+from appscale.datastore.utils import decode_path, encode_index_pb
+
+sys.path.append(APPSCALE_PYTHON_APPSERVER)
+from google.appengine.datastore.datastore_pb import Path, Query_Filter
+
+
+IndexEntry = namedtuple('IndexEntry',
+                        ['encoded_path', 'entity_reference', 'key', 'path'])
+
+
+Cursor = namedtuple('Cursor', ['key', 'inclusive'])
+
+
+class RangeExhausted(Exception):
+  """ Indicates that there are no more entries in the range. """
+  pass
+
+
+class RangeIterator(object):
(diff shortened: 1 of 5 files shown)


AppDB/appscale/datastore/range_iterator.py
```diff
@@ -0,0 +1,231 @@
...
+    start_key = KEY_DELIMITER.join([self.prefix, str(encode_index_pb(path))])
+    end_key = ''.join([start_key, TERMINATING_STRING])
+    if start_key < self._range[0] or end_key > self._range[-1]:
+      raise BadRequest('Restriction must be within range')
+
+    if self._cursor.key > end_key:
+      raise BadRequest('Cursor already exceeds new range')
+
+    self._range = (start_key, end_key)
+    self._cursor.key = max(start_key, self._cursor.key)
```
sjones4: isn't self._cursor immutable? (namedtuple)

AppDB/appscale/datastore/range_iterator.py
```diff
@@ -0,0 +1,231 @@
...
+
+class RangeIterator(object):
+  """ Iterates through a range of index entries.
+
+  This was designed for merge join queries. The range can only be narrowed.
+  """
+  CHUNK_SIZE = 1000
+
+  def __init__(self, db, project_id, namespace, kind, prop_name, value):
+    """ Creates a new IndexReader.
```
whoarethebritons: RangeIterator

whoarethebritons: RangeIterator

AppDB/appscale/datastore/range_iterator.py
cdonati: Whoops. You're right. It also looks like we weren't testing that code path. https://github.com/AppScale/hawkeye/pull/140 should cover that case.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5152/

cdonati:
This branch with https://github.com/AppScale/hawkeye/pull/140: https://ocd.appscale.com:8080/job/Hawkeye/14930/

scragraham:
demo looks good

Issue #2737, opened 2018-05-03
Improve retrying tools


jeanleonov:
Add backoff sequence class for generating exponential delays for retries.

PR #2738, opened 2018-05-03
Issue 2737 Backoff sequence

jeanleonov:
Resolves https://github.com/AppScale/appscale/issues/2737
https://ocd.appscale.com:8080/job/Daily%20Build/4981/


common/appscale/common/retrying.py
```diff
@@ -26,9 +31,135 @@ def not_missed(value_1, value_2):
   raise BothMissedException()
 
 
+class BackoffSequence(object):
+  """
+  Iterable sequence of backoff delays.
+  This class suppose to be in cases when you need to do retries.
+  
+  Usage examples (find more in common/test/unit/test_retrying.py):
+
```
scragraham: It might be helpful to highlight what the defaults are so that when someone inspects the object via dir() or help() they can quickly see what the default attributes are. For example with <span>#</span>1 I don't know how many times it will retry the operation by default.

common/appscale/common/async_retrying.py
```diff
@@ -157,39 +156,40 @@ def check_exception(error):
...
-        while True:
+        for backoff in backoff_sequence:
           # Start of retrying iteration
           try:
             result = func(*args, **kwargs)
             if isinstance(result, gen.Future):
               result = yield result
             break
 
+          except gen.Return:
```
cdonati: It looks like this path bypasses the lock release that happens at the end.

jeanleonov: Good point. And actually we don't need to catch it at all..
If it's coroutine, `gen.Return` exception is catch by tornado decorator, otherwise it shouldn't raise `gen.Return` at all, so we don't need to handle it differently comparing to any other Exception.

Thanks for pointing me to this mistake.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4983/

jeanleonov:
https://ocd.appscale.com:8080/job/Euca-Faststart/2406/

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5580/

PR #2739, opened 2018-05-04
faster way to get cpu count which takes .003ms vs .014ms

scragraham:
Stumbled across this, not sure if we really need the change. But its faster to just use grep on the file and get the count directly via the grep command:

```
1514% time cat /proc/cpuinfo | grep processor | wc -l
4
cat /proc/cpuinfo  0.00s user 0.00s system 48% cpu 0.004 total
grep processor  0.00s user 0.00s system 26% cpu 0.006 total
wc -l  0.00s user 0.00s system 39% cpu 0.004 total


1515% time grep -c processor /proc/cpuinfo
4
grep -c processor /proc/cpuinfo  0.00s user 0.00s system 65% cpu 0.003 total

```

### DIFF
--- a/AppController/lib/datastore_server.rb
+++ b/AppController/lib/datastore_server.rb
@@ -66,7 +66,7 @@ module DatastoreServer
   # Number of servers is based on the number of CPUs.
   def self.number_of_servers
     # If this is NaN then it returns 0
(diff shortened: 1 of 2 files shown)


scragraham:
Well it looks like we don't use this code anymore anyways as the CPU count is gathered via Hermes (which gets it via psutil). This PR is no longer valid as the code is now gone in master.

Issue #2740, opened 2018-05-04
Use Protocol Buffers library with datastore server


cdonati:
This should speed up serializing and deserializing protocol buffer messages and remove the datastore server's AppServer code dependencies.

PR #2741, opened 2018-05-06
Replace stub with TaskQueue client

cdonati:
This allows the datastore to make TaskQueue service requests without importing AppServer code.

These changes make progress toward #2740

### DIFF
--- a/AppDB/appscale/datastore/taskqueue_client.py
+++ b/AppDB/appscale/datastore/taskqueue_client.py
@@ -1,0 +1,91 @@
+""" Performs TaskQueue service enqueue requests. """
+
+import sys
+
+import requests
+from requests.exceptions import ConnectionError, HTTPError
+
+from appscale.common.constants import TASKQUEUE_SERVICE_PORT
+from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
+
+sys.path.append(APPSCALE_PYTHON_APPSERVER)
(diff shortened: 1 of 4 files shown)


AppDB/appscale/datastore/taskqueue_client.py
```diff
@@ -0,0 +1,91 @@
...
+    encoded_api_request = api_request.Encode()
+    headers = {'ProtocolBufferType': 'Request',
+               'AppData': project_id,
+               'Module': service_id,
+               'Version': version_id}
+    api_response = None
+    for location in self._locations:
+      url = 'http://{}'.format(location)
+      try:
+        response = requests.post(url, data=encoded_api_request,
```
jeanleonov: As it's new code, could it be not blocking from the very beginning?
I see, the only user of this method is run in separate thread, and it's not a big performance improvement. But I would vote for making new function asynchronous where it doesn't require big effort.

cdonati: Yes, good point. I'll replace the requests usage with Tornado's async client.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5031/
The demo is that the Java taskqueue_tests.TransactionalTaskTest succeeded: https://ocd.appscale.com:8080/job/Hawkeye/14531/artifact/hawkeye-logs/java-detailed%202018-05-16%2014-50-21.log (line 2888)

PR #2748, opened 2018-05-09
Handle "service" element

### LINKED ISSUE #2747: Support "service" keyword in app.yaml
Currently, instances will crash if the included app.yaml has defined "service" instead of "module".

cdonati:
This allows instances to start with the "service" element instead of "module".

Resolves #2747


scragraham:
demo looks good.

PR #2749, opened 2018-05-10
Define CURRENT_MODULE_ID

cdonati:
This environment variable was introduced in the 1.8.3 SDK. For now, CURRENT_VERSION_ID remains as it was before (eg. service-id:version-id.minor-version) for compatibility reasons.


cdonati:
I'm going to hold off on this one for awhile. I think it would make more sense to first restore the runtime's default task targets behavior so that CURRENT_VERSION_ID is consistent with the SDK.

cdonati:
5da2f9c implements this in a more thorough way.

PR #2754, opened 2018-05-13
Handle reads with the index past the last block

cdonati:
Return an empty string when asked for a chunk that's outside of the stored blob.

If the end of a blob happens to match the end of a block, a BlobReader may ask for the next chunk as it does not know it has read the end of the blob. In that case, it does not make sense to claim that the blob does not exist. Rather, like a typical file handle, it should return an empty byte string.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5056/

PR #2757, opened 2018-05-16
Wait for all hosts before creating tables

cdonati:
The most common error we see when running the prime script is that one or more hosts are immediately marked as "down" after the connect call finishes. Those machines later rejoin and cause a schema disagreement. This change is meant to prevent the script from continuing immediately if not all hosts are available.

### DIFF
--- a/AppDB/appscale/datastore/cassandra_env/schema.py
+++ b/AppDB/appscale/datastore/cassandra_env/schema.py
@@ -250,6 +250,8 @@ def prime_cassandra(replication):
   deadline = time.time() + SCHEMA_CHANGE_TIMEOUT
   while True:
     if time.time() > deadline:
+      logging.warning('Timeout when waiting for hosts to join. Continuing '
+                      'with connected hosts.')
       break
 
     if len(session.get_pool_state()) == len(hosts):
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5032/

AppDB/appscale/datastore/cassandra_env/schema.py
```diff
@@ -248,6 +246,17 @@ def prime_cassandra(replication):
                   timeout=SCHEMA_CHANGE_TIMEOUT)
   session.set_keyspace(KEYSPACE)
 
+  logging.info('Waiting for all hosts to be connected')
+  deadline = time.time() + SCHEMA_CHANGE_TIMEOUT
+  while True:
+    if time.time() > deadline:
+      break
```
jeanleonov: @cdonati shouldn't we report warning in this case?
Without log message we have no straight way to verify if deadline was exceeded and we create table without having all nodes available.. it will look just like a normal creation of tables with all hosts online.

cdonati: @jeanleonov Yes, a warning log line would be appropriate. I'll update.

PR #2758, opened 2018-05-19
Make sure we create static entried for nginx

obino:
This PR ensure that everytime we invoke parse_static_data we get the
full list of static files and directory from app.yaml, even if we ask to
not copy the files into place.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5038/

Issue #2759, opened 2018-05-21
PullQueues implementation using Postgres as a backend


jeanleonov:
Create additional implementation of PullQueues using Postgresql. It should be configurable for project, so one project can work using Cassandra, other - Postgresql.

PR #2760, opened 2018-05-21
PullQueueus on postgresql

jeanleonov:
Resolves https://github.com/AppScale/appscale/issues/2759.

#### Overview ####
This PR allows to use Postgres as a backend for PullQueues (though Cassadra is still used by default).

#### Instructions ####
To make it work you need to create Postgres database (one per project) and create zookeeper node with connection string for the DB. e.g.:
```
create /appscale/projects/hawkeyepython27/postgres_dsn "dbname=appscale-hawkeyepython27 user=appscale password=securepassword host=10.10.9.243"
```

When TaskQueue is started or it is notified about new project it tries to get value of postgres_dsn node in zookeeper, if value is specified, it uses Postgres implementation, otherwise old implementation is used.

#### SQL schema ####
A table with two indexes are created for every pull queue:
```
CREATE TABLE IF NOT EXISTS "{table_name}" (
  task_name varchar(500) NOT NULL,
  time_deleted timestamp DEFAULT NULL,   -- used to mark task as deleted
  time_enqueued timestamp NOT NULL,
  lease_count integer NOT NULL,
  lease_expires timestamp NOT NULL,
  payload bytea,
  tag varchar(500),
  PRIMARY KEY (task_name)
);

-- Partial index aims to keep ONLY active tasks indexed
CREATE INDEX IF NOT EXISTS "{table_name}-eta-retry-tag-index" 
  ON "{table_name}" USING BTREE (lease_expires, lease_count, tag) 
  WHERE time_deleted IS NULL;

-- Partial index aims to keep ONLY active tasks indexed
CREATE INDEX IF NOT EXISTS "{table_name}-retry-eta-tag-index" 
  ON "{table_name}" (lease_count, lease_expires, tag) 
  WHERE time_deleted IS NULL;
```

#### Known issues/doubts ####
 - `PostgresPullQueue` doesn't provide stats about tasks leased during last minute/hour (old implementation does).
 - It doesn't remove tasks which were permanently failed. And I think it's right behavior, though I couldn't find clear documentation on this topic for GAE. Having lots of permanently failed tasks will cause Postgres using less effective index, but it still should work fine.
 - It doesn't remove corresponding table and indices if queue is deleted, so if you delete queue with not finished tasks and then recreate it for different purposes, you can get old tasks.

#### Build ####
https://ocd.appscale.com:8080/job/Daily%20Build/5044/


jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5047/

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+  def tasks_table_name(self):
+    return 'pullqueue-{}'.format(self.name)
+
+  def ensure_tables_created(self):
+    try:
+      self.pg_cursor.execute(
+        'CREATE TABLE IF NOT EXISTS "{table_name}" ('
+        '  task_name varchar(500) NOT NULL,'
+        '  time_enqueued timestamp NOT NULL,'
+        '  retry_count integer NOT NULL,'
```
cdonati: It might make sense to name this lease_count since it's a little more explicit about what it's indicating and it's a little easier to reason about. For example, if the task has been leased exactly once, does that mean that the task has been (re)tried?

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+        '  lease_expires timestamp NOT NULL,'
+        '  payload text,'
```
cdonati: This should probably be `bytea` since the user can store arbitrary blobs.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+      task: A Task object.
+    Raises:
+      InvalidTaskInfo if the task ID already exists in the queue
+        or it doesn't have payloadBase64 attribute.
+    """
+    import psycopg2
+    if not hasattr(task, 'payloadBase64'):
+      raise InvalidTaskInfo('{} is missing a payload.'.format(task))
+
+    enqueue_time = datetime.datetime.utcnow()
```
cdonati: I would recommend letting Postgres generate the timestamp.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+    try:
+      lease_expires = task.leaseTimestamp
+    except AttributeError:
+      lease_expires = datetime.datetime.utcfromtimestamp(0)
```
cdonati: Since Postgres does not suffer from the same clock issues that can affect immediate availabilty on Cassandra, this can use the current timestamp (also generated by Postgres). That allows leases to be ordered by enqueue time when eta is not specified.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+        '  task_name, payload, time_enqueued, '
+        '  lease_expires, retry_count, tag '
+        ')'
+        'VALUES ( '
+        '  %(task_name)s, %(payload)s, %(time_enqueued)s, '
+        '  %(lease_expires)s, %(retry_count)s, %(tag)s '
+        ')'.format(table=self.tasks_table_name),
+        vars={
+          'task_name': task.id,
+          'payload': task.payloadBase64,
```
cdonati: I believe I made a mistake by storing the data as base64-encoded strings in the Cassandra implementation. The base64-encoding should be left as a serialization detail for the REST API layer only.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+
+      self.pg_connection.commit()
+      logger.debug('Added task: {}'.format(task))
+
+    except psycopg2.IntegrityError as err:
+      name_taken_msg = 'Task name already taken: {}'.format(task.id)
+      logger.warning('{msg}. Rolling back transaction({err})'
+                     .format(msg=name_taken_msg, err=err))
+      self.pg_connection.rollback()
+      raise InvalidTaskInfo(name_taken_msg)
```
cdonati: To match GAE behavior (which we don't do in the Cassandra implementation), we would need to differentiate between an existing task and a tombstoned task name. I'm fine with leaving that detail for later, though.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+
+  def delete_task(self, task):
+    """ Deletes a task from the queue.
+
+    Args:
+      task: A Task object.
+    """
+    try:
+      self.pg_cursor.execute(
+        'DELETE FROM "{tasks_table}" '
```
cdonati: We may want to delete all the fields (except the task_name and maybe a field to mark when it was deleted) instead of deleting the whole row in order to tombstone the task name. A periodic task could then clear all names that were tombstoned earlier than a set interval into the past.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+    """ Updates the duration of a task lease.
+
+    Args:
+      task: A Task object.
+      new_lease_seconds: An integer specifying when to set the new ETA. It
+        represents the number of seconds from now.
+    Returns:
+      A Task object.
+    """
+    new_eta = current_time_ms() + datetime.timedelta(seconds=new_lease_seconds)
```
cdonati: This timestamp can be generated by the db.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -235,7 +235,538 @@ def __repr__(self):
...
+    try:
+      self.pg_cursor.execute(
+        'UPDATE "{tasks_table}" '
+        'SET lease_expires = %(new_eta)s '
+        'WHERE task_name = %(task_name)s '
+        '  AND lease_expires > %(current_time)s '
```
cdonati: Any time the current time is need for a db operation, the native postgres function should be used.

AppTaskQueue/appscale/taskqueue/queue.py
jeanleonov: Thanks. As differentiating errors would require changes in http handlers, I would implement it as a separate PR.

AppTaskQueue/appscale/taskqueue/queue.py
jeanleonov: I'd like to keep this PR focused on Postgres implementation of Pull queues and would avoid changes related to all regular Push and Pull queues.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5070/

AppTaskQueue/appscale/taskqueue/queue.py
cdonati: Sorry, I was not suggesting that you change existing implementations. It would just be a detail about how Postgres is storing the payload. You could just decode the base64 string right before storing it. In practice, it would leave behind an unnecessary encode step, but that would be easier to fix later than having to worry about converting existing data.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5077/

jeanleonov:
After merging master into the branch:
https://ocd.appscale.com:8080/job/Daily%20Build/5236/

PR #2773, opened 2018-05-25
Give the runtime the task name that will be used

### LINKED ISSUE #2761: Transactional task names do not match RPC results
When enqueuing transactional tasks, the API server returns a different chosen task name than what ends up being used by the push queue service.

cdonati:
Previously, the final task name was being namespaced twice, and the project ID was not set.

Resolves #2761.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5141/

PR #2775, opened 2018-05-28
Remove indexless sorted ancestor queries

cdonati:
In GAE, sorted ancestor queries require a composite index.

It's not really possible to order the results without fetching the entire entity group, which is not reasonable for large entity groups. It's also not possible to use a cursor since the order of the index does not match the order of the results.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5064/

cdonati:
As far as I can tell, it's not possible to make this kind of query in AppScale right now. The runtime datastore client will complain that an index is required before even sending the request.

I don't know why it was implemented, to be honest. It could be that App Engine used to behave differently.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5187/
https://ocd.appscale.com:8080/job/Euca-Faststart/2552/

PR #2778, opened 2018-05-29
Restart rsyslog so the app logs can be recreated

### LINKED ISSUE #2772: Combined app logs are not re-created
Upon a "down --clean", the combined app logs are removed while rsyslog still has them open. When AppScale is started again, the log files are not recreated because rsyslog is still writing to the deleted file descriptor.

The rsyslog service should be restarted on a "down --clean".

cdonati:
Without this, the daemon continues writing to the file descriptors of the deleted combined application logs.

Resolves #2772.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5071/

PR #2780, opened 2018-05-29
Increase open file limit for nginx

### LINKED ISSUE #2779: Nginx will only open 1024 concurrent connections
On distros with systemd, the nginx service does not use the limit we set in /etc/security/limits.conf.

cdonati:
This allows nginx to open more than 1024 concurrent connections.

Resolves #2779.

### DIFF
--- a/debian/appscale_install_functions.sh
+++ b/debian/appscale_install_functions.sh
@@ -163,9 +163,9 @@ EOF
 
     # On distros with systemd, the open file limit must be adjusted for each
     # service.
-    if which systemctl; then
+    if which systemctl > /dev/null; then
         mkdir -p /etc/systemd/system/nginx.service.d
-        cat <<EOF | tee /etc/systemd/system/nginx.service.d/override.conf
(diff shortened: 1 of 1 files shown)


debian/appscale_install_functions.sh
```diff
@@ -160,6 +160,17 @@ root            soft    nofile           200000
 *               soft    nofile           200000
 *               -       nproc            32768
 EOF
+
+    # On distros with systemd, the open file limit must be adjusted for each
+    # service.
+    if which systemctl; then
+        mkdir -p /etc/systemd/system/nginx.service.d
+        cat <<EOF | tee /etc/systemd/system/nginx.service.d/override.conf
```
scragraham: Why the 'tee' here? Do we need to see the output of this command?

cdonati: Good point. I don't think the output is helpful.

scragraham:
Demo looks good.

PR #2785, opened 2018-06-06
Fix bootstrap from branch options

sjones4:
This pull request simplifies using bootstrap.sh with a branch option.

With the current script to bootstrap the latest code from master you could use:

```
# bootstrap.sh --tag dev
```

but using only branch options:

```
# bootstrap.sh --branch master --tools-branch master
```

results in use of the latest tag rather than the master branch.

You cannot work around this by passing the "dev" tag in addition to the desired branches as with branches other than "master" this will fail with the error "--branch cannot be specified with --tag".

With this fix in place the behavior is:

```
# bootstrap.sh                                       # latest tag
# bootstrap.sh --tag last                            # latest tag
# bootstrap.sh --tag dev                             # master branch
# bootstrap.sh --branch master --tools-branch master # master branch
# bootstrap.sh --branch my-branch                    # my-branch branch and master for tools
# bootstrap.sh --tools-branch my-branch              # master branch and my-branch for tools
```

the following case that previously worked now fails:

```
# bootstrap.sh --tag dev --branch master --tools-branch master
```

When bootstrap is from a git tag the output now shows the tag used rather than a branch.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -12,7 +12,12 @@ APPSCALE_BRANCH="master"
 APPSCALE_TOOLS_BRANCH="master"
 FORCE_UPGRADE="N"
 UNIT_TEST="n"
-GIT_TAG=""
+GIT_TAG="last"
+
+# Options for tag/branches
+OPT_APPSCALE_BRANCH=""
+OPT_APPSCALE_TOOLS_BRANCH=""
+OPT_GIT_TAG=""
 
 usage() {
     echo "Usage: ${0} [--repo <repo>][--tools-repo <repo>][-t]"
@@ -59,7 +64,7 @@ while [ $# -gt 0 ]; do
         if [ -z "${1}" ]; then
             usage
         fi
-        APPSCALE_BRANCH="${1}"
+        OPT_APPSCALE_BRANCH="${1}"
         shift
         continue
     fi
@@ -68,7 +73,7 @@ while [ $# -gt 0 ]; do
         if [ -z "${1}" ]; then
             usage
         fi
-        GIT_TAG="${1}"
+        OPT_GIT_TAG="${1}"
         shift
         continue
     fi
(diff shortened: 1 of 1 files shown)


sjones4:
Looks like existing users of the bootstrap.sh script expect to be able to pass both tag and branch options, so this pull request needs updating.

A reasonable approach that allows for this could be to respect any branch options that are passed when the "dev" tag is used (i.e. "dev" tag is the same as not specifying a tag), and ignore any branch options when some other tag is specified.

cdonati:
@sjones4 I would be fine with breaking backwards compatibility for this script, but I don't feel strongly about it.

Ignoring branch options when tag is something other than "dev" would be strange, though. If the user specifies both a tag and branch, it seems like a mistake to me. I would prefer the command to fail in that case rather than proceed and ignore some arguments.

scragraham:
I wonder if we should just get rid of the --tag option altogether:
```
bootstrap.sh   # always uses last tag
bootstrap.sh --dev  # master
bootstrap.sh --branch <branch|tag> --tools-branch <branch|tag>
```
If only one of --branch or --tools-branch are specified should we use the defaults of bootstrap.sh, which is latest tag? 
```
bootstrap.sh --branch foobar     # appscale:foobar, tools:last tag
bootstrap.sh --tools-branch bar # tools:bar, appscale:last tag
bootstrap.sh --dev --tools-branch bar # tools:bar, appscale:master
```
--dev used with both --tools-branch and --branch should either be ignored (as the user has specified both branches) or error out.

side note: It looks like git always orders the tags, so if you create tags out of order: 2.0.0, 3.0.0, 2.2.0. 3.0.0 will show up as the 'last' tag.

I don't like the use of '--tag {last|dev}' especially since those aren't real tags.

scragraham:
The lazy part of me likes '--dev', but you are correct we could remove it and just specify the branches

sjones4:
I worked around this by conditionally adding the "dev" tag when using the master branch. This isn't a  perfect solution but should be sufficient for now.

PR #2791, opened 2018-06-11
Do not validate SSL certs for SOAP services

### LINKED ISSUE #2789: channel API not working
Environment: clean xenial machine; appscale installed per instructions with boostrap.sh; environment is functioning as expected

Issue: using the channel API sample from http://github.com/AppScale/sample-apps results in this error anytime the **channel.create_channel** method is called:

**SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)**

Will AppScale continue to support the Channel API, or are there plans to deprecate and retire it, as Google has done?  And, if so, will there be something to replace it?

cdonati:
This restores the Channel API's behavior prior to Python 2.7.9. It takes HTTPTransport from SOAPpy and modifies the call method to ignore the self-signed SSL certificates from the AppController and the UAServer.

Resolves #2789.

### DIFF
--- a/AppServer/google/appengine/api/xmpp/unverified_transport.py
+++ b/AppServer/google/appengine/api/xmpp/unverified_transport.py
@@ -1,0 +1,136 @@
+""" A SOAPpy transport that does not validate SSL certificates. """
+import Cookie
+import base64
+import httplib
+import inspect
+import ssl
+import urllib
+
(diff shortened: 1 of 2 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5217/

PR #2793, opened 2018-06-11
Define proper Java version for Trusty

cdonati:
Without this, a build on a clean box fails because JAVA_HOME is set to the OpenJDK 7 directory.


cdonati:
Demo
With master: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/6208/
With branch: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/6209/ (still fails, but it gets past the SDK compilation)

PR #2795, opened 2018-06-11
Limit Pip version

cdonati:
Starting with version 10, Pip does not allow updating packages that were installed with the distro's package manager.

### DIFF
--- a/debian/appscale_install_functions.sh
+++ b/debian/appscale_install_functions.sh
@@ -653,16 +653,17 @@ prepdashboard()
 upgradepip()
 {
     # Versions older than Pip 7 did not correctly parse install commands for
-    # local packages with optional dependencies.
+    # local packages with optional dependencies. Versions greater than Pip 9
+    # do not allow replacing packages installed by the distro.
     case "$DIST" in
         wheezy|trusty)
-            pipwrapper pip
+            pipwrapper 'pip<10'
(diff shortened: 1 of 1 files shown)


debian/appscale_install_functions.sh
```diff
@@ -653,16 +653,17 @@ prepdashboard()
...
         wheezy|trusty)
-            pipwrapper pip
+            pipwrapper 'pip<10'
             # Account for the change in the path to the pip binary.
             hash -r
             ;;
         jessie)
             # The system's pip does not allow updating itself.
-            easy_install --upgrade pip
+            easy_install --upgrade 'pip<10'
```
sjones4: With easy_install I was getting version 10.0.0b2 when using <10. To get a 9.x version this worked:

```
easy_install --upgrade 'pip<10.0.0b1'
```

Which I guess is ok since there will not be any older pre-release versions coming out.

cdonati: Interesting. Thanks!

cdonati:
Demo
With master: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/6319/
With branch: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/6321/ (still fails, but installs google-api-python-client successfully)

PR #2801, opened 2018-06-14
Check db load to scale

obino:


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5120/

obino:
@jeanleonov I like your thinking. I would propose to have this discussed for another PR .This is meant to be reactionary to something that already happened and we are trying to recover from.

AppController/djinn.rb
```diff
@@ -5578,18 +5578,25 @@ def check_services_load(_version_key, delta_appservers)
...
-          return -1
+          Djinn.log_debug("Database at #{host} has normalized load of #{load_now}.")
+          high_load = load_now if load_now > high_load
         end
       }
     }
+
+    # The euristic implied here removes one AppServer per extra unit of
+    # normalized load on the most loaded datastore.
+    delta_appservers = -1 * high_load.ceil if high_load > 0
```
jeanleonov: This looks dangerous to me. One absolute number (max load_avg_norm) lineally controls how many appservers should be terminated. The same level of overload on of DB nodes can kill 5% of appservers on big deployment OR 90% of appservers on small deployment.
This feature will still work properly on some of deployments so I'd agree we can try it but with feature flag (e.g.: DOWNSLACE_ON_SERVICE_OVERLOAD; default=False).

jeanleonov: Or this PR can go as it is, but before release I'd vote for either making it disabled by default or implementing scaling mechanism based on relative numbers.

obino: Shoot, never mind! I understand what you are saying: the logic was meant to be the *extra* load on top of the MAX_LOAD_AVG we have as threashold. Let me fix that.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/5519/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/5522/

AppController/djinn.rb
jeanleonov: @obino thanks for fixing that. I didn't notice that issue though=)
My initial concern is still in place. Let's consider two running deployments:

 1. Min appservers: 1, current_appservers: 10, there are 12 database nodes (average loadavg_norm: 2, max loadavg_norm 11) => downscaling 9 appservers. So instead of 10 we will have 1 appserver.
 2. Min appservers: 1, current_appservers: 40, there are 12 database nodes (average loadavg_norm: 2, max loadavg_norm 11) => downscaling 9 appservers. So instead of 40 we will have 31 appserver.

Though overload is the same, in the first case we kill 90% of appservers. Isn't it dangerous?
I would say we need to use kind of fuzzy logic here:
 - Services are a bit overloaded => downscale appservers a bit (10%).
 - Services are hardly overloaded => downscale a lot (50-70%). If it didn't help in 10-20 minutes, downscaling will repeat.

Again, my example is kind of corner case.. I'm still fine with this PR being merged as it is now to make it easier to test with some deployment(s), but before release I think we need to make this feature disabled by default or make the logic more agile.

obino: @jeanleonov if the load shot up to 11 without us detecting the above 2 before it, I think we are in deep trouble (that number is already smoothed), hence I don't see the case 1) as an overreaction. Ideally we would never see the system drifting too much above the MAX_LOAD_AVG, this is the idea of this PR.  I think your ideas should be more part of the normal autoscaling policy (that is we should eventually improve it) to take in account what is the service load contribution of each application, and allow the individual app to scale up and down differently based on that.

AppController/djinn.rb
```diff
@@ -5242,6 +5269,49 @@ def get_hosts_for_version(version_key)
...
+  def check_services_load(_version_key, num_appservers)
+    # Let's find the most loaded DB node and its load.
+    high_load = 0.0
+    get_all_database_nodes.each { |host|
+      @state_change_lock.synchronize {
+        @cluster_stats.each { |node|
+          next if node['private_ip'] != host
+
+          load_now = Float(node['loadavg']['last_1_min']) / node['cpu']['count']
+          if load_now >= MAX_LOAD_AVG
```
cdonati: I think this check can be omitted, and the inner assignment can be kept.

AppController/djinn.rb
```diff
@@ -5040,78 +5022,123 @@ def get_scaling_info_for_version(version_key)
...
+    # Make sure we don't go over the maximum number of AppServers
+    # indicated for this application.
+    if num_appservers + appservers_to_scale > max
+      appservers_to_scale = max - num_appservers
+      if appservers_to_scale.zero?
+        Djinn.log_debug("#{version_key} has reached the maximum number of allowed " \
+                        "AppServers (#{max}).")
+      end
+    elsif num_appservers + appservers_to_scale < min
+      appservers_to_scale = num_appservers - min
```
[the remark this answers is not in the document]
obino: Right! THanks,

AppController/djinn.rb
```diff
@@ -5040,78 +5022,123 @@ def get_scaling_info_for_version(version_key)
...
-    else
-      Djinn.log_debug("The deployment is within the desired range of load " \
-                      "for #{version_key} - Advising that there is no need " \
-                      "to scale currently.")
-      return 0
     end
+
+    # Finally we have to ensure we have at least the minimum number of
+    # AppServers required.
+    if num_appservers + appservers_to_scale < min
```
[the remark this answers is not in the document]
obino: Yes they are very similar. I like to have them both because they are inexpensive and the former checks that the application will not scale below the minimum, while this one ensures that the applications has a minimum sets of AppServers. I can remove the 5112 if you like it better, but I prefer to have the last check to be the insurance the application will stay up.

PR #2805, opened 2018-06-19
AppDashboard incorrect credential escaping

sjones4:
Remove incorrect use of cgi.escape for passwords when creating new users or changing passwords via the dashboard. The email/password should only be escaped when the values are for use in a view, and in this case should also escape the double quote character (so < > & " are all escaped)

To reproduce the issue, create a user via the dashboard with the password `p&ssword`, and then login using the password `p&amp;ssword`.  The quote character not being escaped can be tested by entering the value `" style="background: lime` in the email field (for example)

You can work around this issue by using the "appscale" command to create users.

### DIFF
--- a/AppDashboard/dashboard.py
+++ b/AppDashboard/dashboard.py
@@ -171,21 +171,21 @@ class NewUserPage(AppDashboard):
     """
     users = {}
     error_msgs = {}
-    users['email'] = cgi.escape(self.request.get('user_email'))
+    users['email'] = self.request.get('user_email')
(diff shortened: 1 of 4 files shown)


sjones4:
There was missing output escaping on the login form email address. I've updated the pull request to perform escaping in the templates and removed use of cgi module. The build with the update is:

https://ocd.appscale.com:8080/job/Daily%20Build/5162/

sjones4:
Fixes appscale#1737

PR #2810, opened 2018-06-21
Remove sshguard

obino:
sshguard is usually not needed with AppScale since instances do not have
password enabled, and it creates issues since the various nodes need (at
this time) to be able to ssh into the head node.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5145/

PR #2811, opened 2018-06-22
Use yaml.safe_load rather than yaml.load

sjones4:
Switch to yaml.safe_load from yaml.load.


sjones4:
https://ocd.appscale.com:8080/job/Daily%20Build/5142/

PR #2816, opened 2018-06-29
Detect missing row when applying transaction

cdonati:
This allows the service to return a different error when the client tries to extend the lease of a task that doesn't exist. Previously, it would return an error indicating that the lease has expired.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5218/

PR #2821, opened 2018-07-03
Reject lease modifications that specify wrong ETA

### LINKED ISSUE #2819: Pull queue clients can modify tasks they don't own
Clients should not be able to update a lease unless they provide the correct ETA in the lease modification request.

cdonati:
This prevents a worker that does not currently hold a lease from updating the task lease. Without this check, it may shorten the lease for another holder.

This also fixes a bug where task ETAs were being encoded with the wrong precision for protobuf responses.

Resolves #2819.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5180/

PR #2824, opened 2018-07-16
Reduce diffs with the 1.8.0 SDK (Part 2)

cdonati:
This is the second (of two) pull requests removing the differences between the 1.8.0 SDK and our code base. Maintaining fewer diffs will make it easier to upgrade SDK support in the future.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5211/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5212/

obino:
@cdonati I did add a demo requested here, but the demo is the build I assume, is that correct? anything in particular we should look up?

Issue #2827, opened 2018-07-19
Fix snapshots caching in Hermes


jeanleonov:
Due to wrong usage of initialize method of RequestHandler Hermes didn't cache stats snapshots.

Issue #2803, opened 2018-06-14
Hermes Overload


whoarethebritons:
If we have many many machines (in my test about 150), Hermes gets overloaded by requests from the appmanagerserver to get statistics for running AppServers and appmanagerserver will not receive a response which prevents the deployment from scaling.

jeanleonov:
@whoarethebritons let me know if you think we need to do additional testing of the fix.

PR #2828, opened 2018-07-19
Fix snapshots caching in Hermes

jeanleonov:
Resolves https://github.com/AppScale/appscale/issues/2827
https://ocd.appscale.com:8080/job/Daily%20Build/5222/


Hermes/appscale/hermes/stats/handlers.py
```diff
@@ -17,10 +17,25 @@ class CurrentStatsHandler(RequestHandler):
...
+      cache_container: a list containing a single element - cached snapshot.
+    """
     self._stats_source = source
     self._default_include_lists = default_include_lists
-    self._snapshot = None
+    self._cache_container = cache_container
+
+  @property
+  def _cached_snapshot(self):
+    return self._cache_container[0]
```
cdonati: Why is self._cache_container a list?

jeanleonov: RequestHandler instance is created for every single request, so cache container should be passed to initialize method.
I was thinking about defining dedicated class with a single attribute, but it looks like over-engineering. So list object seems to be the simplest solution.

jeanleonov:
This PR resolved https://github.com/AppScale/appscale/issues/2803.

PR #2832, opened 2018-07-24
Add appscale-admin restart command

### LINKED ISSUE #2829: No easy way to restart datastore servers
Previously, it was easy to run `monit restart datastore_server-4000` to restart a datastore server process when debugging.

Now that we don't use monit to manage datastore processes, we should have something like `appscale-admin restart datastore-4000`.

cdonati:
This gives the ServiceManager a unix socket at /var/run/appscale/service_manager.sock that accepts management requests.

This allows the user to immediately restart datastore servers without waiting for the ServiceManager to periodically check and fulfill assignments.

Resolves #2829.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5290/
https://ocd.appscale.com:8080/job/Tools-test/3554/

PR #2834, opened 2018-07-27
Create additional table for tag queries

cdonati:
The previous method of using secondary indexes on the ETA index table was inefficient when grouping by tag and ordering by ETA.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5399/

PR #2838, opened 2018-08-02
Force the push worker to time out stuck tasks

cdonati:
The original problem was that the HTTP client would hang indefinitely every once in a while instead of timing out after 10 minutes. Every time this happened, the celery worker's capacity for concurrent tasks was reduced by 1.

This branch prevents that from happening by wrapping the whole task request in a timeout so that even if the HTTP client hangs indefinitely for some reason, the celery worker will abort and retry after 13 min.

This may not be a problem any longer, but I don't think it hurts to have in master as a safety measure. I've had this branch sitting around for about a year, so I decided to submit it rather than delete it. When we change out the push queue implementation, we should no longer need this.

### DIFF
--- a/AppTaskQueue/appscale/taskqueue/push_worker.py
+++ b/AppTaskQueue/appscale/taskqueue/push_worker.py
@@ -11,6 +11,7 @@ from appscale.common.unpackaged import APPSCALE_PYTHON_APPSERVER
 from celery.utils.log import get_task_logger
 from eventlet.green import httplib
 from eventlet.green.httplib import BadStatusLine
+from eventlet.timeout import Timeout as EventletTimeout
 from socket import error as SocketError
 from urlparse import urlparse
 from .distributed_tq import TaskName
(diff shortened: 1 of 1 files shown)


AppTaskQueue/appscale/taskqueue/push_worker.py
```diff
@@ -92,132 +96,143 @@ def execute_task(task, headers, args):
...
-                 'Retrying in {wait} secs.'.format(status=response.status,
-                                                   task=args['task_name'],
-                                                   wait=wait_time))
-      logger.warning(message)
-      raise task.retry(countdown=wait_time)
+  except EventletTimeout as thrown_timeout:
+    if thrown_timeout != timeout:
+      raise
+
+    logger.exception('Task timed out')
```
sjones4: Are there any details for the task that should be logged here?

cdonati: Good point. I added the task ID to the log output.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5448/

PR #2839, opened 2018-08-02
Use real AppScale Cloud Storage when configured

cdonati:
By default, the dispatcher for non-Java runtimes uses the blobstore backend for GCS operations. This change allows users to use a real GCS service instead.


cdonati:
@jeanleonov I agree. Good call.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5459/

cdonati:
I'm going to close this for now since there are some problems with it (I didn't sufficiently test it before submitting it) and it's not a high priority.

PR #2840, opened 2018-08-02
Try to determine which user to login as

cdonati:
This allows logging in even when the default username is not "ubuntu" (on Debian, for instance).


cdonati:
This stretch-based build failed when the head node tried to initialize other nodes: https://ocd.appscale.com:8080/job/Hawkeye/15295/
This is the result when including this PR: https://ocd.appscale.com:8080/job/Hawkeye/15308/ (still fails, but nodes initialize correctly).

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5280/

PR #2842, opened 2018-08-06
Handle ID separator in key name

### LINKED ISSUE #2841: Queries can fail for entities with ":" in the key name
When parsing index entries for entities that contain ":" in the key name, the datastore does not know which part of the index entry represents the kind name and which represents the entity key name.

cdonati:
This change assumes that everything before the first ':' is the kind name and everything after is the key name (or ID). Although the Cloud Datastore API allows ':' in the kind name, we forbid it during puts so that we can make that assumption during queries.

Resolves #2841.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5278/
https://ocd.appscale.com:8080/job/Euca-Faststart/2633/

cdonati:
Demo
End-to-end test (https://github.com/AppScale/appscale/pull/2843) against master: https://ocd.appscale.com:8080/view/All/job/Datastore/30/
Against this branch: https://ocd.appscale.com:8080/view/All/job/Datastore/31/

PR #2853, opened 2018-08-10
Use hmac helper for authentication check

sjones4:
Switch to the hmac module for admin server appscale secret authentication check.


sjones4:
https://ocd.appscale.com:8080/job/Daily%20Build/5292/

PR #2854, opened 2018-08-20
Use security token when making Monit API calls

cdonati:
The cookie field was introduced in Monit 5.20.

This also reduces the number of Monit CLI calls. Monit was recently patched in Ubuntu to address https://security-tracker.debian.org/tracker/CVE-2016-7067. However, the CLI client was not modified to match the new restrictions on the HTTP interface.

https://bugs.launchpad.net/ubuntu/+source/monit/+bug/1786910

These changes are not exhaustive. There are probably more invocations of the monit script that are ineffectual. This just changes enough to pass most of the tests.

### DIFF
--- a/AppManager/app_manager_server.py
+++ b/AppManager/app_manager_server.py
@@ -8,7 +8,6 @@ import os
 import random
 import re
 import signal
-import urllib
 import urllib2
 
 import psutil
@@ -19,7 +18,6 @@ from kazoo.exceptions import NodeExistsError, NoNodeError
 from tornado import gen
 from tornado.escape import json_decode
 from tornado.httpclient import AsyncHTTPClient
-from tornado.httpclient import HTTPClient
 from tornado.httpclient import HTTPError
 from tornado.ioloop import IOLoop, PeriodicCallback
 from tornado.options import options
@@ -48,7 +46,6 @@ from appscale.common import (
   constants,
   file_io,
   monit_app_configuration,
-  monit_interface,
   misc
 )
 from appscale.common.constants import HTTPCodes, MonitStates, VAR_DIR
@@ -56,7 +53,9 @@ from appscale.common.constants import VERSION_PATH_SEPARATOR
 from appscale.common.deployment_config import ConfigInaccessible
 from appscale.common.deployment_config import DeploymentConfig
 from appscale.common.monit_interface import MonitOperator
+from appscale.common.monit_interface import MonitUnavailable
 from appscale.common.monit_interface import ProcessNotFound
+from appscale.common.retrying import retry
 from appscale.hermes.constants import HERMES_PORT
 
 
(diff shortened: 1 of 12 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5319/

common/appscale/common/monit_interface.py
```diff
@@ -248,14 +257,74 @@ def send_command(self, process_name, command):
...
-      yield self._async_client.fetch(process_url, method='POST', body=payload)
+      yield self._async_client.fetch(
+        process_url, method='POST', body=urllib.urlencode(params),
+        headers=headers)
     except HTTPError as error:
-      if error.code == 404:
+      if error.code == httplib.FORBIDDEN:
+        # Retrieve CSRF token (introduced in Monit 5.20).
+        response = yield self._async_client.fetch(process_url)
+        self._csrf_token = re.search('securitytoken=([a-zA-Z0-9]+);',
```
scragraham: Not sure if we'll hit it with monit, but the Cookie header syntax doesn't require a semicolon, so its possible (probably unlikely) to see:
```
Cookie: securitytoken=12345
```
too bad the client doesn't have support for cookie handling/parsing. oh well.

AppManager/app_manager_server.py
```diff
@@ -501,7 +478,9 @@ def unmonitor_and_terminate(watch):
     watch: A string specifying the Monit entry.
   """
   try:
-    unmonitor(watch)
+    monit_retry = retry(max_retries=5, retry_on_exception=MonitUnavailable)
```
jeanleonov: Default backoff sequence is 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8,... If you set `max_retries=5` sum of pauses between retries will be `6.2s`. Is this usually enough for monit to become available again?
`retrying_timeout=30` or 60 could be an alternative.

jeanleonov:
I suspect that failures in previous build was not transient. Trying one more time:
https://ocd.appscale.com:8080/job/Daily%20Build/5322/

common/appscale/common/monit_interface.py
cdonati: Good point. Using Python's cookie library is more robust.

common/appscale/common/monit_interface.py
```diff
@@ -408,3 +410,15 @@ def _reload(self):
...
+  def _parse_security_token(http_response):
+    """ Extracts the security token from an HTTP response.
+
+    Args:
+      A tornado.httpclient.HTTPResponse object.
+    Returns:
+      A string containing the security token.
+    """
+    cookie = SimpleCookie(http_response['Set-Cookie'])
+    return cookie['securitytoken'].value
```
scragraham: KeyError can be thrown here, if there are no cookies in the response, and if securitytoken isn't a cookie. I think the semantics should be that we return None if the token isn't found:
```
try:
    cookie = SimpleCookie(http_response['Set-Cookie'])
    return cookie['securitytoken'].value
except KeyError:
    return None
```

cdonati:
Build for df4df1e:
https://ocd.appscale.com:8080/job/Daily%20Build/5324/
https://ocd.appscale.com:8080/job/Datastore/53/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5326/

PR #2851, opened 2018-08-09
Include trailing value delimiter in range queries

cdonati:
Without this, a query that specifies an empty value can exceed the range of relevant index entries.

For example, the start value could look like
`guestbook\x00\x00Greeting\x00content\x00`
and the end value like
`guestbook\x00\x00Greeting\x00content\x00\xff...`

This ensures the start value looks like
`guestbook\x00\x00Greeting\x00content\x00\x00`
and the end value like
`guestbook\x00\x00Greeting\x00content\x00\x00\xff...`

Resolves #2852.

PR #2855, opened 2018-08-21
Add test for merge join query with null value

cdonati:
Accompanies #2851.


cdonati:
https://ocd.appscale.com:8080/job/Datastore/104/

PR #2861, opened 2018-08-31
Give a list of instances to terminate

obino:
Previously we were terminating one instance at a time when downscaling.
This commit pass a list to the infrastructure manager.


AppController/djinn.rb
```diff
@@ -5224,62 +5226,65 @@ def scale_down_instances
...
+  #                    terminated and removed from this deployment.
+  def terminate_nodes_from_deployment(nodes_to_remove)
+    if nodes_to_remove.nil? || nodes_to_remove.emtpy?
+      Djinn.log_info("Tried to scale down but couldn't find a node to remove.")
       return 0
     end
 
+    # Let's get the list of IDs of the instances to terminate.
+    instances_id = []
+    nodes_to_remove.each{ |node| instances_id << node.instance_id }
```
cdonati: `instance_ids` would be a more idiomatic name. This is also a good fit for map:
`instance_ids = nodes_to_remove.map{ |node| node.instance_id }`

AppController/djinn.rb
```diff
@@ -5224,62 +5226,64 @@ def scale_down_instances
...
   # Args:
-  #   node_to_remove: A node instance, to be terminated and removed
-  #     from this deployment.
-  def terminate_node_from_deployment(node_to_remove)
-    if node_to_remove.nil?
-      Djinn.log_warn("Tried to scale down but couldn't find a node to remove.")
+  #   nodes_to_remove: A node instance, or an Array of nodes to be
+  #                    terminated and removed from this deployment.
+  def terminate_nodes_from_deployment(nodes_to_remove)
+    if nodes_to_remove.nil? || nodes_to_remove.emtpy?
```
sjones4: `emtpy?`

AppController/djinn.rb
```diff
@@ -5240,62 +5242,64 @@ def scale_down_instances
...
       return 0
     end
 
+    # Let's get the list of IDs of the instances to terminate.
+    instance_ids = nodes_to_remove.map{ |node| node.instance_id }
+
     imc = InfrastructureManagerClient.new(@@secret)
     begin
-      imc.terminate_instances(@options, node_to_remove.instance_id)
+      imc.terminate_instances(@options, instance_ids)
```
sjones4: I think error handling would need to be changed to safely allow terminating multiple instances.

When terminating instances one at a time failures are handled instance by instance. When terminating a group of instances it looks like the failure semantics are agent specific (i.e. azure/gce may terminate some instances and fail) and the state of each instance would need to be verified after a termination error.

PR #2862, opened 2018-08-31
Monit random csrf token and switch to appscale start/stop scripts

sjones4:
The common monit interface now uses a random CSRF token rather than one obtained from monit.

Retries for monit commands are updated, we now retry for all cases other than `ProcessNotFound` by default and will retry on that failure for `start` actions when there may be a short period before a previous `reload` completes and the process is recognized.

Start/stop commands are updated to not output error traces for some expected failures.

Monit reloading can now be performed on a separate thread.

### DIFF
--- a/AdminServer/appscale/admin/stop_services.py
+++ b/AdminServer/appscale/admin/stop_services.py
@@ -2,11 +2,13 @@
 import argparse
 import logging
 import socket
+import sys
 import time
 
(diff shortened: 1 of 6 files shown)


sjones4:
https://ocd.appscale.com:8080/job/Daily%20Build/5360/

cdonati:
I think the successful build is sufficient as a demo for this one.

PR #2863, opened 2018-09-05
Postgres connection pool with autoreconnect

jeanleonov:
It helps TaskQueue to deal with restarted Postgres server or with any other problems which can cause connection issues.


cdonati:
It seems to be a common pattern to put the connection back in the pool when the client is done using it: https://medium.com/@thegavrikstory/manage-raw-database-connection-pool-in-flask-b11e50cbad3

Is this no longer needed? If getconn pops a connection from the pool, there doesn't seem to be a benefit from using the pool if the client does not return it.

AppTaskQueue/appscale/taskqueue/queue.py
```diff
@@ -248,24 +248,26 @@ class PostgresPullQueue(Queue):
...
     Args:
       queue_info: A dictionary containing queue info.
       app: A string containing the application ID.
-      pg_connection: A psycopg2 connection to PostgreSQL.
+      pg_connection_pool: A psycopg2 connection pool.
     """
     from psycopg2 import IntegrityError  # Import psycopg2 lazily
     super(PostgresPullQueue, self).__init__(queue_info, app)
-    self.pg_connection = pg_connection
+    self.connection_key = self.app
```
jeanleonov: @cdonati , this PR is aimed to make Taskqueue resistant to postgres connection problems.
Initially I tried to implement connection wrapper which would do reconnection automatically, but that solution wouldn't work with concurrent requests.
Then I decided that client (PostgresPullQueue) shouldn't hold connection or connection wrapped on its own, but it should always ask workable connection from external object (ConnectionPool).

This modified connection pool takes care about reconnect.

ConnectionPool is created for every project. All queues in the project uses single connection (`self.connection_key` is the way how PostgresPullQueue refers to this connection). When the project is deleted, all connections (single connection in this case) are closed.

cdonati: That makes sense. Thanks for the explanation.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5649/
@sjones4 thanks for raising incompleteness of the previous solution.
Let me know if the latest changes make sense to you.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5649/
@sjones4 thanks for raising incompleteness of the previous solution.
Let me know if the latest changes make sense to you.

PR #2866, opened 2018-09-05
Add hmac.compare_digest fallback

cdonati:
The function was introduced in Python 2.7.7, so it is not available on Trusty.


cdonati:
Demo
master: https://ocd.appscale.com:8080/job/Hawkeye/15605/
branch: https://ocd.appscale.com:8080/job/Hawkeye/15609/

PR #2871, opened 2018-09-13
Route instances as soon as they are registered

cdonati:
This moves the instance routing work from the controller's polling-driven approach to ZooKeeper watches managed by the AdminServer on load balancers.

Future work in this area includes:
 - Managing HAProxy processes with cgroups rather than Monit
 - Taking the instance's revision ID into account when updating routes

The most significant change from the previous work is that the manager no longer writes a base configuration and version blocks to the file system. It just writes the full configuration file.

In my opinion, this makes the code a lot simpler. Since we are using custom HAProxy processes instead of the systemd-managed service, I doubt anyone would want to use "sites-enabled" directory to create custom routes that aren't managed by AppScale.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5820/

PR #2873, opened 2018-09-20
Application scaling for version configuration

sjones4:
The admin server is updated to accept manual scaling settings on version creation and to validate the manual or automatic scaling provided.  Version `patch` is updated to allow automatic scaling min/max to be configured.

The controller uses the scaling settings for the min/max instances for an application (version)

A dependency on `jsonschema` is added for use validating a version. The (json) schema resource is packaged with the python code.


AdminServer/appscale/admin/version.schema.json
```diff
@@ -0,0 +1,173 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "AppEngine application version",
+    "description": "An AppEngine application version - apps.services.versions.Version",
+    "definitions": {
+        "duration": {
```
cdonati: This is really cool!

AdminServer/appscale/admin/__init__.py
```diff
@@ -886,8 +893,16 @@ def version_from_payload(self):
       raise CustomHTTPError(HTTPCodes.BAD_REQUEST, message=message)
 
     desired_fields = update_mask.split(',')
-    supported_fields = {'appscaleExtensions.httpPort',
-                        'appscaleExtensions.httpsPort'}
+    supported_fields = {
+      'appscaleExtensions.httpPort',
+      'appscaleExtensions.httpsPort',
+      'automaticScaling.standard_scheduler_settings.max_instances',
```
cdonati: Should this be `standardSchedulerSettings`?

AdminServer/appscale/admin/__init__.py
```diff
@@ -886,8 +893,16 @@ def version_from_payload(self):
...
+      'automaticScaling.standard_scheduler_settings.min_instances'}
+    mapped_fields = {
+      'automaticScaling.standard_scheduler_settings.max_instances':
```
cdonati: What's the reason for converting case on `max_instances`?

AdminServer/setup.py
```diff
@@ -38,6 +39,8 @@
   packages=['appscale',
             'appscale.admin',
             'appscale.admin.instance_manager'],
+  package_data={'': ['*.json']},
```
cdonati: Usually I see the first argument here specifying the package that should include the data. Is the behavior for an empty string documented somewhere?

AdminServer/setup.py
```diff
@@ -38,6 +39,8 @@
...
+  include_package_data=True,
```
cdonati: The only documentation I was able to find for this argument was in reference to a `Manifest.in` file. Does it have a different behavior here?

AdminServer/appscale/admin/version.schema.json
```diff
@@ -0,0 +1,173 @@
...
+            "type": "string",
+            "pattern": "^[+-]?[0-9]{1,12}(?:.[+-]?[0-9]{1,9})?s$"
+        }
+    },
+    "type": "object",
+    "properties": {
+        "automaticScaling": {
```
cdonati: Is there a way to mark these keys as exclusive so that a client cannot define both automaticScaling and manualScaling?

AdminServer/appscale/admin/__init__.py
```diff
@@ -937,6 +954,12 @@ def update_version(self, project_id, service_id, version_id, new_fields):
     version = json.loads(version_json)
     new_ports = utils.assign_ports(version, new_fields, self.zk_client)
     version['appscaleExtensions'].update(new_ports)
+    if not 'manualScaling' in version:
```
cdonati: If a user has manual scaling defined for their version and they decide to change to automatic scaling, it looks like the patch call would succeed without any side effect.

AdminServer/appscale/admin/__init__.py
sjones4: My understanding is that these `automaticScaling` names should be as defined in the api doc here:

https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services.versions/patch

AdminServer/appscale/admin/__init__.py
sjones4: This is a difference in the syntax between the REST api names and names in the version resource.

AdminServer/setup.py
sjones4: IIRC this means any of the previously defined packages. We only have resources in one package so it would be easy to change. It seemed simpler as is (no need to list packages again) but now you mention it the behaviour is not obvious.

AdminServer/setup.py
sjones4: I think it is the same, i.e. that the package data should be distributed.

AdminServer/appscale/admin/__init__.py
cdonati: I think what's happening is that Google's server does some automatic case conversion that applies to _all_ keys. After testing, the API accepts both `automaticScaling.standardSchedulerSettings.maxInstances` and `automatic_scaling.standard_scheduler_settings.max_instances` in the payload of the patch.

It's annoying that their documentation isn't consistent, but I'm fine with what you have here.

sjones4:
Schema validation for versions is updated so scaling settings are optional but exclusive.

Updating automatic scaling settings for a basic scaling version now fails with an error response.

sjones4:
The pull request was updated to address a couple of issues raised by @cdonati , the build with these updates is:

https://ocd.appscale.com:8080/job/Daily%20Build/5430/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5475/

PR #2883, opened 2018-10-04
Move instance fulfillment work to InstanceManager

### LINKED ISSUE #2882: Instance fulfillment time is slowed by duty cycle
Since the compute machine controllers poll for changes to the head node's state, it can take longer than necessary for them to start making AppManager calls to fulfill instance assignments.

This can be improved by triggering fulfillment-related work based on a ZooKeeper watch on the controller state node.

cdonati:
This improves fulfillment times by watching for changes to the controller state node. The compute controller used to poll for changes every ~10 seconds.

Now, the InstanceManager starts work immediately when the state node or version node is updated.

The first two commits just move code around and don't change functionality. The third commit contains the substantial changes.

Resolves #2882.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5457/

AdminServer/appscale/admin/instance_manager/instance_manager.py
```diff
@@ -0,0 +1,610 @@
...
+      # If there are any instances running an older revision, restart them.
+      for instance in version_instances:
+        if instance.revision_key != version.revision_key:
+          yield self._stop_app_instance(instance)
+          yield self._start_instance(version, instance.port)
+
+  @gen.coroutine
+  def _fulfillment_loop(self):
+    """ Continually tries to fulfill assignments. """
+    while True:
```
jeanleonov: I see benefits of using AsyncEvent with endless loop blocked by the event.
But it took ~20 minutes for me to understand how it works=)
Asynchronous code is complicated by its nature. I don't have strong feeling that my approach is better, but anyway I suggest to consider it.

So currently, from what I understood, the flow is (I'll mention only relevant steps):
 1. Initialization: go through assignments declared in ZK, add loop-unlocking callback to every version.
 2. Do idle iterations every GROOMING_INTERVAL
 3. Version is updated in ZK
 4. Version manager is notified about it and does update
 5. Version manager calls loop-unlocking callback
 6. _fulfill_assignments locks loop again
 7. _fulfill_assignments ensures that all versions will unlock endless loop when new revision arrives.

Would _fulfill_assignments be "notified" when **new** version is deployed? (read: Is AC state node updated when new version is deployed)?

My concern is that it looks like InstanceManager is simply subscribed to version updates. This subscription is done by endless loop + asyncevent + callback, which are all comparably complicated tools.

Wouldn't it be better to implement subscription mechanism in GlobalProject manager to raise updates from version level to global project and then to anythink interested in any updates inside projects?
In this case InstanceManager would subscribe once plus add periodicalcallback with GROOMING_INTERVAL.

cdonati: @jeanleonov First of all, thanks for the speedy review of these extensive changes! I agree that this could be a lot clearer. I haven't used async events much, so I'm still trying to learn the best way to use them.

There's a slight difference difference between your description and what I've intended. The machine's assignments come from `/appcontroller/state`, but the version details come from `...services/<service-id>/versions/<version-id>`.

The InstanceManager does not care about changes to versions that aren't assigned to it. So when a new version is deployed, the fulfillment work should not be triggered yet. If the scheduler later assigns some of that new version's instances to the relevant machine, it would update `/appcontroller/state` and trigger the fulfillment work. Similarly, the InstanceManager does not need to perform fulfillment work if a version that it is not assigned to run gets a new revision or deleted, etc.

I am 100% in favor of code readability over efficiency here, but I don't want to mislead the reader into thinking that the InstanceManager cares about _any_ changes to the configured projects. I'll try to make some changes to clarify that fulfillment work is triggered by any of 3 conditions (grooming interval, update to the machine's assignments, and an update to an assigned version).

To address your last concern, the subscription to version changes is idempotent. I wanted to avoid a situation where a version object is deleted and later recreated, losing the callback. I'm going to move the subscription work away from the fulfillment loop, though. I agree that it's a bit weird right now.

Your idea about subscribing directly to the projects manager is a good one. I'll give it a shot.

cdonati:
@jeanleonov Let me know if ec29170 makes the cycle and the relevant events clearer.

jeanleonov:
@cdonati I was focused on my concerns but did have to mention that the work here looks really impressive: it makes appmanager much more pythonic and gets use of eventdriven architecture!

Though I'll anyway continue sharing my thoughts=)
Thanks for clarification you've provided. It's hard to keep all workflows and interactions in mind, but now picture looks much more clear to me.
Your comments in code makes it even more clear.

Sorry for going back and force. I think that restart logic should be decoupled from fulfillment as it's actually different work. But relation between InstanceManager and VersionManager should be reversed.
Instance manager is a singleton so it should be easy to access it from VersionManager. Wouldn't be a clearer approach to invoke InstanceManager.instance().restart_version(version) from VersionManager, it can be invoked asynchronously if needed, but directly from VersionManager.
Other triggers can call fulfillment directly (from periodical task and AC state change).
Though if there will be two methods managing instances you may need to wrap those with tornado lock.

I'm impressed by the work done, it already looks good, so I won't push on addressing this comment, but want to know what do you think about it.

cdonati:
@jeanleonov I agree that the restart work is different from the fulfillment work. I'll make them separate cycles for clarity.

I'm not sure I agree about a VersionManager performing the work. At the moment, a `Version` is behaving as a passive resource that keeps track of version details. The `InstanceManager` uses the projects manager and its related objects just as a convenient lookup. To me, it's clearer and easier to debug if all the event triggers are listed in the `InstanceManager`.

Otherwise, a reader would need to reason about multiple interactions between different active managers. For example, I would probably need to introduce an additional manager for the login property for consistency. I would also need to create the InstanceManager before the project resources, which would be awkward because the `InstanceManager` uses the project resources.

cdonati:
Let me know if bc92bfb is a step in the right direction. I took your advice and split the instance manager's work into 3 different categories: fulfillment, config updates, and health checks. In the process, I was able to remove the async Event and infinite loop because the ZooKeeper watches can now call the required functions directly.

AdminServer/appscale/admin/instance_manager/instance_manager.py
```diff
@@ -0,0 +1,660 @@
...
+from appscale.admin.instance_manager.utils import setup_logrotate
+from appscale.common import appscale_info, monit_app_configuration
+from appscale.common.async_retrying import retry_data_watch_coroutine
+from appscale.common.constants import (
+  APPS_PATH, GO, HTTPCodes, JAVA, MonitStates, PHP, PYTHON27, VAR_DIR,
+  VERSION_PATH_SEPARATOR)
+from appscale.common.monit_interface import DEFAULT_RETRIES, ProcessNotFound
+from appscale.common.retrying import retry
+
+logger = logging.getLogger('appscale-instance-manager')
```
sjones4: I'm not clear on the use of logging vs. logger, may be worth a check to ensure it is as intended.

cdonati: `logging` is a shortcut to the root logger. `logger` here is a custom logger that can have a different format and level than the root one.

Some libraries use the root logger to log messages, and we were recently using `logging` everywhere. In one case, this meant that when we enabled debug-level logging, we got super verbose debug logs from the kazoo library.

Recently, we've been using a custom logger to avoid this problem. The references to `logging` in this module were left over from when I moved the code from the app_manager_server script. In this case, it won't make a difference since they both happen to have the same level and format, and the user doesn't yet have a way to change it.

However, it's helpful that you pointed out those cases since we should be more consistent here, and we will likely make this more configurable. I've created a new issue for it (https://github.com/AppScale/appscale/issues/2893) that I will address for the whole appscale-admin package after this PR goes in.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5505/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5532/

cdonati:
Demo:

Deploying a service for the first time on master:
```
AdminServer: 04:13:38 200 POST /v1/apps/hawkeyejava/services/default/versions
Head AC:     04:13:46 Adding a new AppServer on 10.10.4.12 for hawkeyejava_default_v1
Head AC:     04:13:50 backup_appcontroller_state: updated state
Compute AC:  04:13:53 Reload state (got new assignment)
Compute AC:  04:13:53 Starting hawkeyejava_default_v1 on 10.10.4.12:20001
AppManager:  04:13:58 200 POST /versions/hawkeyejava_default_v1
```
It took __3 seconds__ from the time that the head AC updated its state to the process of starting an instance on the compute machine. This was relatively quick for master. It could be up to 10 seconds due to the compute AC's duty cycle.

---

Deploying a service for the first time on this branch:
```
AdminServer: 15:55:34 200 POST /v1/apps/hawkeyejava/services/default/versions
Head AC:     15:55:44 Adding a new AppServer on 10.10.9.193 for hawkeyejava_default_v1
Head AC:     15:55:47,418 backup_appcontroller_state: updated state
AppManager:  15:55:47,423 New assignments: {... u'hawkeyejava_default_v1': [-1]}
AppManager:  15:55:49 Starting Version<hawkeyejava_default_v1, runtime=java>:20001 
AppManager:  15:55:52 http://10.10.9.193:20001/_ah/health_check returned
```
It took __5 milliseconds__ from the time that the head AC updated its state to the process of starting an instance on the compute machine.

---

Updating a service with a new revision on master:
```
AdminServer: 04:45:16 POST /v1/apps/validity-test/services/default/versions
Head AC:     04:45:16 Remove old AppServers for ["validity-test_default_v1"]
Head AC:     04:45:21 backup_appcontroller_state: updated state
Compute AC:  04:45:27 Reload state : ..."validity-test_default_v1"=>{"appservers"=>[]...
Compute AC:  04:45:28 Terminate the following AppServer: validity-test_default_v1:20001
AppManager:  04:45:29 200 DELETE /versions/validity-test_default_v1/20001
Head AC:     04:45:21 Adding a new AppServer on 10.10.5.5 for validity-test_default_v1
Head AC:     04:45:21 backup_appcontroller_state: updated state
Compute AC:  04:45:37 Reload state : ..."validity-test_default_v1"=>{"appservers"=>["10.10.5.5:-1"...
Compute AC:  04:45:37 Starting validity-test_default_v1 on 10.10.5.5:20001
```
AppManager:  04:45:40 200 POST /versions/validity-test_default_v1
It took __24 seconds__ from the revision update to the process of starting the new revision.

---

Updating a service with a new revision with this branch:
```
AdminServer: 16:29:12,270 200 POST /v1/apps/validity-test/services/default/versions
AppManager:  16:29:12,260 New revision for version: validity-test_default_v1
AppManager:  16:29:12,261 Stopping <Instance: validity-test_default_v1_1540312115507:20002>
AppManager:  16:29:13,151 Starting Version<validity-test_default_v1, runtime=python27>:20002
```
It took __891 milliseconds__ from the revision update to the process of starting the new revision.

PR #2885, opened 2018-10-05
Handle datastore connection errors with TQ server

cdonati:
The TaskQueue server does not use an API Proxy like application runtimes do, so it cannot expect the usual errors from using the client libraries.

### DIFF
--- a/AppTaskQueue/appscale/taskqueue/push_worker.py
+++ b/AppTaskQueue/appscale/taskqueue/push_worker.py
@@ -71,6 +71,41 @@ def get_wait_time(retries, args):
   return wait_time
 
 
+def update_task(task_name, new_state, retries=3):
+  """ Updates the TaskName entity for a given task.
+
+  Args:
+    task_name: A string specifying the TaskName key.
+    new_state: A string specifying the new task state.
+    retries: An integer specifying how many times to retry the delete.
+  """
+  try:
+    task_name_entity = TaskName.get_by_key_name(task_name)
+  except Exception as error:
+    retries -= 1
+    if retries >= 0:
+      logger.warning('Error fetching task name: {}. Retrying'.format(error))
+      return update_task(task_name, retries)
+
+    raise
+
+  if task_name_entity is not None:
+    task_name_entity.state = new_state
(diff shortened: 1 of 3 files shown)


AppTaskQueue/appscale/taskqueue/push_worker.py
```diff
@@ -71,6 +71,41 @@ def get_wait_time(retries, args):
   return wait_time
 
 
+def update_task(task_name, new_state, retries=3):
+  """ Updates the TaskName entity for a given task.
+
+  Args:
+    task_name: A string specifying the TaskName key.
+    new_state: A string specifying the new task state.
+    retries: An integer specifying how many times to retry the delete.
```
sjones4: probably should be "retry the update"

AppTaskQueue/appscale/taskqueue/push_worker.py
```diff
@@ -71,6 +71,41 @@ def get_wait_time(retries, args):
...
+  """ Updates the TaskName entity for a given task.
+
+  Args:
+    task_name: A string specifying the TaskName key.
+    new_state: A string specifying the new task state.
+    retries: An integer specifying how many times to retry the update.
+  """
+  try:
+    task_name_entity = TaskName.get_by_key_name(task_name)
+  except Exception as error:
```
jeanleonov: Would you mind to limit it to `except db.Error as error:`?
As far as few lines below.

cdonati:
@jeanleonov In a runtime environment, the datastore API (which the db library uses) is registered with a remote API stub. Calls are made to the API proxy, which will wrap error responses in a suitable exception.

If we were making calls through the remote API stub, catching `db.Error` would be sufficient. In this case, though, we could encounter `socket.error`, `apiproxy_errors.ApplicationError`, `datastore_errors.InternalError`, etc.

I admit I'm being a bit lazy with catching everything here. I'll track down the possible exceptions.

cdonati:
Demo
Exception raised in application runtime when the datastore is inaccessible during a taskqueue add operation

On master:
```
Traceback (most recent call last):
...
RuntimeError: ProtocolBufferReturnError(500,)
```

With branch:
```
Traceback (most recent call last):
...
InternalError
```

PR #2886, opened 2018-10-05
Install Python MySQL client during build

cdonati:
This allows users to use the Cloud SQL API without installing the library beforehand.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5524/

PR #2888, opened 2018-10-10
Move blob fetching work to DatastoreBlobReader

### LINKED ISSUE #2887: Blobstore uploads fail when using GCS keys
After the 1.8.1 upgrade, the dispatcher no longer handles uploads correctly for GCS keys.

cdonati:
This removes a level of indirection from DatastoreBlobStorage. Previously, OpenBlob would fetch data from the blobstore API via the API Server. The blobstore API would then fetch from the datastore. This was problematic because the blobstore API uses additional constructs like BlobInfo entities that the storage layer does not deal with.

Now, the datastore backend is all contained in the storage layer rather than the stub layer. OpenBlob fills the reader buffer directly from the datastore instead of from the blobstore API.

Resolves #2887.

### DIFF
--- a/AppServer/google/appengine/api/blobstore/blobstore_stub.py
+++ b/AppServer/google/appengine/api/blobstore/blobstore_stub.py
@@ -374,81 +374,10 @@ class BlobstoreServiceStub(apiproxy_stub.APIProxyStub):
       raise apiproxy_errors.ApplicationError(
           blobstore_service_pb.BlobstoreServiceError.BLOB_NOT_FOUND)
 
-    # AppScale: Access the blob data from BlobChunk entities.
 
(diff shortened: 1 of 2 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5472/
https://ocd.appscale.com:8080/job/Tools-test/3692/

cdonati:
Demo
master with https://github.com/AppScale/hawkeye/pull/146: https://ocd.appscale.com:8080/job/Hawkeye/15909/
this branch with https://github.com/AppScale/hawkeye/pull/146: https://ocd.appscale.com:8080/job/Hawkeye/15908/

PR #2890, opened 2018-10-10
Use a patched version of Cassandra

### LINKED ISSUE #2889: Cassandra's "nodetool status" and "nodetool ring" sometimes fail
The version of Cassandra that AppScale uses is sometimes unable to handle `nodetool status` and `nodetool ring`. This appears to be a ByteOrderedPartitioner-specific problem. There are more details about the issue here: https://issues.apache.org/jira/browse/CASSANDRA-12942

This is problematic for AppScale deployments because:

1. The controller that starts non-seed nodes waits for `nodetool status` on the seed node before starting Cassandra.
2. The rebalance script uses both `nodetool status` and `nodetool ring` to determine how much data to move.
3. Our recently-introduced Cassandra monitoring uses `nodetool ring` to check the cluster status.

cdonati:
This uses a version of Cassandra that was built from source. The patch from https://issues.apache.org/jira/browse/CASSANDRA-12942 is the only modification.

This fixes an issue with `nodetool status` and `nodetool ring` occasionally not working. The rebalance script, Cassandra monitoring, and service start all rely on those tools.

Resolves #2889.

### DIFF
--- a/debian/appscale_install_functions.sh
+++ b/debian/appscale_install_functions.sh
@@ -346,8 +346,10 @@ installcassandra()
 {
     CASSANDRA_VER=3.11.2
 
-    CASSANDRA_PACKAGE="apache-cassandra-${CASSANDRA_VER}-bin.tar.gz"
-    CASSANDRA_PACKAGE_MD5="1c1bc0b216f308500e219968acbd625e"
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5474/

Issue #2896, opened 2018-10-18
The endpoints framework example doesn't work


cdonati:
The "echo" demo, available at https://github.com/GoogleCloudPlatform/python-docs-samples/tree/18494f77400406dce4779e1d1b14f840355405e1/appengine/standard/endpoints-frameworks-v2/echo, does not work properly on AppScale.

PR #2897, opened 2018-10-18
Defer to vendored framework for endpoints 2

cdonati:
This change removes the custom endpoints handler for newer versions of the framework. Previously, it would use the outdated builtin endpoints library to try to server requests.

This change was introduced in the 1.9.49 SDK.

Makes progress towards #2896.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5537/

PR #2900, opened 2018-10-18
Use a fallback mechanism on connection failures

### LINKED ISSUE #2899: Load-balanced API calls are not fault-tolerant
Every API server uses the same machine to load balance requests to the datastore and taskqueue services. If that machine fails, the API servers do not switch to another load balancer.

cdonati:
When an API Server fails to establish a connection to the primary load balancer, this allows it to try a different load balancer before giving up.

This is a temporary solution until we implement leader election for load balancers.

Resolves #2899.

### DIFF
--- a/AppServer/google/appengine/api/taskqueue/taskqueue_distributed.py
+++ b/AppServer/google/appengine/api/taskqueue/taskqueue_distributed.py
@@ -21,9 +21,11 @@ from __future__ import with_statement
 __all__ = []
 
 import datetime
+import errno
 import logging
 import os.path
 import random
+import socket
 import string
 
 import taskqueue_service_pb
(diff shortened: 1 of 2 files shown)


AppServer/google/appengine/api/taskqueue/taskqueue_distributed.py
```diff
@@ -446,22 +448,27 @@ def _RemoteSend(self, request, response, method, request_id=None,
       api_request.set_request_id(request_id)
 
     tq_locations = self._GetTQLocations()
+    api_response = remote_api_pb.Response()
     for index, tq_location in enumerate(tq_locations):
```
obino: I think index is not used any longer

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5536/

PR #2901, opened 2018-10-23
Improve restore reliability and performance

cdonati:
The restore script now:
 - Uses smaller batch sizes to reduce the likelihood of excessively large batches causing timeouts
 - Uses retries to ensure that all data is inserted
 - Uses a process pool to decrease the overall restore time


cdonati:
Given our discussion about working towards a different approach for backup and restore operations, I don't think it makes sense to increase the complexity of this tool.

PR #2902, opened 2018-10-24
Add support for Stretch

cdonati:
With the latest InstanceManager changes, a lot of commits from the previous version of this pull request were no longer needed.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5541/

AppTaskQueue/appscale/taskqueue/brokers/rabbitmq.py
```diff
@@ -1,23 +1,12 @@
...
   Returns:
     A string representing the location of RabbitMQ.
   """
-  raw_ips = file_io.read(RABBITMQ_LOCATION_FILE)
-  ips = raw_ips.split('\n')
-  rabbitmq_ip = ips[0]
-
-  return 'amqp://guest:guest@' + rabbitmq_ip + ':' + \
-         str(RABBITMQ_PORT) + '//'
+  return 'amqp://guest:guest@localhost:{}//'.format(RABBITMQ_PORT)
```
obino: Is this needed for Stretch?

cdonati:
@obino By default, the newer rabbitmq-server package on stretch only allows guest connections from localhost.
@sjones4 Yes, I'm open to changing that instead. Even better, it would make sense for us to use a proper user and password if we connect to a non-local rabbitmq node.

In master, we always use the first IP address in the taskqueue_nodes file, which is a randomized list of taskqueue machines. In my opinion, that approach has pretty much the same consequences as connecting to localhost.

sjones4:
@cdonati sounds ok to me. We could track setting up rabbitmq with an appscale specific user as a separate item.

PR #2903, opened 2018-10-25
Implement apps.services.list

cdonati:
This returns a list of services for a given project.

This also removes support for project deletion since we don't have an existing process for removing a particular project's data.

Goes with https://github.com/AppScale/appscale-tools/pull/728.

Makes progress towards https://github.com/AppScale/appscale-tools/issues/727.


AdminServer/appscale/admin/__init__.py
```diff
@@ -429,22 +381,35 @@ def delete(self, project_id):
...
-        ports_to_close.append(port)
+    try:
+      service_ids = self._zk_client.get_children(services_node)
+    except NoNodeError:
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
+                            message='Services node not found for project')
 
-    IOLoop.current().spawn_callback(self.wait_for_delete,
-                                    ports_to_close, project_id)
+    prefix = '/'.join(['apps', project_id, 'services'])
```
obino: Probably not the best PR to comment about this, but should we be under /appscale or something like that instead of /apps?

cdonati: This is the response body of the Admin API call. The service representation is defined here: https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services#Service

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5632/
https://ocd.appscale.com:8080/job/Migration/

PR #2905, opened 2018-10-26
Logging update to use per-module loggers

sjones4:
Logging is updated to use a per-module logger with corresponding unit test changes.

Fixes appscale#2893

### DIFF
--- a/APIServer/appscale/api_server/server.py
+++ b/APIServer/appscale/api_server/server.py
@@ -73,7 +73,7 @@ def main():
     with open(pidfile_location, 'w') as pidfile:
         pidfile.write(str(os.getpid()))
 
-    logger.setLevel(logging.INFO)
+    logging.getLogger('appscale').setLevel(logging.INFO)
 
     zk_client = KazooClient(hosts=','.join(args.zookeeper_locations),
                             connection_retry=ZK_PERSISTENT_RECONNECTS)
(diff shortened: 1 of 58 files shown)


APIServer/appscale/api_server/server.py
```diff
@@ -17,7 +17,7 @@
 from appscale.common.constants import VAR_DIR
 from appscale.common.constants import ZK_PERSISTENT_RECONNECTS
 
-logger = logging.getLogger('appscale-api-server')
+logger = logging.getLogger(__name__)
```
cdonati: I think the `main` function in this module should set the level on the `appscale.api_server` logger instead of the logger defined here so that all of the child loggers are changed.

AppDB/appscale/datastore/utils.py
```diff
@@ -203,7 +203,7 @@ def fetch_and_delete_entities(database, table, schema, first_key,
 
   last_key = first_key + '\0' + TERMINATING_STRING
 
-  logging.debug("Deleting application data in the range: {0} - {1}".
+  logger.debug("Deleting application data in the range: {0} - {1}".
```
cdonati: I think the logger defined in this module is imported by `appscale.datastore.scripts.datastore`, and the level is set there. For consistency, it might make sense to create the logger from the module name here and set the level for the `appscale.datastore` logger in the `appscale.datastore.scripts.datastore.main`.

PR #2910, opened 2018-11-09
Bound update threads

obino:
Ensure we don't have multiple threads querying Hermes on the various systems.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5609/

AppController/djinn.rb
```diff
@@ -1137,38 +1137,44 @@ def get_cluster_stats_json(secret)
...
   end
 
   # Updates our locally cached information about the CPU, memory, and disk
   # usage of each machine in this AppScale deployment.
   def update_node_info_cache
+    threads = []
     new_stats = []
-
-    Thread.new {
+    @state_change_lock.synchronize {
```
cdonati: Why do these fetches need to be under the state change lock?

obino: we are looping through @nodes, which is  a shared variable and could change (say add_nodes etc...)

obino: Probably: since we do this just about anywhere else in the djinn code, I was thinking to leave that to a different PR that works around all the Mutexes not just this one. I'm happy to get it started here though it you think it makes sense.

cdonati: The reason I bring it up for this PR is that you wrapped several other places with the state change lock. For example, `appscale status` now has the potential to hang for 10 more seconds than it did before.

I tried looking into whether `JSON.dump` was atomic (locking the hash while iterating it, for example), but I couldn't find any readily available answers.

obino: Got it. Let me add something now.

obino: @cdonati let me know if something like this makes sense to you or you prefer something simpler. I ended up spawning a thread to update nodes, to ensure we don't lock the duty cycle no matter what (say badly behavior network) and limit the holding of the lock within the update method.

obino:
To show that multiple threads can happen upon slow (or faulty nodes) I added a start/end log message and a sleep of 30 seconds. This what happens on master upon start

W, [2019-01-27T22:18:59.347812 #15130]  WARN -- : +++++ Starting thread to read nodes stats.
W, [2019-01-27T22:19:02.236950 #15130]  WARN -- : +++++ Starting thread to read nodes stats.
W, [2019-01-27T22:19:11.666789 #15130]  WARN -- : +++++ Starting thread to read nodes stats.
W, [2019-01-27T22:19:22.118085 #15130]  WARN -- : +++++ Starting thread to read nodes stats.
W, [2019-01-27T22:19:30.531481 #15130]  WARN -- : +++++ Ending thread to read nodes stats.

With this PR we have only one thread at a time updating values

W, [2019-01-27T22:43:47.579390 #3270]  WARN -- : +++++ Starting thread to read nodes stats.
W, [2019-01-27T22:44:18.322293 #3270]  WARN -- : +++++ Ending thread to read nodes stats.
W, [2019-01-27T22:44:19.312901 #3270]  WARN -- : +++++ Starting thread to read nodes stats.

PR #2911, opened 2018-11-10
Master reload state

obino:
Master now reload the state as any other nodes. This allows the master to pick up changes done by hand to the zookeeper state, as well as the change of roles.


AppController/djinn.rb
```diff
@@ -1834,27 +1837,32 @@ def job_start(secret)
...
+      # of the deployment that each node needs to reload. For example it
+      # contains the current listing of AppServers and Nodes.
+      unless restore_appcontroller_state
+        # Master is responsible to populate the state for the other nodes
+        # consumption, and since we know we could talk to zookeeper,
+        # master needs to go ahead and write the state.
+        unless my_node.is_shadow?
+          @state = "Couldn't reach the deployment state: now in isolated mode"
+          Djinn.log_warn("Cannot talk to zookeeper: in isolated mode.")
+          next
```
cdonati: Should there be a sleep here? Otherwise, this could be a busy loop for a disconnected AC.

AppController/djinn.rb
```diff
@@ -1834,27 +1837,32 @@ def job_start(secret)
...
+      end
 
-      # Every other node syncs its state with the login node state. The
-      # load_balancers need to check the applications that got loaded
-      # this time, to setup the routing.
-      my_versions_loaded = @versions_loaded if my_node.is_load_balancer?
+      # We act here if options or roles for this node changed.
+      check_role_change(old_options, old_jobs)
+
+      # The master node has few more work to do.
```
cdonati: ~few~

AppController/djinn.rb
obino: The busy loop should actually be in pick_zookeeper: I seem to remember that it was intentional, since an isolated node shouldn't quite be functional, so it is sitting waiting to reconnect to zookeeper. We could add a sleep in pick_zookeeper?

cdonati: Okay, I see now that the pick_zookeeper at least sleeps until the zk port is available, so I think this is fine.

obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5665/

PR #2915, opened 2018-11-16
Rewrite package-list.awk to a bash/sed script

scragraham:
- 50% reduction in code
 - code is a bit clearer
 - also rewrite of remove-list.awk to remove-list.sh


scragraham:
eucalyptus image build: https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/6802/console

package comparison to a build from master shows no differences, as expected.

PR #2916, opened 2018-11-20
IaaS restructure

whoarethebritons:
This PR converts the Infrastructure Manager Service to a Tornado webserver. Apologies for the unorganized commits, I worked towards utilizing zookeeper to store operations but then realized that it wasn't really helpful to do so. I think that might be the main confusion for going commit by commit.

I moved the utils file from the IaaS to the backup for AppDB because I removed the dependency on that file from the IaaS.

The system manager's changes are mostly removing the secret key as a parameter for the methods since calls to the IaaS will already be authenticated and changing the name of the class. In the future, I'd like to remove the system manager so the djinn would use hermes instead, but I held off on that.

I deleted the agent unit tests because, as discussed, they belong with the agents.


whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/5638/
https://ocd.appscale.com:8080/job/Datastore/312/

InfrastructureManager/appscale/infrastructure/__init__.py
```diff
@@ -0,0 +1,522 @@
...
+  """
+
+  # States a particular VM deployment could be in
+  STATE_PENDING = 'pending'
+  STATE_SUCCESS = 'success'
+  STATE_FAILED  = 'failed'
+
+  def initialize(self, agent_factory):
+    """ Defines required resources to handle requests.
+4
```
sjones4: Typo

InfrastructureManager/appscale/infrastructure/__init__.py
```diff
@@ -0,0 +1,522 @@
...
+    if num_vms <= 0:
+      logger.warn('Invalid VM count: {0}'.format(num_vms))
+      raise CustomHTTPError(HTTPCodes.BAD_REQUEST,
+                            message='Invalid VM count: {0}'.format(num_vms))
+
+    try:
+      agent, run_params = get_agent(self.agent_factory,
+                                    BaseAgent.OPERATION_RUN, args)
+    except AgentConfigurationException as exception:
+      logger.exception("Error creating agent!")
```
sjones4: Could log some details from the error

InfrastructureManager/appscale/infrastructure/__init__.py
```diff
@@ -0,0 +1,522 @@
...
+      raise CustomHTTPError(HTTPCodes.INTERNAL_ERROR,
+        message='Invalid agent configuration: {0}'.format(exception))
+
+    operation_id = str(uuid.uuid4())
```
sjones4: It could be nice to generate the operation id early and then include it in all logging related to that operation.

InfrastructureManager/appscale/infrastructure/__init__.py
whoarethebritons: I think that logger.exception logs the full exception, is that what you mean?

InfrastructureManager/appscale/infrastructure/__init__.py
whoarethebritons: I looked at this a bit more and realized that everything up until the line where the operation_id is currently assigned is configuration checking/error handling. It's extremely likely that if one request fails due to any of the above config checks then all will, so I'm not certain if an operation id would be very helpful. Additionally we have no need to save information about an operation id if it fails above since it does not actually make a call that we need to check on.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/5827/
https://ocd.appscale.com:8080/job/Taskqueue-e2e-Test/247/

working on fixing Hawkeye 6 node

PR #2918, opened 2018-11-21
Run datastore servers in a custom slice

### LINKED ISSUE #2917: The ServiceManager can lose track of datastore servers
It's possible for datastore servers to be re-assigned to a different memory cgroup than they were originally started in. When this happens, the ServiceManager is not able to locate the running processes.

The easiest way to reproduce this on Xenial is to run `systemctl daemon-reload && systemctl restart monit`.

cdonati:
The memory cgroup is not as stable as the namespace cgroup that the process is running in. It's possible for systemd to reset the process's memory cgroup to the parent slice's memory cgroup.

This change keeps track of datastore processes in an "appscale-datastore" slice to ensure that systemd won't move them.

Resolves #2917.

### DIFF
--- a/AdminServer/appscale/admin/service_manager.py
+++ b/AdminServer/appscale/admin/service_manager.py
@@ -21,9 +21,6 @@ from appscale.common.async_retrying import retry_data_watch_coroutine
 from appscale.common.constants import (ASSIGNMENTS_PATH, CGROUP_DIR, HTTPCodes,
                                        LOG_DIR, VAR_DIR)
 
-# The cgroup used to start datastore server processes.
-DATASTORE_CGROUP = ['memory', 'appscale-datastore']
-
 # The characters allowed in a service identifier (eg. datastore)
 SERVICE_ID_CHARS = '[a-z_]'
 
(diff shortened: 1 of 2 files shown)


AdminServer/appscale/admin/service_manager.py
```diff
@@ -180,9 +220,16 @@ def start(self):
       log_file = os.path.join(LOG_DIR,
                               'datastore_server-{}.log'.format(self.port))
       self._stdout = open(log_file, 'a')
-      self.process = psutil.Popen(
-        ['cgexec', '-g', ':'.join(DATASTORE_CGROUP)] + start_cmd,
-        stdout=self._stdout, stderr=subprocess.STDOUT)
+
+      # With systemd-run, it's possible to start the process within the slice.
+      # To keep things simple and maintain backwards compatibility with
+      # pre-systemd distros, move move the process after starting it.
```
jeanleonov: `move move` -> `move`

cdonati:
Yes, we don't use cgroups for limiting resource usage yet. We just use them for keeping track of datastore processes.

I like your suggestion about adding an additional documentation file. I'll give that a shot.

cdonati:
@sjones4 That's true. Fortunately, it is present on trusty, which is the only distro without systemd that we support right now.

PR #2931, opened 2018-12-05
Rework search query parsing and building

jeanleonov:
Changes:
 - query_parser module was renamed to query_converter.
 - query convertor stoped doing url-encoding of query,
   now it just prepares params dictionary to be sent to solr,
   encoding is done in solr_interface which sends http request.
 - escaping of all solr special signs was replaced with escaping
   of only quote sign. It should be enough as all values are
   wrapped with quotes. Though there're still some escaping issues
   related to incompatibility between GAE and Solr query syntax.
 - protobuffer objects were removed from query_converter.
 - usage of protobuffer objects in solr_interface was a bit reduced.

Outcomes:
 - Date filtering become available in some way. It doesn't match GAE
   API - you have to specify date in following form in query:
   `index.search('date_field:"2018-12-05T00\:00\:00Z")`.
 - AND/OR statements were fixed. Previously built query was invalid.


SearchService/solr_interface.py
```diff
@@ -1,48 +1,47 @@
...
-from datetime import datetime
-
-from query_parser import Document
-
 from appscale.common import appscale_info
 
+import query_converter
+import search_exceptions
+from constants import (
+  HTTP_OK, INDEX_NAME_FIELD, INDEX_LOCALE_FIELD, SOLR_SERVER_PORT
```
scragraham: Not a deal breaker but we should probably replace HTTP_OK with:
httplib.OK (python 2.x)
http.HTTPStatus.OK (python 3.x)

PR #2932, opened 2018-12-06
Check for valid local source archive

cdonati:
In addition to checking if the current machine is listed as a hoster, this checks if the local filesystem contains a valid source archive. This prevents a situation where the fetch is skipped because a previously running machine with the same IP was once a hoster for the archive.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5714/

scragraham:
demo looks good.

PR #2935, opened 2018-12-09
Remove index management from datastore clients

cdonati:
This removes the automatic initialization step where the API Server created and removed indexes according to the local configuration file in the revision's source.

It also removes unnecessary work that was done to find indexes that were relevant to queries and mutations.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5731/

PR #2944, opened 2018-12-21
Limit the number of entities fetched per statement

### LINKED ISSUE #2943: Fetching many entities at a time can time out
Since batch get operations request all entities within a single operation, the coordinator can time out before satisfying the request.

cdonati:
Without this restriction, Cassandra can easily time out while trying to fetch many rows from the entities table.

This particular fix has the potential to increase the latency a bit in some cases, but some increased latency is better than a timeout.

Resolves #2943.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5766/
https://ocd.appscale.com:8080/job/Taskqueue-e2e-Test/190/

PR #2947, opened 2018-12-27
Store hints within the persistent state directory

cdonati:
If the cluster is heavily loaded, the size of Cassandra hint data could be problematic for the OS disk.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5789/

PR #2949, opened 2018-12-28
Use gossip to find live Cassandra nodes

### LINKED ISSUE #2948: Cassandra can take an unreasonable amount of time to start
When non-seed database nodes check if the seed node is ready, the seed node uses `nodetool status` to determine if Cassandra is ready. Since `nodetool status` gets progressively slower when data is added to BOP clusters, this process can take an unreasonable amount of time for each database machine.

Since the ownership calculation is what takes the most amount of time, the check can return faster by using gossip to determine live nodes.

cdonati:
This is more efficient than using `nodetool status` because it is able to skip the ownership calculation, which takes an unreasonable amount of time on a BOP cluster that contains a lot of data.

This also fixes a ThreadError that can occur if the lock was not obtained.

Resolves #2948.

### DIFF
--- a/AppController/lib/cassandra_helper.rb
+++ b/AppController/lib/cassandra_helper.rb
@@ -3,6 +3,7 @@ require 'djinn'
 require 'djinn_job_data'
 require 'helperfunctions'
 require 'monit_interface'
+require 'set'
 
 # A String that indicates where we write the process ID that Cassandra runs
 # on at this machine.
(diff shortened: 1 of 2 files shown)


AppController/lib/cassandra_helper.rb
```diff
@@ -174,12 +175,19 @@ def needed_for_quorum(total_nodes, replication)
...
+# Returns an array of nodes in 'Up Normal' state.
 def nodes_ready
-  output = `"#{NODETOOL}" status`
-  nodes_ready = 0
+  output = `"#{NODETOOL}" gossipinfo`
+  return [] unless $?.exitstatus == 0
+
+  live_nodes = Set[]
+  current_node = nil
   output.split("\n").each { |line|
```
whoarethebritons: would you mind putting a comment about the expected output of this command? in case something breaks in the future it'd be easier to find the problem.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5777/

cdonati:
Demo
On a decently loaded cluster, `nodetool status` takes 3.5 min. `nodetool gossipinfo` takes 1 sec.

PR #2951, opened 2018-12-30
Don't return when we need to run ensure

obino:
The return statement takes precedence over the ensure branch. This
commit removes the return within the try_lock block, ensure we unlock
before returning.


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5781/

cdonati:
It seems to me the ensure is executed despite the return:
```
irb(main):001:0> def test_ensure
irb(main):002:1>   begin
irb(main):003:2*     return
irb(main):004:2>   ensure
irb(main):005:2*     puts 'ensured'
irb(main):006:2>   end
irb(main):007:1> end
=> :test_ensure
irb(main):008:0> test_ensure
ensured
=> nil
```
Am I misunderstanding the problem you are addressing?

obino:
@cdonati I was generically try to address the threadError we saw that seemed to be associated with that code. But I think you are right: the code as we have it, it is indeed correct so I will close this PR.

obino:
Got it, yep makes sense.


-----Original Message-----

PR #2953, opened 2018-12-31
Search data models

Honcharov12:
Creating unified data models for Search API for future improvement of Search Service design. The final idea is for decouple Search Service and SORL interface because currently any changes in interaction with SDK can cause some changes in SOLR API. 
Presently implemented interaction with Search API because interaction with SOLR will improve in more form for further development.
Based on #2931 which need to be merged first.


Honcharov12:
Currently these changes do not applied in Search Service, but they can cause Search Service's code readability after applying it.

PR #2956, opened 2019-01-04
Improve speed of rebalance script

### LINKED ISSUE #2955: The rebalance script is not always fast enough
The speed of the rebalance script decreases as the amount of data in the Cassandra cluster increases. This is due to the `nodetool status` and `nodetool ring` operations. They have the potential to become unreasonably slow.

cdonati:
This removes the call to `nodetool ring`, which can get unreasonably slow as the amount of data in a BOP cluster increases.

It also adds a couple flags that allow the `nodetool status` call to be skipped if the user is already sure the sanity checks will pass.

Resolves #2955.

### DIFF
--- a/AppDB/appscale/datastore/cassandra_env/rebalance.py
+++ b/AppDB/appscale/datastore/cassandra_env/rebalance.py
@@ -1,4 +1,5 @@
 from __future__ import division
+import argparse
 import logging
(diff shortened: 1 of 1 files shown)


scragraham:
Demo looks good, approved for merge.

PR #2960, opened 2019-01-08
Drop requests that don't match the login property

cdonati:
If the user has specified a DNS name for a login property, this change ensures that the host header matches before passing the request on to the application.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5796/

cdonati:
https://github.com/AppScale/appscale/pull/2961 should go in before this one.

PR #2961, opened 2019-01-08
Use login property when setting cron locations

cdonati:
This ensures that the cron client sets the right host header and uses the right location when making requests.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5798/

PR #2966, opened 2019-01-11
Replacing zookeeper_locations.json with zookeeper_locations

jeanleonov:
It's much easier to work with a simple text file containing
zookeeper locations written as lines instead of json formatted file.


jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/5809/

PR #2969, opened 2019-01-19
Only list unique items in project list

cdonati:
This prevents the dashboard from showing duplicate project entries when multiple services are deployed.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6115/

PR #2979, opened 2019-01-26
Wait for result before setting chosen task name

cdonati:
When the task enqueue request fails, it doesn't make sense to include a chosen task name in the error response.

This likely has no effect beyond the cleanliness of the API proxy response.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6156/

PR #2985, opened 2019-02-05
Prevent SDK from starting more than one instance

cdonati:
This makes sure that a single AppScale instance does not map to multiple runtime instance processes.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6116/
http://ocd.appscale.net:8080/job/Hawkeye/17918/

PR #2986, opened 2019-02-05
Continue batch query if limit hasn't been reached

cdonati:
Since the count field is used to specify the batch size, it is not suitable for determining if the query has been satisfied. The original query's limit field makes more sense.

### DIFF
--- a/AppServer/google/appengine/api/datastore_distributed.py
+++ b/AppServer/google/appengine/api/datastore_distributed.py
@@ -486,7 +486,7 @@ class DatastoreDistributed(apiproxy_stub.APIProxyStub):
       del self.__queries[cursor_handle]
       return
 
-    if internal_cursor.get_offset() >= internal_cursor.get_count():
+    if query.limit() and internal_cursor.get_offset() >= query.limit():
       query_result.set_more_results(False)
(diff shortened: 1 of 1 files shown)


AppServer/google/appengine/api/datastore_distributed.py
```diff
@@ -486,7 +486,7 @@ def _Dynamic_Next(self, next_request, query_result, request_id=None):
       del self.__queries[cursor_handle]
       return
 
-    if internal_cursor.get_offset() >= internal_cursor.get_count():
+    if query.limit() and internal_cursor.get_offset() >= query.limit():
```
sjones4: should we check `query.has_limit()` here? (and below)

cdonati: Yes. That's clearer. I wasn't considering the case when the client sets the limit to 0 (in which case, they should get 0 results instead of all of them).

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6158/
http://ocd.appscale.net:8080/job/Autoscale-test-manual-triggers/2430/
http://ocd.appscale.net:8080/job/Blobstore/4357/
http://ocd.appscale.net:8080/job/Datastore-Benchmark/526/

PR #2988, opened 2019-02-07
Re-add editing capabilities to datastore viewer

### LINKED ISSUE #2927: The dashboard's datastore viewer is read-only
The datastore viewer should support create, update, and delete operations.

cdonati:
Now that the datastore service is responsible for keeping track of composite index definitions, the dashboard can safely mutate entities for other applications.

Resolves #2927.

This also makes a few improvements to the editor interface.

When the user enters an empty string as a value, it stores the empty string instead of removing the property. This does mean that it's not possible to remove a property with the editor, but I think this behavior is less surprising.

When displaying a falsy value like `0`, the editor form will now populate the input with the value instead of an empty string.

Most embedded entities will now display as JSON dictionaries unless they contain unserializable values.

The "Create New Entity" button now works in modern browsers.

### DIFF
--- a/AppDashboard/lib/datastore_viewer.py
+++ b/AppDashboard/lib/datastore_viewer.py
@@ -647,13 +647,8 @@ class DatastoreEditRequestHandler(DatastoreViewerPage):
         # property was already empty.
         continue
 
-      if form_value:
-        # TODO: Handle parse exceptions.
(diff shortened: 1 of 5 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5944/

PR #2989, opened 2019-02-08
Pass app identity calls to the external API server

cdonati:
This simplifies runtime instance configuration by routing all calls through one API server. This will make it easier for the java runtime to use the Python SDK's API server.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6202/

PR #2990, opened 2019-02-08
Don't wait for autoscaled nodes

obino:
If the controller get restarted, as part of the initialization, will wait for all nodes to be up and running. This means that if an autoscaled node was terminated or in the process of terminating, it could wait forever for the node to come back. With this PR the AC will try to initialize all nodes, but wait only for the nodes it must wait for (that is all but the autoscaled one).


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/6010/

PR #2992, opened 2019-02-09
Fix bionic php dependencies

cdonati:
The php7.2 metapackage contains a bunch of stuff we don't need, and our dependency build does not allow installing it at the same time as php7.2-bcmath for some reason.


cdonati:
Demo
Dependency build with master: https://ocd.appscale.com:8080/job/Check-Cloud-Ubuntu-Img/6416/
Dependency build with branch: https://ocd.appscale.com:8080/job/Check-Cloud-Ubuntu-Img/6419/ (got past dependency conflict)

PR #2997, opened 2019-02-12
Remove monit maintenance port from the firewall.

obino:


firewall.conf
```diff
@@ -13,28 +13,29 @@ if [ $? -ne 0 ]; then
...
 
-# Allow all SSH traffic 
+# Allow all SSH traffic.
 iptables -A INPUT -p tcp --dport ssh -j ACCEPT
 
-# Allow for all traffic from anywhere for the following services
+# Allow for all traffic from anywhere for the following services.
 iptables -A INPUT -p tcp --dport 80 -j ACCEPT          # nginx
 iptables -A INPUT -p tcp --dport 443 -j ACCEPT         # nginx (SSL)
-iptables -A INPUT -p tcp --dport 2812 -j ACCEPT        # monit dashboard
```
scragraham: Do we want to add this down below in the debugging section?

obino: I wouldn't, but I can be convinced. I doubt that anyone would (and could) interact successfully with monit via that interface.

cdonati: I would vote for removing it. It wouldn't really make sense outside of a single-node deployment, and we don't expose monit's http interface outside of localhost right now.

scragraham: I'm fine with removing it, just saw movement of other ports to the debugging section.

PR #3002, opened 2019-02-26
Access AC and AdminServer with private IPs

cdonati:
When the dashboard is acting as a client, it shouldn't assume that the controller and AdminServer are accessible via the public internet.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6121/

PR #3010, opened 2019-03-03
Enforce min machines

obino:
min_machines variables defines the minimum number of nodes in a deployment: this PR ensures that when changed at run-time machines are added right away (provided we are in a cloud environment).


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5984/

obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/5996/

AppController/djinn.rb
```diff
@@ -1849,33 +1849,37 @@ def job_start(secret)
         APPS_LOCK.synchronize { versions_loaded_now = @versions_loaded.clone }
 
         # Starts apps that are not running yet but they should.
+        versions_to_load = versions_loaded_now - my_versions_loaded
```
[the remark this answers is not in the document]
obino: It's an old habit from when I had to declare all variables before using them, not real purpose a part from that. If it's too disturbing I can revert it.

AppController/djinn.rb
```diff
@@ -1849,33 +1849,37 @@ def job_start(secret)
...
+          }
+          APPS_LOCK.synchronize {
+            scale_deployment
+            scale_up_instances(missing_nodes) if missing_nodes > 0
+          }
+        end
+
+        # Load balancers need to check/update applications that have been
+        # undeployed and nginx/haproxy.
+        APPS_LOCK.synchronize { check_stopped_apps }
```
cdonati: Doesn't the shadow need to run `check_stopped_apps`?

AppController/djinn.rb
```diff
@@ -5298,15 +5297,31 @@ def get_hosts_for_version(version_key)
...
+  # Try to add an AppServer for the specified version, ensuring that a
+  # minimum number of AppServers is always kept.
   #
   # Args:
   #   version_key: A String containing the version key.
   #   delta_appservers: The desired number of new AppServers.
   # Returns:
-  #   An Integer indicating the number of AppServers we didn't start (0
-  #     if we started all).
+  #   An Interger indicating the number of nodes we need to scale to
```
cdonati: "Interger"

AppController/djinn.rb
obino: Good question: on the quick answer, shadow is a load_balancer and in the the short term we cannot put it on a non-load-balancer machine. Long term  I would like to have shadow to potentially be on all roles, but it will take a bit work to get there.

cdonati: I think it's important to maintain the mapping we had from role to function. This improves readability and makes future changes much easier. If I were to move all of the functions that the shadow performs to a separate process, I might miss this bit since there's no reference to shadow here.

PR #3024, opened 2019-03-16
Remove HAProxy stats URL

cdonati:
The stats are already accessible via the unix socket, and we don't need a public interface for them.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6163/
http://ocd.appscale.net:8080/job/Datastore/726/
http://ocd.appscale.net:8080/job/Migration/3713/

PR #3025, opened 2019-03-17
Add initial implementation of datastore-backed sessions

### LINKED ISSUE #3014: persistent sessions do not work as expected
Appscale 3.6, App is in Java.
When we have 2 instances for the appserver, we are observing that our application users are getting logged out  - i.e session is invalidated. We think that this happens when the requests are served by the instance which is not the same as the one which was used to login. 

Including @cdonati's assessment:

I don't think we have any automated tests for the Java runtime's sessions implementation. My guess is that the SDK is storing session data in memory rather than using memcache and the datastore.

cdonati:
Though a much more robust implementation is available in the 1.9.66 SDK, it would require too much effort to backport it to the current SDK right now. This particular implementation is inefficient,
but I think anything is an improvement over the in-memory implementation that it replaces.

Resolves #3014.

### DIFF
--- a/AppServer_Java/src/org/mortbay/jetty/servlet/DatastoreSessionManager.java
+++ b/AppServer_Java/src/org/mortbay/jetty/servlet/DatastoreSessionManager.java
@@ -74,14 +74,16 @@ public class DatastoreSessionManager extends AbstractSessionManager {
         session._invalid = (boolean)entity.getProperty("_invalid");
 
         Blob blob = (Blob)entity.getProperty("_values");
-        try {
-            ByteArrayInputStream bais = new ByteArrayInputStream(blob.getBytes());
-            ObjectInputStream ois = new ObjectInputStream(bais);
-            session._values = (HashMap<String, Object>)ois.readObject();
-            ois.close();
-            bais.close();
-        } catch (ClassNotFoundException | IOException error) {
-            throw new RuntimeException(error);
+        if (blob != null) {
+            try {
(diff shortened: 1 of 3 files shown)


AppServer_Java/src/org/mortbay/jetty/servlet/DatastoreSessionManager.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private AbstractSessionManager.Session decodeEntityAsSession(Entity entity) {
+        String clusterId = (String)entity.getProperty("_clusterId");
+        long created = (long)entity.getProperty("_created");
+        DatastoreSessionManager.Session session = new DatastoreSessionManager.Session(created, clusterId);
+        session._invalid = (boolean)entity.getProperty("_invalid");
+
+        Blob blob = (Blob)entity.getProperty("_values");
+        try {
```
sjones4:
```suggestion
        if (blob != null) try {
```
`blob` will be null if there are no values in the session.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6326/

cdonati:
Demo
https://github.com/AppScale/hawkeye/pull/162 with master: http://ocd.appscale.net:8080/job/Hawkeye/18594/
https://github.com/AppScale/hawkeye/pull/162 with this branch: http://ocd.appscale.net:8080/job/Hawkeye/18593/

PR #3027, opened 2019-03-19
Adds check for system stats

obino:
The call to the controlled statistics get_stats_node_json wasn't
considering the fact that a BAD_SECRET_MSG could have been returned
instead of a JSON string. Also in a periodic call to print stats, no
exception were check, which could have caused the following error:

/root/appscale/AppController/lib/infrastructure_manager_client.rb:237:in `block in get_system_stats': Could not get system statistics (FailedNodeException)
        from /root/appscale/AppController/lib/infrastructure_manager_client.rb:227:in `loop'
        from /root/appscale/AppController/lib/infrastructure_manager_client.rb:227:in `get_system_stats'
        from /root/appscale/AppController/djinn.rb:5757:in `get_node_stats_json'
        from /root/appscale/AppController/djinn.rb:1891:in `job_start'
        from /root/appscale/AppController/djinnServer.rb:157:in `<main>'


obino:
OCD build https://ocd.appscale.com:8080/job/Daily%20Build/6027/

obino:
Demo: set up a deployment (I used a 3 nodes deployment) then stop the iaas_manager on the shadow (for more spectacular failures, and wait few minutes (1/2 hour for safety) and check the logs. Master will hve signs that the controller crashed and restarted (with the stacktrace in the commit). This PR will have the controller still running with errors about trying to talk to the IM and with missing stats (for example appscale status -v will miss one node informations).

AppController/djinn.rb
```diff
@@ -5785,7 +5797,13 @@ def get_node_stats_json(secret)
 
     # Get stats from SystemManager.
     imc = InfrastructureManagerClient.new(secret)
-    system_stats = JSON.load(imc.get_system_stats)
+    begin
+      system_stats = JSON.load(imc.get_system_stats)
+    rescue SOAP::FaultError, FailedNodeException => exception
```
jeanleonov: If we could recognize BadSecret error it would be nice to restart InfrastructureManager. Sometimes the manager is not restarted properly and continues to use old secret.

AppController/djinn.rb
```diff
@@ -1151,26 +1151,28 @@ def update_node_info_cache
...
-        threads << Thread.new {
+      threads << Thread.new {
+        Thread.current[:new_stat] = nil
+        if ip == my_node.private_ip
+          ret = get_node_stats_json(@@secret)
+        else
           acc = AppControllerClient.new(ip, @@secret)
+          ret = acc.get_node_stats_json
+        end
+        if ret == BAD_SECRET_MSG
```
cdonati: What do you think about putting this check in `AppControllerClient.make_call` and raising an exception there? That way, every place that uses the controller client library does not need to perform an additional check.

obino: good idea: let me know if this is what you had in mind

obino:
Since this is been out for a while, and I had to merge the latest master, I re-run the daily job http://ocd.appscale.net:8080/job/Daily%20Build/6211

PR #3028, opened 2019-03-20
Update default config for java runtime

sjones4:
Update `webdefault.xml` to match the currently shipped version but with development functionality removed. Previously the build was not replacing this configuration when rebuilding the JARs.


sjones4:
Daily build no. 6178

PR #3029, opened 2019-03-20
Add B* instance classes

sjones4:
This pull request adds memory sizes for the B* instance classes. This memory size metadata is not typically used as tools do not currently propagate the instance class.


sjones4:
Daily build no. 6154

PR #3030, opened 2019-03-21
Configure open file limit for monit

### LINKED ISSUE #3023: Services have a low open file limit
Some services can run out of available sockets. One possible fix would be to increase the limit for the monit process so that all the services it starts inherit a higher limit.

sjones4:
Configure the open file limit for monit using the same approach we use for nginx.

Fixes #3023


sjones4:
Daily build no. 6149

PR #3031, opened 2019-03-22
Datastore statistics timestamp second resolution

sjones4:
Clear the sub-second components of the timestamp for datastore statistics.

This is as-per GAE and expected by some clients, e.g.:

https://github.com/apache/beam/blob/247e93ef7fbe87dbb1cd82a0ce6b2e6aa5254f7c/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/datastore/DatastoreV1.java#L368


sjones4:
Daily build no. 6138

PR #3034, opened 2019-03-26
Cron schedule list parsing and sub-daily interval crontab generation

sjones4:
Cron schedule lists have whitespace removed so they can be handled by the parser.

Schedules such as `every 45 minutes` now repeat every 45 minutes during a day, rather than running once on the hour and once 45 minutes past the hour. Longer schedules such as `every 90 minutes` now run every 90 minutes rather than every hour. Intervals in this form that are longer than 1 day still incorrectly run once per day. For intervals that do not repeat every 24 hours the last run of the day is dropped so that there is not a short interval.

### DIFF
--- a/AppController/lib/cron_helper.rb
+++ b/AppController/lib/cron_helper.rb
@@ -262,7 +262,7 @@ CRON
   #   standard cron format. It cannot be applied to a crontab because it only
   #   specifies the frequency of the job, and not the action to perform.
   def self.convert_messy_format(schedule)
-    splitted = schedule.split
+    splitted = schedule.gsub(/[[:space:]]*,[[:space:]]*/, ',').split
 
     # Only 3, 4, 5 or 7-token schedules are supported.
     # Examples:
--- a/AppController/lib/cron_helper.rb
+++ b/AppController/lib/cron_helper.rb
@@ -426,6 +426,47 @@ CRON
     cron_lines
   end
 
+  # Generates one or more cron lines in standard cron format for an interval
+  # of the given minutes or hours.
+  #
+  # Where values do not repeat evenly in a 24 hour period the last run for the
+  # day is dropped to avoid a short interval.
+  #
+  # Args:
+  #   num: The interval length.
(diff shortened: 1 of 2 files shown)


sjones4:
Daily build no. 6174

sjones4:
The demo is to deploy an application with the `cron.yaml`:

```
cron:
- description: "90 minute example"
  url: /example
  schedule: every 90 minutes
```

and verify the corresponding crontab e.g. `/etc/cron.d/appscale-guestbook`, before fix:

```
MAILTO=""
*/90 * * * * root curl -sSH "X-Appengine-Cron:true" -H "X-AppEngine-Fake-Is-Admin:XXX" -k -L "http://192.168.102.245:8083/example" 2>&1 >> /var/log/appscale/cron-guestbook.log
```

with fix in place:

```
MAILTO=""
0 0,3,6,9,12,15,18,21 * * * root curl -sSH "X-Appengine-Cron:true" -H "X-AppEngine-Fake-Is-Admin:XXX" -k -L "http://192.168.102.107:8080/example" 2>&1 >> /var/log/appscale/cron-guestbook.log
30 1,4,7,10,13,16,19,22 * * * root curl -sSH "X-Appengine-Cron:true" -H "X-AppEngine-Fake-Is-Admin:XXX" -k -L "http://192.168.102.107:8080/example" 2>&1 >> /var/log/appscale/cron-guestbook.log
```

so with the fix the timing would be 0:00, 1:30, 3:00, 4:30, etc.

Whitespace fix example `cron.yaml` is:

```
cron:
- description: "select days example"
  url: /example
  schedule: 1, 2, 3 of month 07:00
```

before fix:

```
MAILTO=""
```

after fix:

```
MAILTO=""
0 7 1,2,3 * * root curl -sSH "X-Appengine-Cron:true" -H "X-AppEngine-Fake-Is-Admin:XXX" -k -L "http://192.168.102.107:8080/example" 2>&1 >> /var/log/appscale/cron-guestbook.log
```

so the item is no longer dropped.

PR #3039, opened 2019-04-04
Fix python detection in taskqueue tests

scragraham:
**Technical Problem Description**
This line was writing output to the **file** null for stderr.
Python 2.x --version writes **only** to stderr.
Python 3.x --version writes **only** to stdout. 
```
HAVE=$(${PYTHON_EXECUTABLE} --version 2> null | awk '{ print $2 }')
```
This caused any python 2 versions to make the $HAVE variable an empty string. Which technically works...but is not correct.

**Summary of changes**
- dups stderr to stdout and pipe to awk and does not create the 'null' file.
- Skips executables that don't exist on the system. Previously $HAVE would contain unexpected output if the executable (python3.7 at the time) didn't exist.
- Added log message indicating which python version we picked up from the system.

**Demo**
```
2019-04-03 22:35:33: INFO 
2019-04-03 22:35:33: INFO ==================================================
2019-04-03 22:35:33: INFO === Prepare virtualenv for running test script ===
2019-04-03 22:35:33: INFO ==================================================
2019-04-03 22:35:33: INFO Using python: /home/jenkins/.local/bin//python3.7 version: 3.7.3
```

### DIFF
--- a/AppTaskQueue/test/suites/run-e2e-tests.sh
+++ b/AppTaskQueue/test/suites/run-e2e-tests.sh
@@ -176,16 +176,18 @@ log "=================================================="
 PYTHON=
 for PYTHON_EXECUTABLE in python python3 python3.6 python3.7
 do
-    # Get exact version of python interpreter
+    # Skip python executables that don't exist in PATH
+    if ! which ${PYTHON_EXECUTABLE}; then
+        continue
+    fi
+
     possible_python=$(which ${PYTHON_EXECUTABLE})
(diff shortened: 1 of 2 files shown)


scragraham:
hold off on merging this until we pull in the release branch

scragraham:
release-3.7 has been merged to master so this should be good to merge now. I moved the base branch back and forth so that it would pick up the fact that some of the commits were already in master.

PR #3040, opened 2019-04-05
Move Hermes from tornado to aiohttp

jeanleonov:
This branch reimplements Hermes to use aiohttp instead of torando.

You may say that it changes nothing but 45 files. We don't get new features and don't resolve any bugs.
The major reasonable change here is that Hermes become python3 program.

This PR is more about new experiance and new coding practice which can be reused later.
What's new?
 - aiohttp server.
 - aiohttp client.
 - asyncio unix socket.
 - asynchronouse tests with pytest.

Side changes:
 - Hermes profiler was removed.
 - AdminsServer main module was moved from __init__.py file in order to avoid unnecassery import of all its dependencies when you need just a constant.
 - appscale_info.get_load_balancer_ips was changed to be python3 compatible.


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6087/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6089/

PR #3042, opened 2019-04-09
Just a new directory containing SearchService2

jeanleonov:
The service is not integrated to appscale in any way in this branch.

http://ocd.appscale.net:8080/job/Daily%20Build/6105/


SearchService2/setup.py
```diff
@@ -0,0 +1,38 @@
...
+  description='An implementation of the Google App Engine Search API',
+  author='AppScale Systems, Inc.',
+  url='https://github.com/AppScale/appscale',
+  license='Apache License 2.0',
+  keywords='appscale google-app-engine python',
+  platforms='Posix',
+  install_requires=[
+    'appscale-common>=0.0.4',
+    'kazoo==2.6.0',
+    'tornado==5.1.1',
```
cdonati: Do these all need to be pinned to that specific version? Can they be a range?

jeanleonov: As SearchService2 is run in dedicated venv, I think it's reasonable to be as specific as possible.
Whenever we want get use of something new from libraries we can change it and it will be more explicite.

cdonati: Okay, I'm fine with this for now. In the future, we may want to consider introducing a requirements file for this reason. https://packaging.python.org/discussions/install-requires-vs-requirements/ recommends not using install_requires to pin dependencies to specific versions.

jeanleonov: Thanks for the reference to the discussion, I didn't know that `requirements.txt` is preferable over `install_requires`.

PR #3043, opened 2019-04-09
Retry 3 times with backoff on socket err in AppServer

jeanleonov:
The change is applied only for python runtime for requests to Datastore
and Taskqueue.


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6107/

AppServer/google/appengine/api/taskqueue/taskqueue_distributed.py
```diff
@@ -102,6 +106,7 @@ def __init__(self, app_id, host, service_name='taskqueue'):
         service_name, max_request_size=MAX_REQUEST_SIZE)
     self.__app_id = app_id
     self.__nginx_host = host
+    self.__tq_locations = self._GetTQLocations()
```
cdonati: This has the downside of requiring a process restart if the locations file changes, but I think that's acceptable for now.

jeanleonov: Good point.
If nobody minds, I'd prefer to keep it as it is here.
Later we'd have to implement zookeeper watches for services locations in AppServer.

PR #3048, opened 2019-04-12
Removed cached local IP

obino:
HelperFunctions was using a class variable to track the local IP, but it
was not quite a cache, since it was set by djinn (or by calling
local_ip). This variable was used sparingly (starting zookeeper and for
the datastore): since not all services were using it, this could in
theory cause some inconsistencies on multi-networked systems. This PR
removes the inconsistency and print warnings if djinn is using a local
IP that is not the default one.


obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6124/

AppController/lib/helperfunctions.rb
```diff
@@ -71,10 +63,6 @@ module HelperFunctions
                       'm' => 60,
                       's' => 1 }.freeze
 
-  CLOUDY_CREDS = %w[ec2_access_key ec2_secret_key EC2_ACCESS_KEY
```
cdonati: This is still used in obscure_array.

AppController/djinn.rb
```diff
@@ -3098,9 +3098,12 @@ def find_me_in_locations
         all_local_ips.each { |ip|
           if ip == node.private_ip
             @my_index = index
-            HelperFunctions.set_local_ip(node.private_ip)
             @my_public_ip = node.public_ip
             @my_private_ip = node.private_ip
+            if ip != HelperFunctions.local_ip
```
cdonati: This is kind of weird since `all_local_ips` is populated by `get_all_local_ips`. This `local_ip` method also uses `get_all_local_ips`. I think this would be a bit more understandable if we avoid the call to `local_ip` and just check the index.
```
all_local_ips.each_with_index { |ip, index|
  ...
  if index != 0
```

obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6139/

AppController/lib/helperfunctions.rb
obino: yes indeed! will remove them soon,

AppController/djinn.rb
obino: well I could remove that check altogether: I wanted to leave a warning in case this code path is taken (I haven't seen it in a very long time to be honest) just in case we may pick the wrong IP in a multi-ethernet device configuration. Thoughts? should I just remove the check?

cdonati: I would be fine with either removing the warning log or removing the call to `local_ip`.

obino: I removed the use of local_ip (in this function, still one to go) and added a warning when mulitple IPs are detected.

AppController/djinn.rb
```diff
@@ -3089,18 +3089,19 @@ def got_all_data
   def find_me_in_locations
     @my_index = nil
     all_local_ips = HelperFunctions.get_all_local_ips
-    Djinn.log_debug("Searching for a node with any of these private IPs: " \
-      "#{all_local_ips.join(', ')}")
-    Djinn.log_debug("All nodes are: #{@nodes.join(', ')}")
+    if all_local_ips.length > 1
+      Djinn.log_warn("These node has multiple private IPs: " \
```
cdonati: "This node"

PR #3052, opened 2019-04-20
Use pymemcache to implement the memcache API

cdonati:
This makes the following major changes:
- Adds connection pooling to reduce number of created connections to memcached backends
- Fixes a bug that made CAS operations unsafe
- Uses native memcached primitives instead of altering the stored value
- Makes `incr` and `decr` behave more like GAE


AppServer/google/appengine/api/memcache/memcache_distributed.py
```diff
@@ -262,58 +350,33 @@ def _Dynamic_Stats(self, request, response):
       request: A MemcacheStatsRequest.
       response: A MemcacheStatsResponse.
     """
+    # TODO: Gather stats for a project rather than the deployment.
+    hits = 0
+    misses = 0
+    byte_hits = 0
+    items = 0
+    bytes = 0
```
sjones4: could use a different variable name

https://docs.python.org/3/library/functions.html#func-bytes

sjones4:
Current docs for the python runtime state:

> A key cannot be larger than 250 bytes. In the Python runtime, keys that are strings longer than 250 bytes will be hashed. (Other runtimes behave differently.)

https://cloud.google.com/appengine/docs/standard/python/memcache/

cdonati:
@sjones4 The hashing behavior I believe is runtime-specific and only performed by the client. For example, our Python runtime still does that here: https://github.com/AppScale/appscale/blob/c3c8432f0215ced6f672dbb85f6aa8a5f508ade1/AppServer/google/appengine/api/memcache/__init__.py#L172-L173

If the runtime directly sends a protobuf request that contains an excessively long key, the API reports an error rather than trying to hash it.

However, the prefix we add for project_id and namespace could result in a situation where the key itself is less than 250 bytes but the prefixed key is longer than 250 bytes. In GAE, the namespace does not affect the limit. To achieve the same behavior, do you think we should do the following?
 - Return an error if the key is > 250 bytes
 - Hash the prefixed key if the key is <= 250 bytes and the prefixed key > 250 bytes

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6289/

PR #3060, opened 2019-05-01
Support for autoscaling  idle instances

obino:
Added support for min_idle_instances. The parameters max_idle_instances and max_concurrent_requets are recognized, but not acted upon yet.


obino:
Closing for now: while ti works for basic support, it is a bit rough on corner cases like node failures.

PR #3062, opened 2019-05-06
Add support for idle instances

obino:
Add support for min-idle-instances in application configuration. Idle instances will be kept available and used when up-scale is happening.max-idle-instances is not used yet: at this time only min-idle-instances will be kept running in the system.


cdonati:
As far as I can tell, there isn't anything here to prevent idle instances from being routed.

obino:
@cdonati let me know if this way to avoid routing is what you had in mind.

obino:
Closing this PR for now: it will be reopened when reworked.

PR #3065, opened 2019-05-10
Blobstore application id header and relative success url

sjones4:
Update blobstore / java runtime to pass the application id in an HTTP header and to allow use of relative url in the success path for create upload url.

These changes reduce the differences between the java and python runtimes, allowing the java8 runtime to reuse existing functionality.


sjones4:
Daily build no. 6210

PR #3066, opened 2019-05-13
Fix the Ruby client's get and set property methods

cdonati:
This also removes the unused `get_queues_in_use` method.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6226/

PR #3068, opened 2019-05-14
AppServer ensure request type set for request

sjones4:
Update for the python runtime to ensure the request type in the environment is correct.


sjones4:
Daily build no. 6223

PR #3070, opened 2019-05-15
Remove local_ip helper function

cdonati:
The function could return an IP address that was different than what the controller was given as a private IP. This was problematic for our default Vagrant setup since the Cassandra seed node IP is determined from the node layout, but the local IP was determined from the first entry from `ifconfig`. Cassandra, believing the seed node to be different than itself, would incorrectly try to bootstrap from a different node even though it was the only one in the cluster.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6244/

PR #3072, opened 2019-05-15
Dont override hostname

obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6228/


AppController/lib/taskqueue.rb
```diff
@@ -168,21 +167,25 @@ def self.start_slave(master_ip, clear_data, verbose)
...
-        # We couldn't get the name: let's try to use the IP address.
-        master_tq_host = master_ip
+    # We need not to contact the master node, and we need to use its name
+    # (just the hostname not the fqdn).  To resolve it we use
+    # Addrinfo.getnameingo since Resolv will not use /etc/hosts and this
+    # could cause issues on private clusters.
+    master_tq_host = nil
+    begin
+      master_tq_host = Addrinfo.ip(master_ip).getnameinfo[0]
+      if (master_tq_host =~ /[[:digit:]]/).nil?
```
cdonati: I don't understand the regex here. Why does the fqdn need to have a digit?

obino: It doesn't: if the results starts with a number it is an IP address (since no DNS entry can start with a number) hence the warning about using an IP instead of a hostname (which is what rabbitmq wants).

cdonati: Your regex appears to be testing whether or not the fqdn has a digit anywhere in the string. For example,
```
irb(main):001:0> "ubuntu-bionic" =~ /[[:digit:]]/
=> nil
```
So if the fqdn does not contain a digit, it will log a warning and use the fqdn. If it does contain a digit, it will use just the hostname.

scragraham:
```suggestion
      if (master_tq_host =~ /^[[:digit:]]/).nil?
```

scragraham: Am I reading the if stmt wrong? My ruby-fu is weak...
```
if (master_tq_host =~ /^[[:digit:]]/).nil?
```
Doesn't that mean if the master_tq_host doesn't match the regex? eg the regex match is nil. So the if stmt would be true if its a hostname?

```
irb(main):008:0> if ("www.google.com" =~ /^[[:digit:]]/).nil?
irb(main):009:1>    puts "I'm a hostname"
irb(main):010:1> else
irb(main):011:1>    puts "I'm an ip address"
irb(main):012:1> end
I'm a hostname
```
I think the if/else blocks need to be reversed.

Actually maybe this is clearer:
```
# Warn if the hostname is an IP address
if /^[[:digit:]]{1}/.match?(master_tq_host)
    Djinn.log_warn("#{master_ip} didn't resolve to a hostname! Expect" \
                            " problems with rabbitmq clustering.")
else
    master_tq_host = master_tq_host.split('.')[0]
```

obino: ah yes! thanks you! Forgot to change the PR: the intention was to have something like unless (master_tq_host =~ /^[[:digit:]]/).nil? let me fix it.

AppController/lib/taskqueue.rb
```diff
@@ -168,21 +167,24 @@ def self.start_slave(master_ip, clear_data, verbose)
...
-      rescue Resolv::ResolvError
-        # We couldn't get the name: let's try to use the IP address.
-        master_tq_host = master_ip
+    # Look up the TaskQueue master's hostname (not the fqdn). To resolve
+    # it we use Addrinfo.getnameinfo since Resolv will not use /etc/hosts
+    # and this could cause issues on private clusters.
+    master_tq_host = nil
+    begin
+      master_tq_host = Addrinfo.ip(master_ip).getnameinfo[0]
+      unless (master_tq_host =~ /^[[:digit:]]/).nil?
```
cdonati: `unless/else` is like a double negative, which makes it more difficult to read. In this case, I think
```
if master_tq_host =~ /^[[:digit:]]/
  Djinn.log_warn...
else
  master_tq_host =...
end
```
would be clearer.

obino:
Azure test: http://ocd.appscale.net:8080/job/Create-Azure-Image/594/
GCE test: http://ocd.appscale.net:8080/job/Create-GCE-Image/713/
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6257/
EC2 buils has some issues with xmpp test: http://ocd.appscale.net:8080/job/Ec2-Testing/75/

obino:
EC2 test (the failure is the possible race condition of the test not the timeout) http://ocd.appscale.net:8080/job/EC2-Testing/79/

obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6280/
GCE build http://ocd.appscale.net:8080/job/GCE-Testing/45/ (one of the failure is the one we are looking into it that we see at times)

PR #3074, opened 2019-05-17
More determined stop

obino:
This PR ensures that there are no leftover haproxy processes leftover after a down.


AppController/terminate.rb
```diff
@@ -78,6 +78,23 @@ def self.erase_appscale_state
     print "OK"
   end
 
+  # This functions ensure that the services AppScale started that have a
+  # PID in /var/run/appscale got terminated.
+  def self.ensure_services_are_stopped
+    Dir["/var/run/appscale/*.pid"].each { |pidfile|
```
cdonati: I'm a little concerned that very old pidfiles could contain PIDs that might have been reused by other processes.

AppController/djinn.rb
```diff
@@ -711,11 +711,34 @@ def kill(stop_deployment, secret)
           end
         }
       }
-      Djinn.log_info('Waiting for other nodes to acknoledge stop ... ')
-      threads.each { |t| t.join }
     end
 
-    Djinn.log_info('---- Stopping AppController ----')
+    Djinn.log_devug("Stopping local services ...")
```
cdonati: I think we should settle on a single method of stopping services. If both the controller and the terminate script have a similar method of stopping services, we need to keep both in sync. In addition, we would need to make sure they both aren't trying to stop stuff at the same time since that can confuse monit.

To me, it would make sense if sigterm just stopped the controller process rather than all the services on the machine.

AppController/djinn.rb
```diff
@@ -3481,6 +3417,11 @@ def start_hermes
     Djinn.log_info("Done starting Hermes service.")
   end
 
+  def stop_hermes
```
cdonati: This new method no longer appears to be needed.

AppController/djinn.rb
```diff
@@ -4436,6 +4372,11 @@ def start_admin_server
     end
   end
 
+  def stop_admin_server
```
cdonati: No longer used.

AppController/lib/haproxy.rb
```diff
@@ -78,13 +78,23 @@ module HAProxy
...
+    return if MonitInterface.is_running?(:service_haproxy)
+
     start_cmd = "#{HAPROXY_BIN} -f #{SERVICE_MAIN_FILE} -D " \
       "-p #{SERVICE_PIDFILE}"
     stop_cmd = "#{BASH_BIN} -c 'kill $(cat #{SERVICE_PIDFILE})'"
     MonitInterface.start_daemon(
       :service_haproxy, start_cmd, stop_cmd, SERVICE_PIDFILE)
   end
 
+  def self.services_stop
```
cdonati: No longer used.

AppController/terminate.rb
```diff
@@ -7,18 +7,19 @@
...
 
 module TerminateHelper
-
-
   # Erases all AppScale-related files (except database state) from the local
   # filesystem. This is used when appscale is shutdown.
   # TODO: Use FileUtils.rm_rf instead of backticks throughout this
   # method.
   def self.erase_appscale_state
+    `service appscale-controller stop`
```
cdonati: monit will probably just restart the controller right after this. I don't think it's needed since `appscale-stop-services` should stop the controller.

AppController/terminate.rb
```diff
@@ -7,18 +7,19 @@
...
+
     `rm -f #{APPSCALE_CONFIG_DIR}/secret.key`
     `rm -f /tmp/uploaded-apps`
     `rm -f ~/.appscale_cookies`
     `rm -f /etc/nginx/sites-enabled/appscale-*.conf`
-    `rm -f /etc/haproxy/sites-enabled/*.cfg`
+    `rm -f /etc/haproxy/app-site-enabled/*.cfg`
```
cdonati: I think this should be "app-sites-enabled". Also, I would double-check if this is still needed. Looks like there's a reference in Hermes, but the RoutingManager no longer uses intermediate files to compile the final config.

AppController/djinn.rb
```diff
@@ -4352,8 +4283,8 @@ def set_appcontroller_monit
     Djinn.log_debug("Configuring AppController monit.")
     service = `which service`.chomp
     start_cmd = "#{service} appscale-controller start"
-    stop_cmd = "#{service} appscale-controller stop"
     pidfile = '/var/run/appscale/controller.pid'
+    stop_cmd = "/sbin/start-stop-daemon --stop --pidfile #{pidfile} --retry=TERM/30/KILL/5"
```
cdonati: Should the path be resolved first with `which` like we do for other services, or do we have a guarantee that it will always be in /sbin?

AppController/terminate.rb
obino: you are right, not used anymore, I will remove it.

AppController/djinn.rb
obino: for now yes: we already have another hadcoded close by (service) I would say that we can revisit once we switch to systemd?

obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6309/

obino:
Re-run autoscaling test: http://ocd.appscale.net:8080/job/Autoscale-test-manual-triggers/2542/

PR #3076, opened 2019-05-20
Allow lower min machines

obino:


AppController/djinn.rb
```diff
@@ -2514,6 +2514,18 @@ def get_all_compute_nodes
     return ae_nodes
   end
 
+  # This method checks that nodes above index are compute only and thus
+  # can be easily terminated.
+  def can_we_scale_down?(index)
+    @state_change_lock.synchronize {
+      nodes_to_check = @nodes.drop(min_machines)
```
cdonati: I'm guessing this was supposed to be `index`? Between the two names, `min_machines` makes more sense to me from an interface perspective.

PR #3079, opened 2019-05-22
Better venvs support in AppController

jeanleonov:
APIServer was venv pioneer in AppScale. But soon we'll get 3 more services running in own venvs (SearchService2, Hermes on aiohttp and TaskQueue). This PR aims to simplify a bit handling services running in separate venvs.

Changes in details:
 - Move ApiServer venv to /opt/appscale_venvs/api_server.
 - Replace number of build_(hermes/taskqueue/datastore/...) with one
   common function update_python_package with optional pip parameter.
 - Save `pip install` logs to log file.

Later we may want to rework python packages update deeper. In particular we should not rely on `git status` as a marker of code change.


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6260/

jeanleonov:
@sjones4 I think we do need a directory containing all venvs we use. At the moment it contains only API server venv, but Search2, Hermes and TQ venvs are coming soon. I wouldn't like to have them all directly in `/opt/` or `/opt/appscale`.
It seems to be a good practice to have all venvs in a single place, though I'm not sure if `/opt/appscale_venvs/` is a right place.

PR #3080, opened 2019-05-23
Resolving dependencies between tq and adminserver

Honcharov12:
Resolves problem that AdminServer uses some constants from AppTaskQueue that not used anywhere except for AdminServer. Also creates some new exception classes to make it more clear.

### DIFF
--- a/AdminServer/appscale/admin/utils.py
+++ b/AdminServer/appscale/admin/utils.py
@@ -10,10 +10,8 @@ import socket
 import tarfile
 
 from appscale.common.constants import HTTPCodes
-from appscale.common.constants import InvalidConfiguration
 from appscale.common.constants import VERSION_PATH_SEPARATOR
 from appscale.taskqueue import constants as tq_constants
-from appscale.taskqueue.constants import InvalidQueueConfiguration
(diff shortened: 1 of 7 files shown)


Honcharov12:
http://ocd.appscale.net:8080/job/Daily%20Build/6283/

AdminServer/appscale/admin/utils.py
```diff
@@ -23,7 +20,11 @@
   JAVA,
   SOURCES_DIRECTORY,
   Types,
-  UNPACK_ROOT
+  UNPACK_ROOT,
+  REQUIRED_PULL_QUEUE_FIELDS,
+  REQUIRED_PUSH_QUEUE_FIELDS,
+  SUPPORTED_PULL_QUEUE_FIELDS,
+  SUPPORTED_PUSH_QUEUE_FIELDS
 )
```
jeanleonov: It's better to use a single import style at least in scope of a single module. So it looks reasonable to import all needed from constants as it's done here (`from .constants import (...)`) and not use `from . import constants`.
Another benefit of doing so will be using one-line raise exception statements.

PR #3082, opened 2019-05-27
Sending logs to logstash

jeanleonov:
This PR brings ELK logging functionality to python AppServer.
We have being using it on separate deployments since a beginning of 2018.
Before publishing the PR I added a way to make this duplicated logging to JSON files optional and not default.
So if you want to enable logstash logging, you need to run:
`bash /root/appscale/scripts/init-filebeat.sh --logstash <elk-ip>:5044`

### DIFF
--- a/AppServer/google/appengine/api/logservice/logservice_stub.py
+++ b/AppServer/google/appengine/api/logservice/logservice_stub.py
@@ -18,6 +18,8 @@
 
 
 import base64
+from logging.handlers import WatchedFileHandler
+
 import capnp # pylint: disable=unused-import
 import logging
 
@@ -32,8 +34,6 @@ from collections import defaultdict
 import os
 
 from datetime import datetime
-
-import multiprocessing
 
 import threading
 from google.appengine.api import apiproxy_stub
@@ -69,11 +69,11 @@ class RequestsLogger(threading.Thread):
   def __init__(self):
     super(RequestsLogger, self).__init__()
     self.setDaemon(True)
-    self._logs_queue = multiprocessing.Queue(self.QUEUE_SIZE)
-    self._log_file = None
+    self._logs_queue = Queue(self.QUEUE_SIZE)
+    self._logger = None
     self._shutting_down = False
 
-  def _open_log_file(self, request_info):
+  def _init_logger(self, request_info):
     # Init logger lazily when application info is available
     app_id = request_info['appId']
     service_id = request_info['serviceName']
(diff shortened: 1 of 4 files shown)


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6293/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6297/

AppServer/google/appengine/api/logservice/logservice_stub.py
```diff
@@ -20,30 +20,127 @@
...
+          self._log_file.flush()
+          request_info = None
+
+        except (OSError, IOError):
+          # Close file to reopen it again later
+          logging.exception(
+            'Failed to write request_info to log file\n  Request info: {}'
+            .format(request_info or "-"))
+          log_file = self._log_file
+          self._log_file = None
```
cdonati: Would the Python garbage collector take care of closing the descriptor stored at `log_file`? Otherwise, this seems like it could leave behind open file descriptors. For example, if the disk is full, the `.close()` method may raise an exception, and it appears that a new open file descriptor would be created every 5 seconds.

AppServer/google/appengine/api/logservice/logservice_stub.py
```diff
@@ -238,6 +348,59 @@ def end_request(self, request_id, status, response_size, end_time=None):
...
+          )
+          for log in rl.appLogs
+        ])
+      except UnicodeError:
+        app_logs_str = u'\n'.join([
+          u'{} {} {}'.format(
+            LEVELS[log.level],
+            datetime.utcfromtimestamp(log.time/1000000)
+              .strftime('%Y-%m-%d %H:%M:%S'),
+            unicode(log.message, 'ascii', 'ignore')
```
cdonati: It seems like you could place this above (or just the byte string's .decode method) and omit the try since you are coercing bytes to unicode above anyway. I would also recommend using the 'utf-8' codec to avoid stripping non-ascii characters.

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6442/

jeanleonov:
Btw.. I realized that all problems of this ELK integration implementation can be resolved in one obvious way:
logservice should just resend received logs to ELK directly.
As new FDB-based implementation of logservice is not in master yet, I'd move ELK integration to logservice when it's there.

PR #3083, opened 2019-05-27
Backup-restore performance and reliability improvements

jeanleonov:
- Try to stay under the large batch threshold
   (during a restore operation, large batches are significantly slower to apply).
 - Retry batch puts during a restore.
 - Skip redundant deserialization and serialization.


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6298/

PR #3084, opened 2019-05-28
Limit version of cassandra-driver

cdonati:
This avoids using version 3.18.0, which does not seem to be able to connect to the cluster properly.


jeanleonov:
Looks good to me, though I'd prefer if we used fixed package version where possible (3.17.1 worked for me).

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6299/ (Demo is that the hawkeye jobs succeed)

PR #3086, opened 2019-05-28
AppServer admin server module should not be enabled for python runtime

sjones4:
Do not enable administration interface for python runtime as this is not used.


jeanleonov:
Good cleanup and seems to be nice performance improvement.
I may be missing something, but side effect of this change would be that datastore viewer won't work as it uses API at admin port. The viewer is disabled by default, but some of appscale users uses this feature.

sjones4:
Daily build no. 6287

cdonati:
@jeanleonov I believe the dashboard now covers all of the datastore viewer functionality that the SDK's datastore viewer provides.

PR #3087, opened 2019-05-29
Do not modify /root file mode on install

sjones4:
Remove  `/root` file mode change from nginx post install step. This is not needed as nginx no longer uses files under `/root`.


sjones4:
Daily build no. 6301

PR #3091, opened 2019-06-04
Stop services concurrently

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6324/

PR #3095, opened 2019-06-20
Don't trap SIGTERM and let the process stops

obino:
Since the AC is not doing any prticular action on normal stop (that is
SIGTERM reception) we don't need to trap it anymore, and let service
appscale-controller stop do its magic.

### DIFF
--- a/AppController/djinnServer.rb
+++ b/AppController/djinnServer.rb
@@ -145,11 +145,6 @@ server = DjinnServer.new(
   :SSLCertName => nil
 )
 
-trap('TERM') {
-  Djinn.log_debug("Received TERM signal: stopping AppController.")
-  server.shutdown
-}
-
 new_thread = Thread.new { run_server(server) }
 server.djinn.job_start(secret)
 new_thread.join()


AppController/djinnServer.rb
```diff
@@ -145,11 +145,6 @@ def run_server(server)
   :SSLCertName => nil
 )
 
-trap('TERM') {
-  Djinn.log_debug("Received TERM signal: stopping AppController.")
-  server.shutdown
```
scragraham: What does sever.shutdown do? If we remove this code is the process not calling server.shutdown and possibly not shutting down ports cleanly?

obino: That was to shut down the SOAP server: after that call for example you will see that appscale status will error out since the controller is up but cannot be reached. No port are leftover, here is the output of netstat before and after stopping the controller (the process pid is 5635):

root@euca-10-10-6-12:~# netstat -alnp|grep -E '(5635|17443)'
tcp        0      0 0.0.0.0:17443           0.0.0.0:*               LISTEN      5635/ruby       
tcp        0      0 10.10.6.12:45305        10.10.6.12:2181         ESTABLISHED 5635/ruby       
tcp       32      0 10.10.6.12:39766        10.10.6.12:4343         CLOSE_WAIT  5635/ruby       
root@euca-10-10-6-12:~# netstat -alnp|grep -E '(5635|17443)'
root@euca-10-10-6-12:~#

jeanleonov:
Catching `'EXIT'` (special word for ruby's trap) would resolve the issue:
https://github.com/AppScale/appscale/pull/3091/files#diff-4020412951682f6648be0012f9b3eb93R148

obino:
Thanks @jeanleonov much better solution! Closing this now.

PR #3096, opened 2019-06-28
Handle scheme redirects with the dispatcher

cdonati:
This is less efficient than giving nginx the burden of handling client redirects. However, it makes it much simpler to implement the dispatch configuration, and the instance dispatcher can later be optimized if necessary.

Since the dispatcher is already handling static files (usually) and login redirects, it makes sense for it to handle scheme redirects as well.

### DIFF
--- a/AppServer/google/appengine/tools/devappserver2/module.py
+++ b/AppServer/google/appengine/tools/devappserver2/module.py
@@ -674,6 +674,13 @@ class Module(object):
         for handler in handlers:
           match = handler.match(request_url)
           if match:
+
+            # AppScale: Reject requests with the wrong scheme.
+            redirect_response = self._handle_redirect(
+                handler, environ, wrapped_start_response)
+            if redirect_response is not None:
+              return redirect_response
+
             auth_failure = handler.handle_authorization(environ,
                                                         wrapped_start_response)
             if auth_failure is not None:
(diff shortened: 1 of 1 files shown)


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6400/

cdonati:
@jeanleonov Yes, we do already have these redirect handlers in nginx. They enforce the same rules. If a redirect handlers is used in nginx, the request will not go through to any instance dispatchers, so there is no interaction or conflict there.

After this goes in, the plan is to remove the redirect and static handlers from nginx, which will greatly simplify the generated config files there.

cdonati:
Demo
```
$ curl -v -H "X-Forwarded-Proto: http" 192.168.33.10:20001/secure-always
< HTTP/1.1 302 Moved Temporarily
< Location: https://192.168.33.10:20001/secure-always
$ curl -v -H "X-Forwarded-Proto: https" 192.168.33.10:20001/secure-always
< HTTP/1.1 200 OK
$ curl -v -H "X-Forwarded-Proto: http" 192.168.33.10:20001/secure-never
< HTTP/1.1 200 OK
$ curl -v -H "X-Forwarded-Proto: https" 192.168.33.10:20001/secure-never
< HTTP/1.1 302 Moved Temporarily
< Location: http://192.168.33.10:20001/secure-never
$ curl -v -H "X-Forwarded-Proto: http" 192.168.33.10:20001/secure-optional
< HTTP/1.1 200 OK
$ curl -v -H "X-Forwarded-Proto: https" 192.168.33.10:20001/secure-optional
< HTTP/1.1 200 OK
```

PR #3097, opened 2019-07-01
Remove M2Crypto dependency

cdonati:


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6405/

scripts/create_user.py
```diff
@@ -1,7 +1,7 @@
 """ This script creates a new user. Fails if the user already exists. """
 
 import hashlib
-import M2Crypto
+import random
```
jeanleonov: I think `random` can't be used for password generation. `os.urandom` could be used instead with some modification which would limit vocabulary. 

https://docs.python.org/2/library/random.html
> However, being completely deterministic, it is not suitable for all purposes, and is completely unsuitable for cryptographic purposes.

https://docs.python.org/2/library/os.html#os.urandom
> Return a string of n random bytes suitable for cryptographic use.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6537/

PR #3101, opened 2019-07-11
Fix non-cloud mode of AppScale launch.

Honcharov12:
Fixes problem when starting AppScale was not allowed because of `max_machines` option.


Honcharov12:
http://ocd.appscale.net:8080/job/Daily%20Build/6440/

PR #3107, opened 2019-07-22
Add explicit dependency on sudo

obino:
Avoid issues like http://ocd.appscale.net:8080/job/Create-Docker-Container/5809/console


obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6521/

sjones4:
Why do we need `sudo`? seems better to not use it.

scragraham:
I believe @sjones4 is correct, we shouldn't need sudo.

obino:
At this time sudo is used in few places, as part of the management that we talked about just this morning. I agree it would be nice to not needing it, but I'm not sure there is a way out of it easily. For not this PR makes it explicit so docker isn't failing out right.

obino:
you are right, I think we can simply remove sudo from the install script. I will close this PR and create another one. Thanks @sjones4

PR #3109, opened 2019-07-23
Support custom service accounts for datastore_admin backups

cdonati:
Apologies for the number of commits. I wanted to maintain readable diffs for each version of the SDK upgrade. Here's a summary of the changes:

1. Upgrade the datastore_admin and mapreduce builtins to the latest version. This ensures the object output format is compatible with GAE.
2. Bundle the appengine-gcs-client library to power the updated builtins.
3. Add support for the `service_account_name` field for `GetAccessTokenRequest`s.
4. Modify the datastore_admin backup for to support a custom service account.


AppServer/google/appengine/ext/datastore_admin/remote_api_put_stub.py
```diff
@@ -236,8 +276,8 @@ def get_remote_appid(remote_url, extra_headers=None):
   if not response.startswith('{'):
     logging.info('Response unparasable: %s', response)
     raise ConfigurationError(
-        'Invalid response recieved from server: %s' % response)
-  app_info = yaml.safe_load(response)
+        'Invalid response received from server: %s' % response)
+  app_info = yaml.load(response)
   if not app_info or 'rtok' not in app_info or 'app_id' not in app_info:
```
sjones4: Could keep `yaml.safe_load` here

APIServer/appscale/api_server/app_identity.py
```diff
@@ -130,19 +136,55 @@ def get_access_token(self, scopes, service_account_id=None,
         Raises:
             UnknownError if the service account is not configured.
         """
-        if self._key is None:
```
Honcharov12: Moving it to https://github.com/AppScale/appscale/blob/cd3bcc68ccaefa66af4e0366cfcdaea49e7664d7/APIServer/appscale/api_server/app_identity.py#L182-L183 seems that `if service_account_name is not None` then `self._key is not None` anyway?

cdonati: Good catch. I modified it a bit so that a user can still get an access token even if `self._key` is None by specifying a custom key name.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6605/
The hawkeye tests failed since they are trying to verify the signature using the wrong padding. Here is a build with https://github.com/AppScale/hawkeye/pull/171:
http://ocd.appscale.net:8080/job/Hawkeye/19385/

PR #3111, opened 2019-07-23
Be more flexible when selecting a composite index

cdonati:
This allows the datastore to use an index that has a prefix (initial properties that match equality filters) defined in a different order than the query.

It also sometimes allows using an index that is defined in a different direction than the query order.


cdonati:
This does not affect the current tests, so I won't post a build.

Here is a demo of the IndexVersatility test from hawkeye:
master: runTest (tests.datastore_tests.IndexVersatility) ... FAIL
branch: runTest (tests.datastore_tests.IndexVersatility) ... ok

PR #3112, opened 2019-07-24
Search backup restore

jeanleonov:


SearchService2/appscale/search/backup_restore/storage.py
```diff
@@ -0,0 +1,128 @@
...
+      index_name: a str - search index name.
+      index_documents_pb: a search_pb2.IndexDocumentRequest.
+    """
+    if not namespace:
+      namespace = '__default__'
+    first_doc_id = index_documents_pb.params.document[0].id
+    key = '{}/{}/{}/{}...'.format(
+      project_id, namespace, index_name, first_doc_id
+    )
+    body = index_documents_pb.SerializeToString()
```
cdonati: One downside to this storage format is that it requires the stored objects to be small enough to fit into memory. If someone has a ridiculous amount of documents, this might make paging through the relevant objects a lengthy process. It also impacts the API costs since PUT and GET operations are charged per object.

I don't think that's a large enough concern to block this. I just wanted to mention it as a suggestion for future improvements.

jeanleonov: Good point. I was kind of worried about having huge amount of small objects, but I haven't though about how S3 operations are charged.
I totally vote for improving this part or at least taking it in account when we develop another backup tool.

At the moment I'd keep it as simple as it is. Let's see how it serves users' needs.

SearchService2/appscale/search/backup_restore/backup_from_v2.py
```diff
@@ -0,0 +1,287 @@
...
+    self.start_time = self.ioloop.time()
+    self.status = 'In progress'
+
+    collections, broken = await self.solr_adapter.solr.list_collections()
+    if broken:
+      logger.warning('There are {} broken collections: {}. It will be ignored.'
+                     .format(len(broken), list(broken)))
+    for collection_name in collections:
+      if not collection_name.startswith('appscale_'):
+        logger.info('Collection {} does not belong to Search Service. Ignoring.'
```
Honcharov12: Where is it ignored? :)

SearchService2/appscale/search/backup_restore/storage.py
```diff
@@ -0,0 +1,128 @@
...
+    Args:
+      project_id: a str - GAE project ID.
+      namespace: a str - namespace name.
+      index_name: a str - search index name.
+      index_documents_pb: a search_pb2.IndexDocumentRequest.
+    """
+    if not namespace:
+      namespace = '__default__'
+    first_doc_id = index_documents_pb.params.document[0].id
+    key = '{}/{}/{}/{}...'.format(
```
scragraham: why the '...' in the keyname?

jeanleonov: Because it allowed and it emphasizes that the document ID is the first in range of those documents which are hold in a blob.

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6623/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6624/

PR #3113, opened 2019-07-26
Add a rough interface for managing service accounts

cdonati:
This adds a page to the dashboard that allows users to list a project's service accounts and upload custom ones.


cdonati:
@jeanleonov Yes, I think the IAM API should later be served via a different process and port. I left it in the AdminServer for now mostly due to memory constraints. As soon as we remove our Cassandra dependency, I'd like to separate the Admin API, Cloud Resource Manager API, and IAM API to separate packages.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6589/

PR #3116, opened 2019-07-30
Flatten path before extracting the entity group

cdonati:
This fixes an issue with the FDB backend when making a query inside a transaction.

This also:
- Fixes a bug where incomplete entities were not auto-assigned IDs when created in a transaction
- Applies mutations concurrently on transaction commit
- Fixes an issue where index entries could be left behind if an entity is mutated multiple times in a transaction
- Adds an additional check when passing old entries to the garbage collector to be deleted later

### DIFF
--- a/AppDB/appscale/datastore/fdb/fdb_datastore.py
+++ b/AppDB/appscale/datastore/fdb/fdb_datastore.py
@@ -12,6 +12,7 @@ transactions: transaction metadata
 See each submodule for more implementation details.
 """
 import logging
+import six
 import sys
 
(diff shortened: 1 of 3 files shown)


cdonati:
This does not affect the current tests, so I won't post a build.

Here is a demo using https://github.com/AppScale/hawkeye/pull/169:
master: test_query_in_transaction (tests.datastore_tests.QueryInTransaction) ... FAIL
branch: test_query_in_transaction (tests.datastore_tests.QueryInTransaction) ... ok

PR #3117, opened 2019-07-31
UA Server backup/restore

Honcharov12:
Adds UA Server backup and restore.

### DIFF
--- a/AppDB/appscale/datastore/scripts/ua_server_backup.py
+++ b/AppDB/appscale/datastore/scripts/ua_server_backup.py
@@ -1,6 +1,7 @@
 """ UA Server data backup. """
 
 import argparse
+import csv
 import errno
 import logging
 import os
@@ -20,6 +21,8 @@ from appscale.datastore.utils import tornado_synchronous
 # The location where the backup files will be stored.
 BACKUP_FILE_LOCATION = "/opt/appscale/backups/"
 
+logger = logging.getLogger(__name__)
+
 
 @tornado_synchronous
 @gen.coroutine
@@ -27,7 +30,7 @@ def get_table_sync(datastore, table_name, schema):
   """ Gets data from datastore.
 
   Args:
-    datastore: Datastore entity.
+    datastore: Cassandra adapter.
     table_name: Table name in datastore.
     schema: Table schema.
   """
@@ -51,30 +54,14 @@ def create_backup_dir(backup_dir):
 
   Args:
     backup_dir: Backup directory name.
-  Returns:
-    True if directory was created or existed before, false otherwise.
   """
-  logger = logging.getLogger(__name__)
   try:
     os.makedirs(backup_dir)
-    logger.info("Backup dir created: {0}".format(backup_dir))
-    return True
-  except OSError, os_error:
-    if os_error.errno == errno.EEXIST:
-      logger.warn("OSError: Backup directory already exists.")
-      return True
-    elif os_error.errno == errno.ENOSPC:
-      logger.error("OSError: No space left to create backup directory.")
-      logger.error(os_error.message)
-      return False
-    elif os_error.errno == errno.EROFS:
-      logger.error("OSError: READ-ONLY filesystem detected.")
-      logger.error(os_error.message)
-      return False
-  except IOError, io_error:
-    logger.error("IOError while creating backup dir.")
-    logger.error(io_error.message)
-    return False
+  except OSError as os_error:
+    if os_error.errno != errno.EEXIST:
+      raise
+
+  logger.info("Backup dir created: {0}".format(backup_dir))
 
 def main():
   logging.basicConfig(format=LOG_FORMAT, level=logging.INFO)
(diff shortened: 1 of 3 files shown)


AppDB/appscale/datastore/scripts/ua_server_restore.py
```diff
@@ -0,0 +1,80 @@
...
+
+  input = args.input
+
+  with open(input, 'r') as fin:
+    # Skip headers
+    next(fin)
+    # Iterate through all users in file
+    for line in fin:
+      array = line.rstrip().split(',')
+      put_entity_sync(db, USERS_TABLE, array[0], user_schema, array)
```
jeanleonov: If you're interested you can take a look into `appscale.datastore.cassandra_env.cassandra_interface.DatastoreProxy.batch_put_entity` which provides similar interface for adding entities but makes it in batches.
We don't have strong performance requirements to UAServer backup and restore as we don't expect to have many users in DB, so I would leave one-by-one insert for now, though you could make this improvement in separate PR when this one is merged.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,134 @@
...
+  if not is_created:
+    return
+
+  backup_timestamp = time.strftime("%Y-%m-%d_%H-%M-%S")
+  output = '{0}ua_server_{1}.csv'.format(BACKUP_FILE_LOCATION, backup_timestamp)
+
+  with open(output, 'w') as fout:
+    fout.write(','.join(USERS_SCHEMA) + '\n')
+    for row in reshaped_table:
+      fout.write(','.join(row) + '\n')
```
jeanleonov: If `row` has `,` or `\n` somewhere inside then this row won't be read properly on restore. You need to escape commas and `\n` here and unescape it during restore.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,134 @@
...
+    break
+
+  schema_cols_num = len(USERS_SCHEMA)
+  table = get_table_sync(db, USERS_TABLE, user_schema)[1:]
+  reshaped_table = reshape(table, schema_cols_num)
+
+  is_created = create_backup_dir(BACKUP_FILE_LOCATION)
+
+  if not is_created:
+    return
```
jeanleonov: `sys.exit(1)` would tell user that command was not succeeded.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,134 @@
...
+from appscale.datastore.utils import tornado_synchronous
+
+# The location where the backup files will be stored.
+BACKUP_FILE_LOCATION = "/opt/appscale/backups/"
+
+
+@tornado_synchronous
+@gen.coroutine
+def get_table_sync(datastore, table_name, schema):
+  """ Gets data from datastore.
```
jeanleonov: It actually gets data from Cassandra. Though it uses datastore code for doing it.
Though you had a reason to name `datastore` this way (this object has type `DatastoreProxy` I think) it's confusing. This object is rather Cassandra adapter..

AppDB/appscale/datastore/scripts/ua_server_backup.py
cdonati: I would recommend using the builtin [csv module](https://docs.python.org/2/library/csv.html) to avoid reinventing the wheel.

AppDB/appscale/datastore/scripts/ua_server_backup.py
cdonati: Raising an exception might be even better in this case since a non-zero exit code is not always obvious when running a script.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,134 @@
...
+
+def create_backup_dir(backup_dir):
+  """ Creates backup directory.
+
+  Args:
+    backup_dir: Backup directory name.
+  Returns:
+    True if directory was created or existed before, false otherwise.
+  """
+  logger = logging.getLogger(__name__)
```
cdonati: I would recommend defining this for the module rather than at the time that the function is executed.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,134 @@
...
+    logger.info("Backup dir created: {0}".format(backup_dir))
+    return True
+  except OSError, os_error:
+    if os_error.errno == errno.EEXIST:
+      logger.warn("OSError: Backup directory already exists.")
+      return True
+    elif os_error.errno == errno.ENOSPC:
+      logger.error("OSError: No space left to create backup directory.")
+      logger.error(os_error.message)
+      return False
```
cdonati: As a general Python style tip, it's usually fine to let the exceptions propagate up unless there is a reason to silence or convert them. Returning `False` or `True` to indicate the success of the operation is not as conventional.

For example, this function can be simplified to
```
def create_backup_dir(backup_dir):
  try:
    os.makedirs(backup_dir)
  except OSError as os_error:
    if os_error.errno != errno.EEXIST:
      raise

  logger.info("Backup dir created: {0}".format(backup_dir))
```

AppDB/appscale/datastore/scripts/ua_server_restore.py
```diff
@@ -0,0 +1,79 @@
...
+      retries -= 1
+      time.sleep(backoff)
+      continue
+    break
+
+  # If no response from cassandra
+  if retries == -1:
+    raise AppScaleDBConnectionError('No response from cassandra.')
+
+  input = args.input
```
cdonati: This is fine, but just be mindful that it's usually best to avoid shadowing builtin keywords like `input`.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,122 @@
...
+
+  # If no response from cassandra
+  if retries == -1:
+    raise AppScaleDBConnectionError('No response from cassandra.')
+
+  schema_cols_num = len(USERS_SCHEMA)
+  table = get_table_sync(db, USERS_TABLE, user_schema)[1:]
+  reshaped_table = reshape(table, schema_cols_num)
+
+  try:
```
cdonati: Since an uncaught exception will already write the error code and message to stdout, the try/except here is not necessary.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,122 @@
...
+  except OSError as os_error:
+    logger.error('OSError occurred!')
+    logger.error(os_error.message)
+    raise
+
+  backup_timestamp = time.strftime("%Y-%m-%d_%H-%M-%S")
+  output = '{0}ua_server_{1}.csv'.format(BACKUP_FILE_LOCATION, backup_timestamp)
+
+  with open(output, 'w') as fout:
+    writer = csv.writer(fout, delimiter=',')
```
cdonati: I think it would be helpful if the output file contained a headers line to make the fields clear. It might also be a good idea to add a comment line at the top indicating that this is a v1 output format so that when we change the backend implementation or the interface, we know how to handle this backup format.

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,120 @@
...
+  schema_cols_num = len(USERS_SCHEMA)
+  table = get_table_sync(db, USERS_TABLE, user_schema)[1:]
+  reshaped_table = reshape(table, schema_cols_num)
+
+  create_backup_dir(BACKUP_FILE_LOCATION)
+
+  backup_timestamp = time.strftime("%Y-%m-%d_%H-%M-%S")
+  output_file = '{0}ua_server_{1}.csv'.format(BACKUP_FILE_LOCATION, backup_timestamp)
+
+  # v1 output format
```
cdonati: Sorry, I meant a comment at the beginning of the output file so that future tools can recognize the file as having been generated by this version of the backup script.

The headers you added are probably enough, though, so this comment should not block a merge.

Honcharov12: Version can be mentioned in filename suffix. Will it be reasonable?

AppDB/appscale/datastore/scripts/ua_server_backup.py
```diff
@@ -0,0 +1,120 @@
...
+      user_schema = user_schema[1:]
+    else:
+      retries -= 1
+      time.sleep(backoff)
+      continue
+    break
+
+  # If no response from cassandra
+  if retries == -1:
+    raise AppScaleDBConnectionError('No response from cassandra.')
```
jeanleonov: Just to let you know, python has `while: ... else: ...` statement. So `if retries == -1:` could be replaced with `else:`.
It is just to let you know, no need to change it.

Honcharov12:
http://ocd.appscale.net:8080/job/Daily%20Build/6593/

AppDB/appscale/datastore/scripts/ua_server_backup.py
cdonati: I think it would be best to avoid storing format metadata in the filename since the user may want to rename their backup file. For example, they might want to keep a "latest" backup that they overwrite each time.

PR #3118, opened 2019-07-31
Store FoundationDB clusterfile content in Zookeeper

jeanleonov:
FDB clusterfile holds information needed to open connection to FDB database.
This branch adds functionality to easily set this information to zookeeper.

Probably this solution requires the smallest amount of code line changes and coding time.
I believe it's the only strong side of this PR.

It seems that AppController options is the only way we have for setting deployment properties. And I don't think it's perfect for cases like fdb-clusterfile-content.


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6574/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6578/

AppDB/appscale/datastore/scripts/datastore.py
```diff
@@ -889,32 +900,49 @@ def main():
...
       datastore_batch, transaction_manager, zookeeper=zookeeper,
       log_level=logger.getEffectiveLevel(),
       taskqueue_locations=taskqueue_locations)
   else:
     from appscale.datastore.fdb.fdb_datastore import FDBDatastore
+    clusterfile_path = args.fdb_clusterfile
+    if not clusterfile_path:
+      try:
+        clusterfile_content = zk_client.get(FDB_CLUSTERFILE_NODE)[0]
+        clusterfile_path = '/tmp/appscale-datastore-fdb.cluster'
```
sjones4: /tmp may not be a good location, perhaps /run/appscale/... or /var/lib/appscale/...?

jeanleonov: Is there some guideline on how to use linux directories properly?

sjones4: In this case it seems like the file:

- is long lived
- has a well known name

which suggests `/tmp` is not the best location.

jeanleonov: (facepalm) I don't know what I thought about. Yes, I see that this case is not the one requiring a guideline to understand that it's not tmp file.
I'll change it tomorrow.

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6610/

PR #3122, opened 2019-08-01
AppController firewall configuration disabled via APPSCALE_FIREWALL env

sjones4:
Allow the controllers firewall update functionality to be disabled via `APPSCALE_FIREWALL` environment variable. 

This environment variable could be set manually on hosts, be included in images, etc, to disable firewall modification if this is otherwise provided for the deployment (e.g. security groups and network acls in ec2)

The built-in firewall rules can be overly restrictive if we want to allow access to zookeeper state (e.g. external datastore with fdb backend) or datastores (e.g. cloud datastore proxy) and are more difficult to automate.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -290,7 +290,7 @@ class Djinn
   # A boolean that indicates whether or not we should turn the firewall on,
   # and continuously keep it on. Should definitely be on for releases, and
   # on whenever possible.
-  FIREWALL_IS_ON = true
+  FIREWALL_IS_ON = 'true' == (ENV['APPSCALE_FIREWALL'] || 'true')
 
   # The location on the local filesystem where AppScale-related configuration
   # files are written to.


obino:
Question: why disable it altogether instead of saying making it easier to modify? For example with some --allow zookeeper or whtaver?

sjones4:
In cloud environments where we are managing access via security groups (for example) having an additional layer of redundant configuration will make it harder to modify, even if we add simple options to do so.

sjones4:
Daily build no. 6584

sjones4:
Demo, set environment for `appscale-controller.service` to stop updating the firewall:

```
#
# mkdir -p /etc/systemd/system/appscale-controller.service.d
#
# echo -e "[Service]\nEnvironment=APPSCALE_FIREWALL=false\n" \
    > /etc/systemd/system/appscale-controller.service.d/20-disable-firewall.conf
#
# systemctl daemon-reload
#
# iptables -F
#
# appscale up
...
#
# iptables -L -n
Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
#
#
```

on systemd changes branch using environment defaults:

```
# echo "APPSCALE_FIREWALL=false" > /etc/default/appscale
#
# iptables -F
#
# appscale up
...
#
# iptables -L -n
Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
#
#
```

PR #3123, opened 2019-08-01
Use monotonic module when measuring intervals

cdonati:
This prevents system time changes from affecting the interval measured.

This does not address every instance of interval measurement, but it fixes most cases.

### DIFF
--- a/AdminServer/setup.py
+++ b/AdminServer/setup.py
@@ -17,7 +17,7 @@ install_requires = [
   'mock',
 ]
 if sys.version_info < (3,):
-  install_requires.append('futures')
+  install_requires.extend(['futures', 'monotonic'])
 
 setup(
   name='appscale-admin',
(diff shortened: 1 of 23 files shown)


AdminServer/setup.py
```diff
@@ -17,7 +17,7 @@
   'mock',
 ]
 if sys.version_info < (3,):
-  install_requires.append('futures')
+  install_requires.extend(['futures', 'monotonic'])
```
sjones4: don't we need to install the dependency when using python 3 in order to use `import monotonic`?

cdonati: Oops, you're right. Python 3 has the monotonic function under the `time` module, so I was thinking that we don't need an extra dependency there. I guess until we fully upgrade to Python 3 it makes sense to install the monotonic module in both cases.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6618/

PR #3125, opened 2019-08-02
Install FoundationDB client libraries

jeanleonov:
- Download foundationdb-clients_6.1.8-1_amd64.deb from S3 and install it.
 - Add foundationdb package as a requirement for datastore.

### DIFF
--- a/AppDB/setup.py
+++ b/AppDB/setup.py
@@ -17,7 +17,8 @@ setup(
     'monotonic',
     'mmh3',
     'SOAPpy',
-    'tornado'
+    'tornado',
+    'foundationdb==6.1.11'
   ],
   classifiers=[
     'Development Status :: 5 - Production/Stable',
--- a/AppDB/setup.py
+++ b/AppDB/setup.py
@@ -18,7 +18,7 @@ setup(
     'mmh3',
     'SOAPpy',
     'tornado',
-    'foundationdb==6.1.11'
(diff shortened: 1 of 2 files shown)


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6573/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6576/

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6585/

sjones4:
Why install the 6.1.11 python library and the 6.1.8 client package? would it be better to use the same version of each?

jeanleonov:
I used latest version of both. Good point though. 

We should use specific version of deb package as it is manually downloaded, so 6.1.8 is our choice, but for python package I would just skip minor version and set 6.1.

Would it work?

sjones4:
I think we need >= 6.1.8 since we use api version 610. Having looked into it a bit more 6.1.11 seems fine to me.

jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6602/

jeanleonov:
@sjones4 using foundationdb~=6.1.8 should allow us using any version inside 6.1 greater than 6.1.8.
So we'll get some latest minor updates and bug fixes but won't get API changes.

PR #3126, opened 2019-08-02
Deployment open port check should not use login ip

sjones4:
Update the open port check to use only the load balancers and to fail only if none have open ports.

The login host cannot always be used to check if any load balancer has an open port.


sjones4:
Daily build no. 6572

obino:
Question: why the login host cannot be used? We encountered some issues on a private cluster setup by hand with redundant LBs, but do we expect issues in general? Btw I like this way better, since we isolate possible misconfiguration of the external LB ...

AdminServer/appscale/admin/__init__.py
```diff
@@ -98,30 +98,32 @@ def wait_for_port_to_open(http_port, operation_id, deadline):
...
 
-  for load_balancer in appscale_info.get_load_balancer_ips():
-    while True:
-      if time.time() > deadline:
-        # The version is reachable from the login IP, but it's not reachable
-        # from every registered load balancer. It makes more sense to mark the
-        # operation as a success than a failure because the lagging load
-        # balancers should eventually reflect the registered instances.
-        break
+    if len(passed_lbs) == len(all_lbs):
```
obino: This would prevent the system to operate in case one of the LB failed: I don't think we remove a failed LBs from the list of nodes yet, do we?

sjones4: It means we do not proceed until the deadline is reached, which I believe is the same as the current behaviour.

sjones4:
This pull request is a rewrite of a change we made in a deployment to avoid checking the login ip. It may be the case that the original issue is best solved by making changes to the network.

I'm not sure that checking the login ip here is particularly useful as checking the load balancers is sufficient to verify the deployment and a failure to access via the login ip does not indicate an application issue.

I opened this pull request as a draft/discussion so we could decide if this change is an improvement or not (if not it can be closed)

sjones4:
Daily build no. 6802 # build with master merged to resolve conflicts

Note that the file this pull request was updating was renamed from `__init__.py` to `admin_server.py`. I kept this pull request as it has some discussion/review context but would be fine with closing it and opening a new one due to the changes in master.

PR #3130, opened 2019-08-06
Stop unassigned services

### LINKED ISSUE #3127: Removing datastore assignments does not stop datastore servers
When a machine's assignments node is deleted or the datastore service is removed from its assignments, the ServiceManager on that machine should stop the datastore servers.

cdonati:
This also waits until services are scheduled before ensuring that they are running.

Resolves #3127.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6642/

PR #3132, opened 2019-08-07
Limit connection retries when stopping celery

cdonati:
This fixes a bug where starting AppScale on bionic will hang when stopping nonexistent celery workers.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6598/

PR #3133, opened 2019-08-07
FoundationDB initialization script

jeanleonov:
- the script downloads and installs required deb packages.
 - configures foundationdb.
 - configures systemd unit (instead of init.d).
 - runs user command on fdbcli (creation of new DB is assumed).

This script is placed here just as a helper script to be run on the first boot if user requested to start FDB on deployment machines. We may want to add high-level provisioning script which would take care of mounting disks and starting different third-party services according to specified environmental variables.

### DIFF
--- a/scripts/foundationdb/foundationdb.service
+++ b/scripts/foundationdb/foundationdb.service
@@ -1,0 +1,33 @@
+# /etc/systemd/system/foundationdb.service
+[Unit]
+Description=FoundationDB Service
+AssertPathIsDirectory=/var/run/foundationdb
+AssertPathExists=/etc/foundationdb/foundationdb.conf
+AssertPathExists=/etc/foundationdb/fdb.cluster
+After=syslog.target network.target remote-fs.target nss-lookup.target
+RequiresMountsFor=${DATA_DIRS}
+
+[Service]
+User=foundationdb
+Group=foundationdb
+PIDFile=/var/run/foundationdb/fdbmonitor.pid
+
+Restart=always
+RestartSec=5
+
+ExecStart=/usr/lib/foundationdb/fdbmonitor --lockfile /var/run/foundationdb/fdbmonitor.pid --conffile /etc/foundationdb/foundationdb.conf
+ExecReload=/bin/kill -s HUP $MAINPID
+ExecStop=/bin/kill -s QUIT $MAINPID
+
+#MemoryLow=${MEMORY_LOW}
+#MemoryHigh=${MEMORY_HIGH}    IS NOT SUPPORTED IN cgroup-v1
+#MemoryMax=${MEMORY_MAX}
+MemoryLimit=${MEMORY_MAX}
+
+CPUAccounting=yes
+MemoryAccounting=yes
+TasksAccounting=yes
+#IOAccounting=yes
+
+[Install]
+WantedBy=multi-user.target
--- a/scripts/foundationdb/foundationdb.service
+++ b/scripts/foundationdb/foundationdb.service
@@ -1,33 +1,25 @@
 # /etc/systemd/system/foundationdb.service
 [Unit]
 Description=FoundationDB Service
-AssertPathIsDirectory=/var/run/foundationdb
+AssertPathIsDirectory=/run/foundationdb
 AssertPathExists=/etc/foundationdb/foundationdb.conf
 AssertPathExists=/etc/foundationdb/fdb.cluster
 After=syslog.target network.target remote-fs.target nss-lookup.target
-RequiresMountsFor=${DATA_DIRS}
 
 [Service]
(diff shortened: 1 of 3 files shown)


scripts/foundationdb/init-foundationdb.sh
```diff
@@ -0,0 +1,238 @@
...
+/etc/init.d/foundationdb stop
+update-rc.d foundationdb disable
+# rm /etc/init.d/foundationdb
+
+
+### Ensuring FDB directories are accessible ###
+#---------------------------------------------#
+log 'Making sure FDB directories are created and are owned by foundationdb user.'
+mkdir -pv /var/run/foundationdb
+chown foundationdb:foundationdb /var/run/foundationdb
```
sjones4: You need to use tmpfiles.d for this (so present after reboot)

scripts/foundationdb/init-foundationdb.sh
```diff
@@ -0,0 +1,238 @@
...
+log "Installing ${FDB_CLIENTS_PKG}"
+dpkg --install ${APT_CACHE}/foundationdb-clients_6.1.8-1_amd64.deb
+
+
+### Installing FDB server package ###
+#-----------------------------------#
+FDB_SERVER_PKG='foundationdb-server_6.1.8-1_amd64.deb'
+FDB_SERVER_MD5='80a427be14a329d864a91c9ce464d73c'
+log "Making sure ${FDB_SERVER_PKG} is in ${APT_CACHE}"
+cachepackage "${FDB_SERVER_PKG}" "${FDB_SERVER_MD5}"
```
cdonati: What do you think about installing the server package on our images so that someone trying appscale on vagrant, for example, could still start things without an internet connection?

scripts/foundationdb/init-foundationdb.sh
```diff
@@ -0,0 +1,238 @@
...
+  cp ${UNIT_FILE} "${UNIT_FILE}.$(date +'%Y-%m-%d_%H-%M-%S')"
+fi
+# Give FDB processes at most half of total memory (kill if greater).
+TOTAL_MEM_KB=$(awk '/MemTotal/ { print $2 }' /proc/meminfo)
+FDB_MEM_MAX=$(echo "$TOTAL_MEM_KB" | awk '{ printf "%d", $1 * 0.9 / 1024 }')
+DATA_DIRS=''
+for server_id in $(seq 4500 $((4500 + SERVER_PROCESSES_NUM - 1))); do
+  DATA_DIRS="${DATA_DIRS}${DATA_DIR}/${server_id} "
+done
+export MEMORY_MAX="${FDB_MEM_MAX}M"
```
cdonati: I think leaving the default configured memory limit (imposed by FDB itself) is fine for dev/test use cases.

We may want to configure the storage engine as `ssd` rather than the default so that available memory is less of a consideration.

scripts/foundationdb/foundationdb.service
```diff
@@ -0,0 +1,33 @@
+# /etc/systemd/system/foundationdb.service
```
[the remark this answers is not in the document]
jeanleonov: Not really, it just seems to be good practice to use more modern systemd.
And I'm trying practicing work with it where applicable.

scripts/foundationdb/init-foundationdb.sh
jeanleonov: I like it. But I think we still need to leave package downloading here as the script could be used on any machine, not necessarily appscale vm. And using cache should help to avoid downloading if deb is already in place.

I'm thinking to move fdb initialization script to separate repo (https://github.com/AppScale/appscale-thirdparties), so I would add fdb packages downloading to appscale_install_functions.sh as a separate PR.

jeanleonov:
It would probably be better to have backend initialization scripts in a separate repo: https://github.com/AppScale/appscale-thirdparties/pull/1
What do you think?

scripts/foundationdb/foundationdb.service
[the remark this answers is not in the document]
cdonati: Unfortunately, changing the script that systemd uses makes upgrading the FDB package more difficult in the future. If there's a crucial difference that makes the default systemd behavior unusable for our case, I can be convinced. However, I would also advocate for pushing that change upstream so that we don't need customizations in the future.

scripts/foundationdb/init-foundationdb.sh
cdonati: The separate repo seems reasonable. In that case, I don't think we would need to install the server package when building the main "appscale" repo. The appscale-thirdparties repo could contain a build script that installs what it needs. I mainly want to avoid fetching and installing things when it comes time to run them since there's a lot that can go wrong during that process.

PR #3135, opened 2019-08-07
Remove use of monit for appscale-controller, updates for systemd support

sjones4:
Update `appscale-controller` systemd configuration and remove use of monit for the controller service. 

Changes are:

- remove controller init scripts, add systemd unit (etc)
- controller no longer under monit control
- controller logs go to journal and then to log via rsyslog
- monit configs now in /run/appscale/monit.conf.d/ so removed on reboot
- changed paths to /run instead of using the /var/run symlink
- replace use of `service` command with `systemctl`

example `/var/log/appscale/controller.log` output:

```
Aug  7 17:13:32 euca-10-10-3-104  I, [32.827791#26251]  INFO -- : Installed certificate/key in /etc/nginx/appscaledashboard.pem.
Aug  7 17:13:32 euca-10-10-3-104  I, [32.828029#26251]  INFO -- : Installed certificate/key in /etc/nginx/appscaledashboard.key.
Aug  7 17:13:32 euca-10-10-3-104  I, [32.850613#26251]  INFO -- : Reloading nginx service.
```

The `appscale-controller` service now uses systemd functionality for capabilities/limits, dependencies, environment, and system protection (private `/tmp`, read-only system files)

There are related tools changes.

Daily build no. 6600


sjones4:
Related tools change AppScale/appscale-tools#759

sjones4:
Will open a new pull request when the full changes for monit removal are ready.

PR #3138, opened 2019-08-09
Display a selection menu for available accounts

cdonati:
This makes it easier for the user to see what service accounts they have available.

### DIFF
--- a/AppDashboard/lib/service_accounts.py
+++ b/AppDashboard/lib/service_accounts.py
@@ -63,7 +63,7 @@ class ProjectServiceAccounts(AppDashboard):
       raise AppHelperException(result.content)
 
     try:
-      accounts = json.loads(result.content)
+      accounts = json.loads(result.content)['accounts']
     except ValueError:
       raise AppHelperException('Invalid response: {}'.format(result.content))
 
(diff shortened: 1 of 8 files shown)


AppDashboard/lib/service_accounts.py
```diff
@@ -63,7 +63,7 @@ def get(self, project_id):
       raise AppHelperException(result.content)
 
     try:
-      accounts = json.loads(result.content)
+      accounts = json.loads(result.content)['accounts']
     except ValueError:
```
scragraham: This block can now throw a KeyError if 'accounts' isn't in the dictionary

AppServer/google/appengine/ext/datastore_admin/backup_handler.py
```diff
@@ -129,6 +130,19 @@
...
+def get_service_account_names():
+  """ AppScale: Fetch list of service accounts from IAM API. """
+  project_id = app_identity.get_application_id()
+  iam_location = 'https://127.0.0.1:17441'
+  url = iam_location + '/v1/projects/{}/serviceAccounts'.format(project_id)
+  token = app_identity.get_access_token(
+      ['https://www.googleapis.com/auth/cloud-platform'])[0]
+  headers = {'Authorization': 'Bearer {}'.format(token)}
+  response = urlfetch.fetch(url, headers=headers, validate_certificate=False)
+  accounts = json.loads(response.content)['accounts']
```
scragraham: not sure how pedantic you want to be, but this could throw an exception if 'accounts' isn't in the json

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6630/

PR #3139, opened 2019-08-12
Remove support for Debian Jessie

cdonati:
Since Debian no longer offers cloud images for Jessie, we are not able to test it easily.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6627/

PR #3144, opened 2019-08-16
Allow custom service accounts for restore operations

cdonati:
This extends the datastore admin interface to allow custom service accounts for restore operations.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6659/

PR #3147, opened 2019-08-21
Install newer psutils to handle kernel 5.X output of /proc/diskstats

scragraham:
on Azure the base ubuntu image seems to be using the 5.0 kernel, whereas GCP and EC2 are running 4.x. The issue here is that the number of fields in diskstats is different on the newer kernel which causes an issue with psutils 5.1.3.

### DIFF
--- a/Hermes/setup.py
+++ b/Hermes/setup.py
@@ -14,7 +14,7 @@ setup(
     'appscale-common',
     'kazoo',
     'tornado',
-    'psutil==5.1.3',
+    'psutil==5.6.3',
     'attrs>=18.1.0',
     'mock',
   ],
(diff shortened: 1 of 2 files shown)


scragraham:
Sigh, looks like I need to find the other places psutil is pinned
```
Installing collected packages: psutil
  Found existing installation: psutil 5.6.3
    Uninstalling psutil-5.6.3:
      Successfully uninstalled psutil-5.6.3
Successfully installed psutil-5.1.3
```

scragraham:
build/demo: http://ocd.appscale.net:8080/job/Create-Azure-Image/636/

PR #3151, opened 2019-08-22
Use systemd instead of monit for service management

sjones4:
Use systemd for service management instead of monit.

Some notable changes are:

* Build/install - systemd unit and tmpfiles.d are installed, monit is not. `systemctl` now used.
* Controller - `service_helper` replaces `monit_interface`, init script removed
* AdminServer - `ServiceOperator` replaces `MonitOperator`, not concerned with service dependencies, services are named with `appscale-` prefix.
* Hermes - Uses systemctl for service listing
* common - `monit_interface` replaced with `service_helper`, static templates moved

Now that systemd is used we generally do not write pid files and services log via the journal.

There is a corresponding tools branch that is required with this pull request.


sjones4:
Will open a new pull request with additional changes.

PR #3152, opened 2019-08-22
Reuse TCP connections for datastore calls

cdonati:
This reduces latency for datastore API calls in most cases.

HAProxy keeps the connection to the client open and load balances HTTP requests between backend servers.

### DIFF
--- a/AppServer/google/appengine/api/datastore_distributed.py
+++ b/AppServer/google/appengine/api/datastore_distributed.py
@@ -33,6 +33,9 @@ import sys
 import threading
 import warnings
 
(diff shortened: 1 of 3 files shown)


sjones4:
With previous versions of this change there were issues with the dashboard datastore viewer due to urllib3 not being available.

cdonati:
@sjones4 Good point. I made some changes so that it can still make requests inside the runtime sandbox.

cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6732/

PR #3154, opened 2019-08-28
Register and route other services

cdonati:
This updates the UA, TaskQueue, and blobstore service to behave like the datastore and search2 service in that they add a registration node in ZooKeeper and have requests routed to them based on those entries.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6701/

PR #3155, opened 2019-08-29
Firewall configuration wrapper script using iptables-restore to apply rules

sjones4:
This pull request adds a `firewall-update.sh` script for wrapping the existing `firewall.conf`. 

The wrapper script collects the rules for application on completion of the wrapper. The wrapper checks for changes before using `iptables-restore` to apply, so will result in the `FILTER` table being flushed less frequently. Due to the reduced flush any packed/byte counters would not be frequently reset, which is a change in behaviour (but seems preferable)

When the wrapper is used we no longer flush tables other than the `FILTER` table so it would be up to the user to ensure these were correct for use with an AppScale deployment.

### DIFF
--- a/AppController/djinn.rb
+++ b/AppController/djinn.rb
@@ -2777,7 +2777,7 @@ class Djinn
     # list of nodes in the firewall.
     write_locations
     if FIREWALL_IS_ON
-      Djinn.log_run("bash #{APPSCALE_HOME}/firewall.conf")
+      Djinn.log_run("bash #{APPSCALE_HOME}/scripts/firewall-update.sh #{APPSCALE_HOME}/firewall.conf")
     end
   end
 
--- a/firewall.conf
+++ b/firewall.conf
@@ -1,4 +1,6 @@
 #!/bin/bash
+# Run using scripts/firewall-update.sh firewall.conf for atomic updates.
+# update script redefines "iptables" to collect rules and apply on exit.
 
 # Flush any current firewall settings
 iptables -F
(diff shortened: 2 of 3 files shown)


sjones4:
Daily build no. 6694

sjones4:
> Is there a reason the tempfile is needed instead of changing the format of the firewall.conf file itself?

The `firewall.conf` file is a bash script that gets run to generate rules. I'm not sure if we could eliminate the need to template/generate the firewall configuration (using ipsets for hosts in the cluster for example)

Longer term, we would probably want to move away from using iptables directly (perhaps firewalld)

sjones4:
One thing that needs to change with this pull request is using only:

```
iptables -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
```

and removing the older:

```
iptables -A INPUT -m state --state ESTABLISHED,RELATED -j ACCEPT
```

either should generally work, but the check for a failing iptables command does not make sense with the wrapper (which will accept all rules and apply them later)

scragraham:
Eventually firewall.conf needs a proper home in the src tree other than at the root level.

obino:
I think I'm with Chris here (and with Steve on the long term): why don't we switch firewall.conf to be in the format of ipdates-save so we can use iptables-restore directly? The only script that modifies it is the datastore viewer script. This way we remove a shell script (we use iptables-restore instead) and we just have a config file instead.

sjones4:
Closing this pull request. We can look at simpler solutions with static firewall configuration.

PR #3157, opened 2019-08-30
Use FDB backend if clusterfile content is defined

cdonati:


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6803/

PR #3160, opened 2019-09-05
AppController terminate update for newer ifconfig/monit output

sjones4:
Update ` AppController/terminate.rb` to allow the newer `ifconfig` and `monit summary` output as per bionic.

Daily build no. 6742


sjones4:
As previously noted, the build for this change is daily build no. 6742

Not planning on a demo for this change as the monit part will be replaced for 4.0 and the other change is to copy the regex we already use here:

https://github.com/AppScale/appscale/blob/5ee594d8aa930a69da72dc4f584fedd04e47fd65/AppController/lib/helperfunctions.rb#L423

This change also relates to cassandra use, so may be removed entirely.

PR #3164, opened 2019-09-09
Slow stop services

obino:
This PR makes some startup time faster: this is particularly noticeable on autoscaled node.  In details:

- stop services only if they are actually running (avoid unnecessary calls to monit)
- skip check DB is primed on autoscaled nodes
- make hermes and admin_server started in a thread too (like the others services)
- use appscale-amin summary instead of monit summary (faster in most cases)


obino:
OCD build http://ocd.appscale.net:8080/job/Daily%20Build/6758/

obino:
There have been few rebuilds of the autoscale triggers test and they all got back blue
http://ocd.appscale.net:8080/job/Autoscale-test-manual-triggers/2877/

AppController/djinn.rb
```diff
@@ -3310,28 +3310,40 @@ def start_stop_api_services
...
-    Djinn.log_info('Ensuring data layout version is correct')
-    layout_script = `which appscale-data-layout`.chomp
-    retries = 10
-    loop {
-      output = `#{layout_script} --db-type cassandra 2>&1`
-      if $?.exitstatus == 0
+    # Autoscaled nodes do not need to check if the datastore is primed: if
+    # we got this far, it must be primed.
+    am_i_autoscaled = false
+    get_autoscaled_nodes.each { |node|
```
cdonati: As a ruby tip, `.any?` is useful for these cases.

PR #3169, opened 2019-09-17
Add test for batch put side effects

cdonati:
This makes sure that only the final mutation's index entries are written during a batch put.


cdonati:
http://ocd.appscale.net:8080/job/Datastore/1308/

PR #3105, opened 2019-07-22
TaskQueue Python 3

Honcharov12:
Moves tq to py3

PR #3172, opened 2019-09-19
Add postgres_dsn option to djinn

jeanleonov:
This PR adds postgres_dsn property which is than set to zookeeper:/appscale/tasks/postgres_dsn.
Mentioned zookeeper node is not used in taskqueue yet, but another PR will be published a bit later, after https://github.com/AppScale/appscale/pull/3105 is fixed and merged.

http://ocd.appscale.net:8080/job/Daily%20Build/6817/


jeanleonov:
http://ocd.appscale.net:8080/job/Daily%20Build/6818/

PR #3175, opened 2019-09-25
Pass update_indexes request to datastore service

cdonati:
This allows the datastore service to handle the implementation details for how index metadata is stored. This does not change the implementation yet, but future work in this area includes storing index definitions in FDB for that backend.


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6832/

PR #3176, opened 2019-09-27
Store composite index metadata in FDB

cdonati:
This allows index read and write operations to be consistent with the index definition state.

The CompositeIndexManager keeps track of what indexes are defined for a project and which ones can be queried. An elected datastore server will watch for schema changes and backpopulate indexes when necessary.

Potential future work in this area includes distributed index rebuilding for faster population when there are a lot of existing entities. Slices can be distributed based on the scatter property.


sjones4:
Daily build no. 6992

PR #3181, opened 2019-10-01
UA Server Postgres

Honcharov12:
Adds postgres option as a backend to UA Server.


AppDB/appscale/datastore/scripts/ua_server.py
```diff
@@ -64,6 +65,115 @@
 
 logger = logging.getLogger(__name__)
 
+zk_client = None
+
+table_name = "ua_server"
```
sjones4: use a name that relates to the contents of the table `ua_user(s)`?

AppDB/appscale/datastore/scripts/ua_server.py
```diff
@@ -64,6 +65,115 @@
...
+          '  cookie_exp integer,'
+          '  cksum integer,'
+          '  enabled boolean,'
+          '  type varchar(9),'
+          '  is_cloud_admin boolean,'
+          '  capabilities varchar(255),'
+          '  PRIMARY KEY (email)'
+          ');'
+            .format(table=table_name)
+        )
```
sjones4: We might want to create a schema for the ua server so it has its own namespace, e.g.

```CREATE SCHEMA IF NOT EXISTS  "appscale_ua";```

though we would then need to check if the schema name can collide with other schemas (taskqueue)

File: AppDB/appscale/datastore/scripts/ua_server_backup.py [hunk 3]
```diff
@@ -34,6 +114,17 @@ def get_table_sync(datastore, table_name, schema):
     table_name: Table name in datastore.
     schema: Table schema.
   """
+  if pg_connection_wrapper:
+    pg_connection = pg_connection_wrapper.get_connection()
+    with pg_connection:
+      with pg_connection.cursor() as pg_cursor:
+        pg_cursor.execute(
+          'SELECT * FROM "{table}" '
```
sjones4: should list columns here

AppDB/appscale/datastore/scripts/ua_server.py
```diff
@@ -64,6 +65,115 @@
 
 logger = logging.getLogger(__name__)
 
+zk_client = None
+
+table_name = "ua_server"
+
+
+def is_connection_error(err):
```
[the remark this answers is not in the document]
Honcharov12: I also thought about it, but it also adds dependency from psycopg2 to all services that use common module

Honcharov12:
Thanks @sjones4 for such a quick review :)

sjones4:
Daily build no. 6927

The build is run with changes to configure postgres for hawkeye and an additional commit that sets the postgres dsn for ua server use when the dsn is set in the appscale file.

PR #3183, opened 2019-10-07
Tasks should use http to avoid scheme redirects since they use HAProxy

whoarethebritons:
Addresses an issue where tasks attempt to go to https for tasks and headers are missing for a proper redirect. Since tasks send requests through HAProxy, there is no need for scheme to be involved and this will bypass the part of the AppServer that does the scheme redirection.


cdonati:
Demo (https://github.com/AppScale/hawkeye/pull/179)
Against master: http://ocd.appscale.net:8080/job/Hawkeye/20287/
Against branch: http://ocd.appscale.net:8080/job/Hawkeye/20286/

PR #3185, opened 2019-10-11
Docker image build updated to allow systemctl command failure

sjones4:
Fix the docker image build by allowing `systemctl daemon-reload` to fail.

With this fix the image build will complete but the resulting image is not functional.


sjones4:
Demo is that a manual docker build works when modified to use this branch:

```
# Get the bootstrap script, and then run it.
RUN cd /root && \
    curl -o bootstrap.sh -Ls https://raw.githubusercontent.com/AppScale/appscale/master/bootstrap.sh && \
    bash ./bootstrap.sh --repo git://github.com/sjones4/appscale.git --branch issue-docker-build-systemd
```

PR #3189, opened 2019-10-16
Catch unexpected errors when checking process line

cdonati:
The PosixPsutil raises a `NoMethodError` when the process name no longer exists. This is meant to catch that and similar exceptions that the library has not dealt with.

### DIFF
--- a/AppController/lib/ejabberd.rb
+++ b/AppController/lib/ejabberd.rb
@@ -63,7 +63,10 @@ module Ejabberd
           begin
             next unless process.name == 'epmd'
             process.terminate if process.cmdline.include?('-daemon')
-          rescue PosixPsutil::NoSuchProcess
+          # The PosixPsutil library does not always raise NoSuchProcess when
+          # the given process no longer has a name. StandardError is meant
+          # as a fallback when it doesn't behave properly.
+          rescue PosixPsutil::NoSuchProcess, StandardError
             next
           end
         }


cdonati:
http://ocd.appscale.net:8080/job/Daily%20Build/6940/

cdonati:
Demo
I was able to reproduce this issue by running an executable with a newline character in the process name. The regex match in the psutil library stored the process name as `nil`.
```
$ cat test_psutil.rb 
require 'posixpsutil'
PosixPsutil::Process.processes.each { |process|
  begin
    process.name
  rescue PosixPsutil::NoSuchProcess
    puts "skipping process"
  end
}
$ cp /bin/sleep 'sleep'$'\n''script'
$ './sleep
script' 300 &
$ ruby test_psutil.rb
Traceback (most recent call last):
test_psutil.rb:1:in `<main>': uninitialized constant PosixPsutil (NameError)
root@ubuntu-bionic:~# vim test_psutil.rb
root@ubuntu-bionic:~# ruby test_psutil.rb 
Traceback (most recent call last):
	3: from test_psutil.rb:2:in `<main>'
	2: from test_psutil.rb:2:in `each'
	1: from test_psutil.rb:4:in `block in <main>'
/var/lib/gems/2.5.0/gems/posixpsutil-0.1.0-x86_64-linux/lib/posixpsutil/process.rb:284:in `name': undefined method `length' for nil:NilClass (NoMethodError)
```
and after adding StandardError to the rescue line:
```
# ruby test_psutil.rb 
skipping process
```

PR #3191, opened 2019-10-16
Allow ovewrite of deployed files

obino:
This is needed to have the up --update option to work correctly.


obino:
Roger: @sjones4 would you think that having a separate rule for the appcontroller would be better? We could have the ProtectSystem=true for all the other services ...

sjones4:
We could add `ProtectSystem=true` to the individual unit files other than the app controller (`system/units/appscale-controller.service`), I think `system/units/appscale-instance-run@.service` would be of most benefit (for more secure apps)

PR #3198, opened 2019-10-28
AppController set postgres dsn in zookeeper for ua server

sjones4:
Set the `postgres_dsn` for ua server so postgres is used when available.


sjones4:
Daily build no. 6949

Tests show that there are no failures due to this change.  For postgres to be used during tests changes from appscale#3177 must also be present ( i.e. appscale@3b28748160ee75fb06b8c64817e3e1c96f6ad938 )

PR #3200, opened 2019-10-29
Datastore foundationdb put fixes

sjones4:
Update datastore foundationdb backend so puts for new entities without identifiers will not be collapsed to a single entity. This update also modifies puts for new entities so that an allocated key is returned in the response.


sjones4:
Daily build no. 6955

sjones4:
Daily build no. 7000

Build after master merged to resolve conflicts.

sjones4:
Demo is that `Hawkeye` and `Datastore` tests pass using FDB with this fix in place, as per build no. 6998.

PR #3177, opened 2019-09-30
Remove cassandra backend for PullQueues

jeanleonov:
Should go after https://github.com/AppScale/appscale-testing/pull/434 and https://github.com/AppScale/appscale-thirdparties/pull/2.

PR #3201, opened 2019-10-29
AppTaskQueue end-to-end test suite postgres setup fix

sjones4:
The AppTaskQueue end-to-end test suite fails to configure postgres as the package is now installed in the image but the postgresql service is not running.

Other minor changes align this setup script with the third party postgres configuration script.

https://github.com/AppScale/appscale-thirdparties/blob/65905287ba21198c14a9d2c0f265b415d234a6c5/postgres/configure-and-start-postgres.sh#L76-L99


sjones4:
Daily build no. 6956

sjones4:
This is not needed now that #3177 is merged

Issue #3203, opened 2019-10-30
ZooKeeper sometimes fails to start


cdonati:
When performing a `appscale down --clean && appscale up`, ZooKeeper failed to start for me. The following was in the logs:
```
java.io.IOException: Unable to create data directory /opt/appscale/zookeeper/version-2
        at org.apache.zookeeper.server.persistence.FileTxnSnapLog.<init>(FileTxnSnapLog.java:85)
        at org.apache.zookeeper.server.ZooKeeperServerMain.runFromConfig(ZooKeeperServerMain.java:110)
        at org.apache.zookeeper.server.ZooKeeperServerMain.initializeAndRun(ZooKeeperServerMain.java:87)
        at org.apache.zookeeper.server.ZooKeeperServerMain.main(ZooKeeperServerMain.java:53)
        at org.apache.zookeeper.server.quorum.QuorumPeerMain.initializeAndRun(QuorumPeerMain.java:116)
        at org.apache.zookeeper.server.quorum.QuorumPeerMain.main(QuorumPeerMain.java:78)
```

sjones4:
Looks like there is a timing issue to reproducing this but it happens frequently if not every time.

On `appscale down --clean` we should perform a `systemctl daemon-reload` so that any runtime systemd configuration that was deleted is applied. I think this will prevent zookeeper attempting to start too soon on `appscale up`.

The zookeeper drop-in unit (`10-appscale-zookeeper.conf`) should probably be updated to include:

```AssertPathExists=/opt/appscale/zookeeper```

as we delete this directory on `appscale down --clean` and zookeeper cannot run without it.

PR #3206, opened 2019-10-31
AppController reload systemd configuration on terminate

sjones4:
On `appscale down --clean` we remove runtime systemd configuration but were not reloading to apply the changes.

Fixes appscale/appscale#3203


sjones4:
Daily build no. 6964

sjones4:
Demo was to run `appscale down --clean && appscale up` a few times to verify zookeeper restarts.

PR #3210, opened 2019-11-01
Handle deleted entity entries correctly

cdonati:
This fixes a bug in the stats buffer update handling.

The "present" attribute indicates whether or not there is an FDB key value present for the datastore entity key. Fetching a deleted entity will return a VersionEntry with "present"=True and "has_entity"=False.


sjones4:
Daily build no. 6997

PR #3214, opened 2019-11-05
Bootstrap allow options for git clone

sjones4:
Add bootstrap script configuration for git clone options. The default value skips the checkout as we always perform an explicit checkout after cloning.

This option can be used with installations where a full git clone of each repository is not required.

This option is omitted from the usage/help as it is not expected to be widely used.


sjones4:
Daily build no. 6979

PR #3225, opened 2020-02-13
Remove datastore cassandra support

sjones4:
Remove all datastore/ua server cassandra support and do not install cassandra.


sjones4:
Daily build no. 7140

